Issue #209, opened 2024-08-01
`pull-was-reviewed` is coming for the person who didn't do the review


yegor256:
Check this report: https://zerocracy.github.io/judges-action/zerocracy.html

Search for `pull-was-reviewed` and `zerocracy/baza#154`. As you see, we have two reports for the same issue, but for two different `who`s. While the review was done by one person (`526301`), we got two facts. Something is wrong with `github-events.rb`.

<img width="1566" alt="Screenshot 2024-08-01 at 08 21 08" src="https://github.com/user-attachments/assets/a820b252-0e72-4245-bc86-f298033bb419">

It happened here: https://github.com/zerocracy/baza/pull/154

Suban05:
@yegor256 may I take this one?

yegor256:
@Suban05 yes, please try

Suban05:
@yegor256 for `zerocracy/#43` `pull-was-reviewed` is also incorrectly defined?
![2024-08-01_20-08-40](https://github.com/user-attachments/assets/af58132d-e22d-4467-a912-c1e2e1fa5647)

yegor256:
@Suban05 yes, the same problem

Suban05:
@yegor256 Do I understand correctly that the user who submitted a pull request should not be the reviewer?

yegor256:
@Suban05 yes, of course. The reviewer is the one who submitted a review. Well, technically, maybe, it should be possible for a PR author to submit a review. But this is not what has happened in our cases.

Suban05:
@yegor256 I added this code
```rb
if json[:payload][:pull_request][:user][:id] != json[:payload][:review][:user][:id] && 
json[:payload][:action] == 'created'
```
instead of
```rb
case json[:payload][:action]
when 'created'
```      
If I understood everything correctly, then this should help, but I don't understand how to write test in this repository to be sure.

yegor256:
@Suban05 I think I understand the problem. When JSON arrives on `PullRequestReviewEvent`, its structure looks like this:

```
{
  actor: { id: ... } # this is the author of the pull request
  payload: {
    pull_request: {
      user: { id: ... } # this is also the author of the PR
    },
    review: {
      user: { id: ... } # this is the guy who created the review
    }
  }
}
```

We are using (by mistake) the id from the `actor`. Am I right?

yegor256:
@Yegorov I will create a test now, to show you an example

yegor256:
[...]

PR #215, opened 2024-08-02
#209 `pull-was-reviewed` is coming for the person who didn't do the review

Suban05:
I added a test with similar JSON data that arrives to us.
There, I added two events similar to our problem
In `github-events.rb`, when the event is `PullRequestReviewEvent`, I added a check to see if the user is a reviewer or not.

### DIFF
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -93,7 +93,10 @@ Fbe.iterate do
       end
 
     when 'PullRequestReviewEvent'
-      if json[:payload][:pull_request][:user][:id].to_i != fact.who && json[:payload][:action] == 'created'
+      case json[:payload][:action]
+      when 'created'
+        skip_event(json) if json[:payload][:pull_request][:user][:id].to_i == fact.who
+
         fact.issue = json[:payload][:pull_request][:number]
         fact.what = 'pull-was-reviewed'
         pull = Fbe.octo.pull_request(json[:repo][:name], fact.issue)
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -43,10 +43,6 @@ Fbe.iterate do
     raise Factbase::Rollback
   end
 
-  def self.reviewer?(json)
-    json[:payload][:pull_request][:user][:id] != json[:payload][:review][:user][:id]
-  end
-
   def self.fill_up_event(fact, json)
     fact.when = Time.parse(json[:created_at].iso8601)
     fact.event_type = json[:type]
@@ -97,7 +93,7 @@ Fbe.iterate do
       end
 
     when 'PullRequestReviewEvent'
-      if reviewer?(json) && json[:payload][:action] == 'created'
+      if json[:payload][:pull_request][:user][:id].to_i != fact.who && json[:payload][:action] == 'created'
         fact.issue = json[:payload][:pull_request][:number]
         fact.what = 'pull-was-reviewed'
         pull = Fbe.octo.pull_request(json[:repo][:name], fact.issue)
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -43,6 +43,10 @@ Fbe.iterate do
     raise Factbase::Rollback
   end
 
+  def self.reviewer?(json)
+    json[:payload][:pull_request][:user][:id] != json[:payload][:review][:user][:id]
+  end
+
   def self.fill_up_event(fact, json)
     fact.when = Time.parse(json[:created_at].iso8601)
     fact.event_type = json[:type]
@@ -93,8 +97,7 @@ Fbe.iterate do
       end
 
     when 'PullRequestReviewEvent'
-      case json[:payload][:action]
-      when 'created'
+      if reviewer?(json) && json[:payload][:action] == 'created'
         fact.issue = json[:payload][:pull_request][:number]
         fact.what = 'pull-was-reviewed'
         pull = Fbe.octo.pull_request(json[:repo][:name], fact.issue)
--- a/test/judges/test-github-events.rb
+++ b/test/judges/test-github-events.rb
@@ -146,7 +146,7 @@ class TestGithubEvents < Minitest::Test
           }
         },
         {
-          id: '40623323541',
+          id: '40623323542',
           type: 'PullRequestReviewEvent',
           public: true,
           created_at: '2024-07-31 12:45:09 UTC',
@@ -209,6 +209,11 @@ class TestGithubEvents < Minitest::Test
     )
     stub_request(:get, 'https://api.github.com/user/42').to_return(
       body: { id: 42, login: 'torvalds' }.to_json, headers: {
+        'content-type': 'application/json'
+      }
+    )
+    stub_request(:get, 'https://api.github.com/user/526200').to_return(
+      body: { id: 526_200, login: 'test' }.to_json, headers: {
         'content-type': 'application/json'
       }
     )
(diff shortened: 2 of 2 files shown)


Suban05:
@yegor256 review it, please

test/judges/test-github-events.rb
```diff
@@ -55,6 +55,120 @@ def test_create_tag_event
           type: 'CreateEvent',
           repo: { id: 42 },
           payload: { ref_type: 'tag', ref: 'foo' }
+        },
```
yegor256, line 58: @Suban05 what's the point of putting two test scenarios into one test method? It's definitely a bad practice, read this: https://www.yegor256.com/2018/12/11/unit-testing-anti-patterns.html Just create a new test methods, leaving the existing one untouched. Code duplication that may happen in this case is not a problem.

judges/github-events/github-events.rb
```diff
@@ -43,6 +43,10 @@ def self.skip_event(json)
     raise Factbase::Rollback
   end
 
+  def self.reviewer?(json)
```
yegor256, line 46: @Suban05 it's not a good idea in general to make methods with one line just to make code more readable. Instead, you can use a boolean variable.

judges/github-events/github-events.rb
```diff
@@ -93,8 +97,7 @@ def self.fill_up_event(fact, json)
       end
 
     when 'PullRequestReviewEvent'
-      case json[:payload][:action]
-      when 'created'
+      if reviewer?(json) && json[:payload][:action] == 'created'
```
yegor256, line 100: @Suban05 this doesn't sound like a solution. I believe, we should simply read `who` from `json[:payload][:review][:user]`

yegor256 (changes requested):
@Suban05 I suggest some improvements

test/judges/test-github-events.rb
Suban05: @yegor256 I completely agree with you that this is a wrong approach. However, I couldn't make it work using a separate test. I tried using a different method, but every time, one of the tests failed, and each time by accident, I can't figure out why. I did it the same way as the first test, but each time in one of the tests, the `$fb` variable does not contain data, and the query returns an empty result.
Maybe I didn't understand how to use `Fastbase` correctly

yegor256: @Suban05 in such a situation you shouldn't blame yourself and waste time on experimenting. Just submit a new ticket, claiming that our testing system is not stable -- it doesn't allow you to add a new test method. I will fix this bug, you will earn points for reporting a ticket. Then, you continue with this PR.

judges/github-events/github-events.rb
Suban05: @yegor256 I added two examples of JSON, which we fetch from GitHub.
Here is the first example
```
payload: {
  action: 'created',
  review: {
    id: 2_210_067_609,
    node_id: 'PRR_kwDOL6GCO86DuvSZ',
    user: {
      login: 'yegor256',
      id: 42,
      node_id: 'MDQ6VXNlcjUyNjMwMQ==',
      type: 'User'
    },
}
```
Here the reviewer is a user with id = 42
But in the second example, we see:

```
payload: {
  action: 'created',
  review: {
    id: 2_210_067_609,
    node_id: 'PRR_kwDOL6GCO86DuvSZ',
    user: {
      login: 'test',
      id: 526_200,
      node_id: 'MDQ6VXNlcjUyNjMwMQ==',
      type: 'User'
    }
}    
```          
Here `json[:payload][:review][:user]` is not the reviewer, because if we look at who the author of the PR is, we will see:

```
pull_request: {
  url: 'https://api.github.com/repos/yegor256/judges/pulls/93',
  id: 1_990_323_142,
  node_id: 'PR_kwDOL6GCO852oevG',
  number: 93,
  state: 'open',
  locked: false,
  title: 'allows to push gizpped factbase',
  user: {
    login: 'test',
    id: 526_200,
    node_id: 'MDQ6VXNlcjE2NDYwMjA=',
    type: 'User',
    site_admin: false
  }
}
```
I thought we had to skip examples like the second one.

Suban05:
@yegor256 I have updated the tests and changed the condition for adding a pull request review event
what do you think about it https://github.com/zerocracy/judges-action/pull/215#discussion_r1701498134?
Is this the wrong reasoning?

yegor256:
@Suban05 if the JSON's that you put into the tests are the one you received from GitHub API then you reasoning is more or less correct. Did you get them from the API?

Suban05:
@yegor256 yes, I just deleted unnecessary fields to reduce the size of the JSON
I sent a request to this url https://api.github.com/repositories/799113787/events?per_page=100.
If you want, you can check it. Just execute it and find event with id = 40626707785.
This is an example of an event that we should skip.
And the example with id 40694129321, 40685823188 are examples where reviewers should be identified correctly

judges/github-events/github-events.rb
```diff
@@ -93,8 +93,7 @@ def self.fill_up_event(fact, json)
       end
 
     when 'PullRequestReviewEvent'
-      case json[:payload][:action]
-      when 'created'
+      if json[:payload][:pull_request][:user][:id].to_i != fact.who && json[:payload][:action] == 'created'
```
yegor256, line 96: @Suban05 to make this place in the script more compatible with other similar places I would rewrite it like this:

```
when 'PullRequestReviewEvent'
  case json[:payload][:action]
  when 'created'
    skip_event(json) if json[:payload][:pull_request][:user][:id].to_i == fact.who
    # here we process it 
```

Suban05: @yegor256 done

Suban05:
@yegor256 sometimes, tests still fail if they are run many times in a row. I have also checked this on the master branch. Can I create a ticket for this?

yegor256:
@Suban05 of course, when something doesn't work as you expect -- make a ticket immediately

Suban05:
@yegor256 I merged branch `zerocracy:master` into this branch. Now tests are passing

yegor256:
@Suban05 thanks!

0crat:
@yegor256 Thank you for your code review! You've earned +39 points: +25 as a base reward, +14 for reviewing 140 hits-of-code, but -10 for having only 4 comments. According to our policy, we encourage more detailed feedback, aiming for at least 6 comments per review. Your current balance is [-564](https://zerocracy.github.io/judges-action/zerocracy-vitals.html). Keep up the good work, and remember that more comprehensive comments can lead to higher rewards in future reviews!

0crat:
@Suban05 Hey there, great work on your contribution! 🎉 You've racked up +23 points for this one: +30 as the base, with a small deduction of -7 for exceeding 200 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (you hit 326). Don't worry though, it's still a solid score! Remember, [keeping it concise](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) can boost your points next time. Your total balance is now [+129](https://zerocracy.github.io/judges-action/zerocracy-vitals.html). Keep the contributions coming - we love seeing your code!

0crat:
@Suban05 Thanks for your review! You've earned +5 points: +25 base, -40 for self-review (not recommended per policy), +0.1 for 5 hits-of-code, -20 for fewer than 6 comments, +5 for the comments made (capped at 20), -10 for <15 hits-of-code, adjusted to minimum 5 points. Remember, more comments and reviewing others' code can boost your score next time! Your current balance is [+302](https://zerocracy.github.io/judges-action/zerocracy-vitals.html).

PR #582, opened 2025-04-29
#539: implement fill up abandoned facts for quality-of-service

### LINKED ISSUE #539: quality-of-service doesn't fill up abandoned facts
Currently, this is how it works. The `incremate` function tries to collect all necessary data, but stops when time is running out. Then, the fact created remains incomplete, with some data missed. We never get back to this fact and don't try to add what's missing. This is wrong.

Let's fix it by adding a new cycle to the `quality-of-service.rb` script. It should go through all `what:quality-of-service` facts, see what's missing, and call `incremate` for them. The `incremate` should check whether the data is already there and avoid duplicate calculations. It should only add what's missing.

Yegorov:
Need merge this PR https://github.com/yegor256/factbase/pull/226 before

Closes #539

### DIFF
--- a/judges/quality-of-service/average_backlog_size.rb
+++ b/judges/quality-of-service/average_backlog_size.rb
@@ -27,7 +27,3 @@ def average_backlog_size(fact)
   end
   { average_backlog_size: issues.empty? ? 0 : issues.sum.to_f / issues.size }
 end
-
-def average_backlog_size_props
-  %w[average_backlog_size]
-end
--- a/judges/quality-of-service/average_backlog_size.rb
+++ b/judges/quality-of-service/average_backlog_size.rb
@@ -27,3 +27,7 @@ def average_backlog_size(fact)
   end
   { average_backlog_size: issues.empty? ? 0 : issues.sum.to_f / issues.size }
 end
+
+def average_backlog_size_props
+  %w[average_backlog_size]
+end
--- a/judges/quality-of-service/average_build_success_rate.rb
+++ b/judges/quality-of-service/average_build_success_rate.rb
@@ -40,7 +40,3 @@ def average_build_success_rate(fact)
     average_build_mttr: ttrs.any? ? ttrs.sum / ttrs.size : 0
   }
 end
-
-def average_build_success_rate_props
-  %w[average_build_success_rate average_build_duration average_build_mttr]
-end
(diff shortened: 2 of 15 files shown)


Yegorov:
@yegor256 please check

judges/quality-of-service/average_backlog_size.rb
```diff
@@ -27,3 +27,7 @@ def average_backlog_size(fact)
   end
   { average_backlog_size: issues.empty? ? 0 : issues.sum.to_f / issues.size }
 end
+
```
yegor256, line 30: @Yegorov this solution looks like a hack, which will definitely lead to errors in the future: we duplication information here. Maybe it's enough to rely on the Ruby file name? In each of our `average_*` scanners we produce a property that is named exactly as the file.

yegor256 (changes requested):
I suggest we rely on .rb file names, this should be enough

judges/quality-of-service/average_backlog_size.rb
Yegorov: @yegor256 what to do in this case: function called `average_hoc_size` but the added property is called `average_pull_hoc_size`? Can we painlessly rename function name to `average_pull_hoc_size`?

```ruby
def average_hoc_size(fact)
# ...
  {
    average_pull_hoc_size: hocs.empty? ? 0 : hocs.sum.to_f / hocs.size,
    average_pull_files_size: files.empty? ? 0 : files.sum.to_f / files.size
  }
end
```
[source](https://github.com/zerocracy/judges-action/blob/26365ddc0bd42f27f24eea5f19a597bd3f27f668/judges/quality-of-service/average_hoc_size.rb#L15C5-L15C21)

Yegorov: @yegor256 ok, it's probably better to add these test to another PR.

yegor256:
@rultor merge

0crat:
@yegorov Hey there, coding champ! 🚀 Great job on your contribution! You've snagged +7 points for this one: +16 for showing up, +7 for those 144 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) you cranked out (nice work!), but we had to subtract 16 because there wasn't a code review this time. Remember, reviews are super important – they can save you from point deductions and help improve your code. Plus, if you can bump up your [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) to 200+, you'll avoid that 8-point deduction next time. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the code flowing, and let's aim for even more points on your next contribution!

PR #667, opened 2025-05-31
#659: implement pull-was-merged judge

### LINKED ISSUE #659: `issue-was-assigned` facts not generated for assigned issues
When someone is assigned to an issue, we should catch this event in `github-events` and create a new fact with `what=issue-was-assigned`. We should set `who` to the person who was assigned. Also, let's set `assigner` to the person who did this.

Yegorov:
Test fail, because this https://github.com/zerocracy/fbe/pull/190 PR need merge before
Closes #659

### DIFF
--- a/judges/assignee-was-attached/assignee-was-attached.rb
+++ b/judges/assignee-was-attached/assignee-was-attached.rb
@@ -10,16 +10,17 @@ require 'fbe/octo'
 
 Fbe.conclude do
   quota_aware
-  on "(and (eq where 'github') (eq what 'issue-was-opened'))"
+  on "(and
+        (eq where 'github')
+        (eq what 'issue-was-opened')
+        (empty
+          (and
+            (eq where 'github')
+            (eq repository $repository)
+            (eq issue $issue)
+            (eq what 'issue-was-assigned'))))"
 
   consider do |f|
-    next if Fbe.fb.query(
-      "(and
-         (eq where 'github')
-         (eq repository #{f.repository})
-         (eq issue #{f.issue})
-         (eq what 'issue-was-assigned'))"
-    ).each.to_a.first
     Fbe.octo.issue_events(
       Fbe.octo.repo_name_by_id(f.repository),
       f.issue
--- a/judges/assignee-was-attached/assignee-was-attached.rb
+++ b/judges/assignee-was-attached/assignee-was-attached.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 Zerocracy
+# SPDX-License-Identifier: MIT
+
+# Judge that monitors open issue and create missing issue-was-assigned facts
+
+require 'fbe/conclude'
+require 'fbe/octo'
+
+Fbe.conclude do
+  quota_aware
+  on "(and (eq where 'github') (eq what 'issue-was-opened'))"
+
+  consider do |f|
+    next if Fbe.fb.query(
+      "(and
+         (eq where 'github')
+         (eq repository #{f.repository})
+         (eq issue #{f.issue})
+         (eq what 'issue-was-closed'))"
+    ).each.to_a.first
+    Fbe.octo.issue_events(
+      Fbe.octo.repo_name_by_id(f.repository),
+      f.issue
+    ).select { _1[:event] == 'assigned' }.each do |event|
+      fact =
+        Fbe.if_absent do |fact|
+          fact.where = f.where
+          fact.what = 'issue-was-assigned'
+          fact.repository = f.repository
+          fact.issue = f.issue
+          fact.issue_event_id = event[:id]
+        end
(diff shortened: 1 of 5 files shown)


Yegorov:
@yegor256 please check

judges/assignee-was-attached/assignee-was-attached.rb
```diff
@@ -0,0 +1,41 @@
...
+
+require 'fbe/conclude'
+require 'fbe/octo'
+
+Fbe.conclude do
+  quota_aware
+  on "(and (eq where 'github') (eq what 'issue-was-opened'))"
+
+  consider do |f|
+    next if Fbe.fb.query(
```
yegor256, line 16: @Yegorov we don't need to check this. Even if the issue is already closed, we are interested to know who it was assigned to. In order to reward.

judges/assignee-was-attached/assignee-was-attached.rb
```diff
@@ -0,0 +1,41 @@
...
+      "(and
+         (eq where 'github')
+         (eq repository #{f.repository})
+         (eq issue #{f.issue})
+         (eq what 'issue-was-closed'))"
+    ).each.to_a.first
+    Fbe.octo.issue_events(
```
yegor256, line 23: @Yegorov it seems, we will go to GitHub API for every `issue-was-opened` fact, even if it already has `issue-was-assigned`. Maybe you've made a typo above?

Yegorov: @yegor256 yes of course, because you can assign multiple users or reassign to another user

judges/assignee-was-attached/assignee-was-attached.rb
Yegorov: @yegor256 We need to fix it so that if a issue is already assigned, we don't check in GitHub API anymore?

judges/assignee-was-attached/assignee-was-attached.rb
Yegorov: @yegor256 Can issue be assigned to multiple users? Or just only one? And not been reassigned?

yegor256: @Yegorov indeed, it can be assigned many times. We don't care about all events. We are only interested in who it's assigned to, at this very moment.

judges/assignee-was-attached/assignee-was-attached.rb
yegor256: @Yegorov the marker is the presence of `issue-was-assigned`. If it's present, we should not check again. Later, when we register `issue-was-closed` fact, we should do an extra check -- who it is assigned to -- and replace the `issue-was-assigned` fact with a new one (if assignee is changed).

judges/assignee-was-attached/assignee-was-attached.rb
Yegorov: @yegor256 ok, thanks, fixed.

judges/assignee-was-attached/assignee-was-attached.rb
Yegorov: @yegor256 thanks, fixed.

Yegorov:
@yegor256 please check

judges/assignee-was-attached/assignee-was-attached.rb
```diff
@@ -0,0 +1,41 @@
...
+  quota_aware
+  on "(and (eq where 'github') (eq what 'issue-was-opened'))"
+
+  consider do |f|
+    next if Fbe.fb.query(
+      "(and
+         (eq where 'github')
+         (eq repository #{f.repository})
+         (eq issue #{f.issue})
+         (eq what 'issue-was-assigned'))"
```
yegor256, line 21: you can do better:

```
Fbe.conclude do
  quota_aware
  on "(and 
    (eq where 'github') 
    (eq what 'issue-was-opened'))
    (empty 
      (and
        (eq where 'github')
        (eq repository $repository)
        (eq issue $issue)
        (eq what 'issue-was-assigned'))))"
```

This will work faster, because will use internal factbase indexes.

yegor256 (changes requested):
@Yegorov now it's much better! See one suggestion above

judges/assignee-was-attached/assignee-was-attached.rb
```diff
@@ -0,0 +1,49 @@
...
+        (eq what 'issue-was-opened')
+        (empty
+          (and
+            (eq where 'github')
+            (eq repository $repository)
+            (eq issue $issue)
+            (eq what 'issue-was-assigned'))))"
+
+  consider do |f|
+    next if Fbe.fb.query(
```
yegor256, line 24: @Yegorov no need to check again, the query above already guarantees that you get proper facts

Yegorov:
@yegor256 please check

yegor256 (approved):

PR #672, opened 2025-06-03
#660: implement find-missing-open-issues judge

### LINKED ISSUE #660: `issue-was-opened` facts missing for some existing issues on GitHub
This offen happens: the `issue-was-closed` (or any other with the `issue` property) fact exists, while the `issue-was-opened` is missing. Let's create a new judge that will scan the factbase and find such situations. Then, it will check whether issues/pulls are in GitHub (via API). If they are present, the judge will add add either `issue-was-opened` or `pull-was-opened` fact to the factbase (with all necessary properties).

Yegorov:
Test fail, because this https://github.com/zerocracy/fbe/pull/193 PR need merge before
Closes #660

### DIFF
--- a/judges/find-missing-open-issues/find-missing-open-issues.rb
+++ b/judges/find-missing-open-issues/find-missing-open-issues.rb
@@ -11,41 +11,51 @@ require 'fbe/conclude'
 require 'fbe/issue'
 require 'fbe/who'
 
-%w[issue pull].each do |type|
-  what = "#{type}-was-opened"
-  Fbe.conclude do
-    quota_aware
-    on "(and
-          (eq where 'github')
-          (exists repository)
-          (exists issue)
-          (not (eq what '#{what}'))
-          (matches what '#{type}')
-          (empty
-            (and
-              (eq where 'github')
-              (eq repository $repository)
-              (eq issue $issue)
-              (eq what '#{what}'))))"
+Fbe.conclude do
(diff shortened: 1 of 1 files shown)


Yegorov:
@yegor256 please check

judges/find-missing-open-issues/find-missing-open-issues.rb
```diff
@@ -0,0 +1,52 @@
...
+%w[issue pull].each do |type|
+  what = "#{type}-was-opened"
+  Fbe.conclude do
+    quota_aware
+    on "(and
+          (eq where 'github')
+          (exists repository)
+          (exists issue)
+          (not (eq what '#{what}'))
+          (matches what '#{type}')
```
yegor256, line 23: @Yegorov this predicate will make this query extremely slow. Imagine, 10K records and all of them will need a regular expression match. Indexes won't be used.

yegor256 (changes requested):
@Yegorov looks correct, but won't work with large factbases :( see above

judges/find-missing-open-issues/find-missing-open-issues.rb
Yegorov: @yegor256 ok, but another solution is to add this check to the `consider` block. Or maybe it can be done in some other way?

yegor256:
@rultor merge

Issue #956, opened 2025-08-31
`entries/` tests slow increasing execution time, implement parallel execution for `make entries.sh`


vbvictor:
Currently, takes 30s to run `make entries` and this number will grow linearly if we add more tests.
It would be great to speed-up tests by running them in parallel.

yegor256:
@vbvictor indeed

0crat:
@vbvictor 🎉 Fantastic bug report! You've earned +12 points and helped improve our project's quality. Keep your eyes peeled for anything that [doesn't seem quite right](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) in the repository. Remember, we [welcome all bug reports](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html), so don't hesitate to [let us know](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) if you spot something. Your keen observations make our software better!

PR #960, opened 2025-09-01
#956: add parallel test execution for 'entries tests'

vbvictor:
This PR speed up test execution 2.5 times (from `30sec` to `12sec`) and eliminating further slowness by running tests in parallel.

Closes https://github.com/zerocracy/judges-action/issues/956.

## Summary by CodeRabbit

* **Tests**
  * Test suite now runs in parallel for faster execution, with per-test directories/logs and clear pass/fail output; the run halts on first failure.
* **CI**
  * GitHub Actions workflow installs GNU parallel before running the test job.
* **Documentation**
  * README prerequisites updated to list GNU parallel alongside GNU make.

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -21,5 +21,5 @@ jobs:
           ruby-version: 3.3
           bundler-cache: true
       - run: gem install rubocop:1.63.5
-      - run: sudo apt install parallel
+      - run: sudo apt-get update && sudo apt-get install -y parallel
       - run: make
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -21,4 +21,5 @@ jobs:
           ruby-version: 3.3
           bundler-cache: true
       - run: gem install rubocop:1.63.5
+      - run: sudo apt install parallel
       - run: make
--- a/makes/entries.sh
+++ b/makes/entries.sh
@@ -7,8 +7,10 @@ set -e -o pipefail
 base=$(realpath "$(dirname "$0")/..")
 
 mkdir -p "${base}/target/entries-logs"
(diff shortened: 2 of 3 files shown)


makes/entries.sh
```diff
@@ -7,13 +7,46 @@ set -e -o pipefail
...
+done < <( cd "${base}/entries" && find . -type f -name '*.sh' -print0 )
+
+# Wait for all tests to complete and collect results
+exit_code=0
+for i in "${!pids[@]}"; do
+    if ! wait "${pids[$i]}"; then
+        failed_tests+=("${test_names[$i]}")
+        exit_code=1
+    fi
+done
```
yegor256, line 41: @vbvictor I believe, you can do the same, but with the help of [GNU Parallel](https://www.gnu.org/software/parallel/). The code will be much shorter.

yegor256 (changes requested):

yegor256:
@vbvictor looks very good, thanks! See one comment above.

makes/entries.sh
yegor256: @vbvictor yes, it's fine. Just don't forget to document it in the README.

.github/workflows/make.yml
```diff
@@ -21,4 +21,5 @@ jobs:
           ruby-version: 3.3
           bundler-cache: true
       - run: gem install rubocop:1.63.5
+      - run: sudo apt install parallel
```
vbvictor, line 24:
```suggestion
      - run: sudo apt-get update && sudo apt-get install -y parallel
```

yegor256 (approved):

yegor256:
@vbvictor thanks!

PR #945, opened 2025-08-28
added an example of pure bash coverage

ostvld:
added an example of pure bash coverage

## Summary by CodeRabbit

- New Features
  - Added a script to generate coverage reports for the project's main workflow with verbose output enabled.

- Refactor
  - Enhanced runtime tracing/logging so verbose command tracing is available with standardized trace prefixes when enabled.

- Chores
  - Integrated an external coverage tool as a submodule to support automated report generation.

PR #989, opened 2025-09-04
added an example of pure bash coverage

ostvld:
[added an example of pure bash coverage](https://github.com/zerocracy/judges-action/pull/945)

## Summary by CodeRabbit

- **New Features**
  - Added a built-in coverage report runner and a wrapper to invoke the bundled coverage generator.
  - Introduced a verbose/debug mode that prints annotated debug traces (file:line) when enabled.

- **Chores**
  - Added and updated an external coverage tool as a Git submodule; initialize/update submodules after checkout to fetch it.

- **Notes**
  - No behavioral changes unless verbose mode is enabled.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "calc_coverage/pure_bash_coverage"]
+	path = calc_coverage/pure_bash_coverage
+	url = git@github.com:ostvld/pure_bash_coverage.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "calc_coverage/pure_bash_coverage"]
 	path = calc_coverage/pure_bash_coverage
-	url = git@github.com:ostvld/pure_bash_coverage.git
+	url = https://github.com/ostvld/pure_bash_coverage.git 
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,5 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 Zerocracy
+# SPDX-License-Identifier: MIT
 [submodule "calc_coverage/pure_bash_coverage"]
 	path = calc_coverage/pure_bash_coverage
 	url = https://github.com/ostvld/pure_bash_coverage.git 
--- a/calc_coverage/run_cov.sh
+++ b/calc_coverage/run_cov.sh
@@ -2,4 +2,6 @@
 set -euo pipefail
 export INPUT_VERBOSE="true"
 echo "${INPUT_VERBOSE}"
-./pure_bash_coverage/generate_coverage_report.sh ../entry.sh
+DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
+${DIR}/pure_bash_coverage/generate_coverage_report.sh ${DIR}/../entry.sh
+
--- a/calc_coverage/run_cov.sh
+++ b/calc_coverage/run_cov.sh
@@ -1,2 +1,5 @@
-# !bin/bash
-./pure_bash_coverage/generate_coverage_report.sh ../entry.sh
+#!/usr/bin/env bash
+set -euo pipefail
+export INPUT_VERBOSE="true"
+echo "${INPUT_VERBOSE}"
+../pure_bash_coverage/generate_coverage_report.sh ../entry.sh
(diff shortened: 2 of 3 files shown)


entry.sh
```diff
@@ -2,8 +2,14 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024-2025 Zerocracy
 # SPDX-License-Identifier: MIT
 
+export PS4='+${BASH_SOURCE}:${LINENO}:'
+echo "INPUT_VERBOSE: ${INPUT_VERBOSE}"
```
yegor256, line 6: @ostvld this doesn't look nice, maybe you left it here by mistake?

calc_coverage/run_cov.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 Zerocracy
+# SPDX-License-Identifier: MIT
+set -euo pipefail
+export INPUT_VERBOSE="true"
+echo "${INPUT_VERBOSE}"
+DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
+"${DIR}/pure_bash_coverage/generate_coverage_report.sh" "${DIR}/../entry.sh"
```
yegor256, line 8: @ostvld what will it generate? Will the file be suitable for codecov.io?

yegor256 (changes requested):
@ostvld looks very interesting! Please, address a few concerns above.

.gitmodules
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 Zerocracy
```
vbvictor, line 1: I think this file should be added to `Reuse.toml` instead. It's odd to see copyright notice in configs

.gitmodules
```diff
@@ -0,0 +1,5 @@
...
+# SPDX-License-Identifier: MIT
+[submodule "calc_coverage/pure_bash_coverage"]
+	path = calc_coverage/pure_bash_coverage
+	url = https://github.com/ostvld/pure_bash_coverage.git
```
vbvictor, line 5: Could you add to installation instructions in `Readme.md` of how to install submodules. I think they are not installed automatically.

calc_coverage/run_cov.sh
kreinba, line 8: Nothing in this repository invokes run_cov.sh: no Makefile target, no workflow under .github/workflows, no docs entry. The make.yml workflow also uses actions/checkout@v5 without submodules: recursive, so pure_bash_coverage is never cloned during CI even if a job were to call this script. The example as added is therefore dead code that cannot run. Either wire it into a workflow or Make target that opts into submodule checkout, or drop calc_coverage/ entirely and keep only the verbose-mode change to entry.sh under a separate, scoped PR.

entry.sh
```diff
@@ -2,8 +2,14 @@
...
 set -e -o pipefail
 
+if [ "${INPUT_VERBOSE}" == 'true' ]; then
```
kreinba, line 9: Master has moved on since this branch was opened: entry.sh now calls ::add-mask:: on INPUT_TOKEN and INPUT_GITHUB-TOKEN before any tracing is enabled, precisely so that set -x cannot echo secrets into Actions logs. This diff places set -x at line 9-11, above where those masks live on master, so a rebase that keeps this hunk as-is would re-open the leak. Rebase against current master and move the set -x block to after the ::add-mask:: lines, or drop the relocation and leave the existing later-in-file set -x in place.

kreinba (changes requested):
The submodule added by this branch is never fetched in CI, since .github/workflows/make.yml uses actions/checkout@v5 without submodules: recursive, and no workflow, Makefile target, or docs reference calc_coverage/run_cov.sh. The example therefore cannot run anywhere, and CI staying green confirms the wrapper is unreachable. Beyond that, the branch is nine months stale and entry.sh on master has since been restructured to call ::add-mask:: on INPUT_TOKEN and INPUT_GITHUB-TOKEN before enabling tracing; this diff hoists set -x above where those masks now live and would silently undo that protection on rebase. Two inline comments below name those two concerns. The smaller suggestions raised by earlier reviewers (Russian text in run_cov.sh, duplicate set -x block, codecov format, README install steps, REUSE.toml placement, trailing whitespace) still apply.

yegor256:
Not merging: this adds a git submodule pointing at a personal repository into an action that ships as a Docker image, and it moves unconditional `echo` and `PS4` output into `entry.sh`. If the coverage report is worth having, vendor the script under `makes/` and keep `entry.sh` quiet.

yegor256:
The `rake` CI check is failing on this PR and the CI last ran over 3 months ago (June 1). Could you please rebase onto the latest master and fix the failing test before this can be merged?

---

PR #1059, opened 2025-09-22
Fix fetch release info

uchitsa:
## Summary by CodeRabbit

* Bug Fixes
  - Improves reliability of release stats by safely handling missing comparison data.
  - Ensures commit counts, file change totals, and last commit details are calculated accurately when available.
  - Adds warning logs when comparison data is unavailable without disrupting other information.

* Tests
  - Adds comprehensive tests covering successful comparisons, missing data, fallback behaviors, and empty results to ensure consistent release info handling.


judges/github-events/github-events.rb
```diff
@@ -81,8 +81,6 @@ def self.fetch_release_info(fact, repo)
     rescue Octokit::NotFound
       $loog.info("Compare API returned 404 for #{repo} between #{tag} and #{fact.tag}")
     end
-    $loog.debug("The repository ##{fact.repository} has this: #{info.inspect}")
-    info
   end
```
kreinba, line 82: Removing `info` from the end of the method breaks the return value. `Fbe.octo.compare(...).then { |json| ... }` returns the value of the block, and the block's last expression is `info[:last_commit] = json[:commits].first[:sha]`, which is the SHA string, not the `info` hash. The `rescue Octokit::NotFound` branch now returns whatever `$loog.info` returns. Callers such as `fill_fact(fact, fetch_release_info(fact, rname))` at line 252 expect a hash with `:commits`/`:hoc`/`:last_commit`. Restore the trailing `info` (the debug log can stay removed if the goal is to cut noise) so the method ends on the hash in every branch.

test/judges/test-github-events.rb
```diff
@@ -2524,6 +2526,147 @@ def test_write_supervision_log_if_raise_error
     end
   end
 
+  if defined?(Fbe::GithubGraph)
+   def test_fetch_release_info_successful_comparison
```
kreinba, line 2751: These tests call `Fbe::GithubGraph.fetch_release_info`, but `fetch_release_info` is not defined on `Fbe::GithubGraph`. In `judges/github-events/github-events.rb` it is declared with `def self.fetch_release_info` inside the `Fbe.iterate do ... end` block, i.e. as a singleton method on the iteration block's `self`, not on the `Fbe::GithubGraph` module. The `if defined?(Fbe::GithubGraph)` guard passes (the constant exists), so the body runs and `NoMethodError` blows up the suite — which is what the red `rake` job is reporting. To exercise this method, refactor it out of the `Fbe.iterate` block into a module of its own (or move the logic into `Fbe::GithubGraph` upstream) and have the production code call the extracted method; otherwise these tests cannot reach it.

kreinba (changes requested):
The change drops the trailing `info` (and the debug log) from `fetch_release_info`, which silently breaks the method's return contract: on the success path the method now returns the value of the last assignment inside the `.then` block — the SHA string from `info[:last_commit] = ...` — instead of the `info` hash, and on the `Octokit::NotFound` path it returns whatever `$loog.info` returns rather than `info`. Every caller, including `fill_fact(fact, fetch_release_info(fact, rname))` at line 252, expects a hash, so the production code path is broken in two of three branches. The added tests in `test/judges/test-github-events.rb` call `Fbe::GithubGraph.fetch_release_info`, but `fetch_release_info` is defined as a singleton on the block passed to `Fbe.iterate` in this file — it is not a method of `Fbe::GithubGraph` — so the tests would `NoMethodError` even if the production return value were intact. CI is red on `rake` for that reason. Two inline comments below; the PR also needs a `Fixes #N` link in the body.

VasilevNStas:
@yegor256 @edmoffo This PR is 9 months old, has CHANGES_REQUESTED from kreinba, and the codebase has evolved significantly since. The diff no longer applies cleanly. Suggest closing as stale and reopening a fresh issue if the problem is still relevant.

edmoffo:
@VasilevNStas Agreed on the assessment: 253 days old, draft, no longer mergeable against current master, and CHANGES_REQUESTED still open. Close-as-stale needs a maintainer; flagging.

VasilevNStas:
@yegor256 This PR is 253 days old, a draft, no longer mergeable against master, and CHANGES_REQUESTED is still open. Please close as stale. @edmoffo flagged the same.

kreinba:
@VasilevNStas The CHANGES_REQUESTED review is still standing because the diff breaks the return contract of `fetch_release_info` (the success path now returns the SHA string from the last assignment instead of the `info` hash, and the `Octokit::NotFound` path returns the value of `$loog.info`), and the added test wires assertions to the wrong receiver. Until new commits address those two points, the staleness assessment matches the review state.

uchitsa:
Good catches, thanks.

2 right moments:
- restored the trailing info so fetch_release_info returns the hash on the success path and the Octokit::NotFound path.
- extracted fetch_release_info out of the Fbe.iterate block into its own module.

PR #1145, opened 2025-10-20
Add milestone-was-set judge

kodsurfer:
## Summary by CodeRabbit

* **New Features**
  * Added milestone event tracking: detects GitHub milestone creations and closures, records title, description, creator, timestamps, optional deadline, and human-readable details; filters by update time, skips duplicates, and summarizes per-repository scan totals.
* **Tests**
  * Added tests validating milestone event creation, skipping already-recorded milestones, and handling milestones with and without deadlines.


judges/milestone-was-set/milestone-was-set.rb
```diff
@@ -0,0 +1,115 @@
...
+    id = nil
+    total = 0
+    detected = 0
+    rstart = Time.now
+    milestones = Fbe.octo.milestones(rname, state: 'all')
+    milestones.each do |milestone|
+      total += 1
+      updated_time = Time.parse(milestone[:updated_at].iso8601).to_i
+      id = updated_time
+      id = [id, updated_time].compact.max
```
kreinba, line 82: These two lines together leave `id` equal to the current iteration's `updated_time` — `id = updated_time` overwrites whatever the previous loop wrote, and `[id, updated_time].compact.max` then takes the max of two copies of the same number. The `compact.max` line is dead. Drop the `id = updated_time` line and let `id = [id, updated_time].compact.max` accumulate across iterations, so the returned watermark is the maximum `updated_at` seen on this page rather than the value from whichever milestone happened to be iterated last.

judges/milestone-was-set/milestone-was-set.rb
```diff
@@ -0,0 +1,115 @@
...
+      "Milestone ##{json[:number]} (#{json[:title]}) in #{repo} ignored"
+    )
+    raise Factbase::Rollback
+  end
+
+  def self.fill_up_milestone(fact, json, repo_id, rname)
+    fact.when = Time.parse(json[:created_at].iso8601)
+    fact.milestone = json[:number]
+    fact.repository = repo_id
+    fact.who = json[:creator][:id].to_i if json[:creator]
```
kreinba, line 33: Guarding `fact.who = ...` on `json[:creator]` but then calling `Fbe.who(fact)` inside the `details` string (lines 41 and 50) and the debug log (lines 44 and 53) raises when the API returns a milestone whose creator is nil — which happens on deleted or suspended accounts. Either drop the guard and let a nil creator crash earlier with a clearer trace, or have `fill_up_milestone` fall through to `skip_milestone(json, rname)` when `json[:creator]` is nil, the same way the `else` branch on line 56 handles unknown states.

kreinba (changes requested):
The judge has two correctness defects on its main path and CI is red on `make` and `rake`. First, the per-iteration `id = updated_time` immediately followed by `id = [id, updated_time].compact.max` makes the `compact.max` a no-op — both operands are the same value — so the returned watermark is the last-iterated milestone's `updated_at`, not the maximum across the page, which silently regresses the cursor whenever the API returns milestones in any order other than ascending `updated_at`. Second, `fill_up_milestone` only sets `fact.who` when `json[:creator]` is truthy but then unconditionally interpolates `Fbe.who(fact)` into the `details` string and the debug log; a milestone whose creator the API has nulled out (deleted account, suspended user) will raise instead of being recorded. Inline comments below. Please also drop `fact.deadline = ... : nil` in favour of the `fact.X = ... if ...` pattern the sibling judges use; assigning `nil` to a factbase property is not a documented operation.

yegor256:
We are closing this pull request, because it has merge conflicts with `master` and has been open for more than a month. Please, forgive us for this: it says nothing about the quality of your work, it's just housekeeping.

If the change still matters to you, please submit a new pull request. It may be very similar to this one, but it has to take into account everything that landed in `master` since then.

Thank you for contributing, and sorry again for the inconvenience!

PR #1333, opened 2026-04-23
#1330: rescue Octokit::Forbidden in all user/repo/issue lookup sites

vitaly-andr:
## Summary

Closes #1330 (primary), #1331 (side-finding), #1332 (tagalong).

Three related fixes landing together, discovered as a domino chain
while writing per-site reproduction tests for the primary bug.

## Note after rebase onto updated master

This branch was rebased onto `origin/master` after #1325 landed
(thanks to @igor-belousov for the independent fix to
`label-was-attached.rb` from #645). After rebase:

- Our `next` → `next issue` change in `label-was-attached.rb` is now
  **no-op** — #1325 already applied it. Credit goes there.
- Scope of issue #1331 narrows to the remaining **2 judges**
  (`issue-was-opened.rb`, `pull-was-opened.rb`) — both
  `Fbe.conclude + draw` pattern needing `throw :rollback`.
- Our new 403 test in `test-label-was-attached.rb` is
  **complementary** to #1325's 404 test — same rescue block, different
  exception trigger, matching recovery assertion (`stale: 'issue'`).
- Also picks up the recent `minitest` / `minitest-mock` version bumps
  from #1327 in `Gemfile.lock` (no action needed — just fresh deps).

See the **Status update** in issue #1331 body for scope detail.

### Code changes (14 files)

- `lib/nick_of.rb` — helper, +split rescue Forbidden (closes #1330)
- 13 judge files — +split rescue Forbidden, 1 to 2 Pattern A sites
  each (closes #1330)
- 2 of the 13 ALSO get rescue-exit fix (closes #1331):
  - `issue-was-opened.rb` — `next` → `throw :rollback`
  - `pull-was-opened.rb` — `next issue` → `throw :rollback`
  - (`label-was-attached.rb` was also in this category originally,
    but its fix landed separately via #1325 before we rebased)
- `is-human-or-robot.rb` ALSO gets `require 'fbe/issue'` (closes #1332)

### Tests (17 files)

**For #1330 — 14 per-site reproduction tests:**
- 6 cherry-picked from `feature/forbidden-rescue-verification` and
  flipped to post-fix assertions
- 8 new test methods in canonical test files on master
- 1 new test file (`test-fix-missing-branch.rb`)
- 1 structural regression test (`test/test_pattern_a_coverage.rb`)
  enforcing every Pattern A rescue block has a matching
  `rescue Octokit::Forbidden => e` — prevents future uniformity
  regression

[...]

PR #1338, opened 2026-04-28
Fix #1337: rescue Forbidden no longer marks facts stale (transient errors will retry)

### LINKED ISSUE #1337: Regression from #1336: transient 403 silently tombstones live data
Following up on #1333 with an unintended consequence I want to flag.

That PR added `rescue Octokit::Forbidden` next to the existing `rescue NotFound` in Pattern A judges, so a single inaccessible repo or issue would no longer crash the whole batch. The new branch copied the NotFound branch exactly, including the call to `Jp.issue_was_lost(...)` and `f.stale = 'who'` for user lookups.

The two errors are not the same. 404 NotFound is permanent — the resource is gone, tombstone is right. 403 Forbidden is almost always transient: rate limit, SSO config change, abuse rate limit, scope mismatch, temporary API degradation. The resource still exists; we just can't see it right now.

Treating them the same way means a transient 403 silently marks live issues and contributors as lost. Tombstone has no `unbury!`, stale facts are filtered by downstream queries (`(absent stale)`), so the data does not come back next cycle — it is gone for good. A 30-minute GitHub rate limit during a run can permanently strip valid contributor and issue facts from the published dashboard.

Apologies for missing this in review. The PR also added 12 tests that pin the wrong semantics:
test/judges/test-issue-was-opened.rb:27, test/judges/test-issue-was-closed.rb:188, test/judges/test-issue-was-assigned.rb:101, test/judges/test-issue-was-unassigned.rb:140, test/judges/test-label-was-attached.rb:65, test/judges/test-find-all-issues.rb:327, test/judges/test-find-missing-issues.rb:43, test/judges/test-fix-missing-branch.rb:27, test/judges/test-add-review-comments.rb:103 and :125, test/judges/test-pull-was-opened.rb:26, test/judges/test-pull-was-merged.rb:335, test/judges/test-is-human-or-robot.rb:79, test/judges/test-who-has-name.rb:103.

Fix is coming as a follow-up PR. The rescue Forbidden branches will skip the iteration without tombstoning or staling; the next cycle will retry naturally. `test-revive-user.rb` already has the correct semantic (stale user stays stale on 403) and will not be touched.

Note on already-corrupted data: this PR will only stop further leakage. Already-tombstoned facts can be scoped via `issue-was-lost` facts created since the #1333 merge (a few days, likely a few hundred at most). Happy to propose a recovery script in a separate follow-up if useful.

vitaly-andr:
Closes #1337.

#1333 added `rescue Octokit::Forbidden` next to the existing NotFound branch and copied the same body — `Jp.issue_was_lost(...)` or `f.stale = 'repository'`. That turned transient 403s into permanent tombstones on live data. This PR removes the tombstoning from the Forbidden branches; the next cycle retries naturally.

Scope is issue and repository sites only: 11 production files, 11 test assertions flipped. User-side rescues (`is-human-or-robot`, `who-has-name`, `eliminate-ghosts`, `who-is-alive`, `lib/nick_of.rb`) are untouched here because `revive-user` already auto-removes `stale='who'` on the next successful lookup.

The 12 tests that pinned the wrong semantics now assert the fact stays live (no stale, no tombstone). Batch survival (the goal of #1333) is preserved — every `rescue Forbidden` block still catches the exception, just without the permanent marking.

Already-corrupted facts from this regression window are not auto-restored by this PR. Happy to add a recovery script as a separate follow-up if you want.

### DIFF
--- a/judges/add-review-comments/add-review-comments.rb
+++ b/judges/add-review-comments/add-review-comments.rb
@@ -27,8 +27,10 @@ Fbe.consider(
       f.stale = 'repository'
       next
     rescue Octokit::Forbidden => e
-      $loog.warn("[#{$judge}] Access forbidden to repository #{f.repository}: #{e.class}: #{e.message}")
-      f.stale = 'repository'
+      $loog.warn(
+        "[#{$judge}] Access forbidden to repository #{f.repository} " \
+        "(transient, will retry next cycle): #{e.class}: #{e.message}"
+      )
       next
     end
   json =
@@ -39,8 +41,10 @@ Fbe.consider(
       Jp.issue_was_lost(f.where, f.repository, f.issue)
       next
     rescue Octokit::Forbidden => e
-      $loog.warn("[#{$judge}] Access forbidden to issue ##{f.issue} in #{repo}: #{e.class}: #{e.message}")
-      Jp.issue_was_lost(f.where, f.repository, f.issue)
+      $loog.warn(
+        "[#{$judge}] Access forbidden to issue ##{f.issue} in #{repo} " \
+        "(transient, will retry next cycle): #{e.class}: #{e.message}"
+      )
       next
     end
   c = json[:review_comments]
--- a/judges/find-all-issues/find-all-issues.rb
+++ b/judges/find-all-issues/find-all-issues.rb
@@ -37,10 +37,9 @@ require_relative '../../lib/issue_was_lost'
           next 0
         rescue Octokit::Forbidden => e
           $loog.warn(
-            "[#{$judge}] Access forbidden to #{type} ##{issue}, " \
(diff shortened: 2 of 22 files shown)


0crat:
@vitaly-andr Thanks for the contribution! You've earned **+24 points** for this: +24 as a basis, +9 for the 122 hits-of-code that you wrote, then -9 to stay within the 16-point cap for code volume as outlined in our policy. This solid contribution shows great productivity while maintaining quality standards. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #1346, opened 2026-04-29
#1345: Fix metric correctness in QoS and QoD judges (forward-only)

### LINKED ISSUE #1345: Inaccurate metric calculation in quality-of-service and quantity-of-deliverables components
@yegor256 — heads-up. While auditing `quality-of-service` and `quantity-of-deliverables` judges I found five metric-correctness issues. They produce wrong values that get persisted into facts (with `avoid_duplicate: true` in `Jp.incremate`, so they're locked in once written).

The dashboard chart in pages-action's `qo-section.xsl` only renders `n_composite` by default — every `n_<prefix>_<suffix>` series is `hidden:true` because of the `contains(substring-after($n, 'n_'), '_')` rule on line 78. So none of these mismeasurements are visible on the public page unless someone toggles a hidden dataset in the chart legend, or consumes the factbase directly. **Practical impact today: low.** Filing in case any external consumer relies on the raw values, or in case the rendering logic ever changes.

## The findings

**1. `some_merged_pulls` counts all closed PRs, not merged**

`judges/quality-of-service/some_merged_pulls.rb:13` — query is `type:pr closed:RANGE` with no `is:merged` filter, so it returns total closed (merged + rejected). Companion `some_unmerged_pulls` is correctly filtered with `is:unmerged`, so a consumer can recover `actual_merged = some_merged_pulls - some_unmerged_pulls`, but the property name is misleading.

**2. `some_release_interval` skips in-window releases when the newest is too new**

`judges/quality-of-service/some_release_interval.rb:13` — `Fbe.octo.releases` returns newest-first; the loop does `break if published_at < since || published_at > when`. If the newest release is past `fact.when`, the very first iteration breaks and no in-window releases are counted. Should `next` for too-new and `break` only when below `since`.

**3. `some_release_hoc_size` and `some_release_commits_size` — same break-on-newest-first bug**

`judges/quality-of-service/some_release_hoc_size.rb:15` — same `break` pattern as #2. Backfilled historical slices that include older releases but where the repo also has newer releases will get empty `hoc/commits` arrays.

**4. `total_builds_ran` ignores slice end (`fact.when`)**

`judges/quantity-of-deliverables/total_builds_ran.rb:13` — query uses `created: >SINCE` without an upper bound. For backfilled historical slices the count grows over time as new builds are created in the world. A May-2024 slice computed today reports far more builds than the slice actually contains. Fix: `created: SINCE..WHEN` range syntax.

**5. `total_reviews_submitted` ignores slice end**

`judges/quantity-of-deliverables/total_reviews_submitted.rb:21` — counts reviews matching `submitted_at > fact.since`, no upper bound. Same backfill issue: old slices accumulate future reviews.

## A backfill question

Because of `avoid_duplicate: true` in `Jp.incremate`, fixing any of these forward only affects new slices. Existing facts retain their wrong values. Three options if you want correctness:

- Accept the trend discontinuity (old slices wrong, new ones correct)
- Add new properties (e.g. `some_truly_merged_pulls`) and leave old ones untouched
- Backfill — clear the affected properties and let `incremate` recompute

Ready to send PRs in any of those directions if you'd like the fixes. Otherwise this is just a flag.

vitaly-andr:
Per your "good catch. let's ignore all slices. only new should be fixed" reply on #1345 — fixes the five metric-correctness issues forward-only.

Forward-only is automatic: `Jp.incremate(..., avoid_duplicate: true)` already skips assignment for existing properties, so historical facts retain their (incorrect) values; only newly-computed slices use the corrected code.

## Fixes

- **`some_merged_pulls.rb`** — query missing `is:merged`; was counting all closed PRs (merged + rejected).
- **`some_release_interval.rb` / `some_release_hoc_size.rb`** — break-on-newest-first: Octokit returns releases newest-first, so an unconditional `break` on out-of-window killed the loop before any in-window release was seen. Switched to `next` for too-new, `break` only when below `since`. The shared loop is extracted into `Jp.in_window_releases` (`lib/in_window_releases.rb`).
- **`total_builds_ran.rb`** — search had no upper bound (`created: ">SINCE"`); now uses `created: SINCE..WHEN` range.
- **`total_reviews_submitted.rb`** — review predicate had no upper bound; added `&& <= fact.when`.

## Tests

- New unit test for `Jp.in_window_releases` covering the bug-fix path (newest release past window, in-window release below).
- `total_builds_ran` test stub URLs updated to range syntax.
- `total_reviews_submitted` test shifted `Time.stub` to 2025-10-05 so `Fbe::Graph::Fake`'s review dates fall within the window — this validates the upper bound is actually wired, rather than just asserting that everything ends up excluded.
- `fill_up_abandoned_facts_with_exists_when_and_absent_since`: `some_merged_pulls` expectation drops from `[2]` to `[1]`. The `[2]` was the bug-output (counting closed PRs); `[1]` matches the helper's existing `is:merged` stub which was correct all along.

Full `bundle exec rake test` + `bundle exec rubocop` clean.

Closes #1345

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -45,6 +45,7 @@ Elegant/GoodMethodName:
     - total_reviews_submitted
     - cover_qo
     - fill_fact_by_hash
+    - in_window_releases
     - issue_was_lost
     - nick_of
     - comments_info
--- a/judges/quality-of-service/some_merged_pulls.rb
+++ b/judges/quality-of-service/some_merged_pulls.rb
@@ -11,7 +11,7 @@ def some_merged_pulls(fact)
   rejected = []
   Fbe.unmask_repos do |repo|
     pulls << Fbe.octo.search_issues(
-      "repo:#{repo} type:pr closed:#{fact.since.utc.iso8601}..#{fact.when.utc.iso8601}"
+      "repo:#{repo} type:pr is:merged closed:#{fact.since.utc.iso8601}..#{fact.when.utc.iso8601}"
     )[:total_count]
     rejected << Fbe.octo.search_issues(
       "repo:#{repo} type:pr is:unmerged closed:#{fact.since.utc.iso8601}..#{fact.when.utc.iso8601}"
--- a/judges/quality-of-service/some_release_hoc_size.rb
+++ b/judges/quality-of-service/some_release_hoc_size.rb
@@ -5,14 +5,14 @@
 
 require 'fbe/octo'
 require 'fbe/unmask_repos'
+require_relative '../../lib/in_window_releases'
 
 def some_release_hoc_size(fact)
   grouped = {}
   hocs = []
   commits = []
   Fbe.unmask_repos do |repo|
-    Fbe.octo.releases(repo).each do |json|
-      break if json[:published_at] < fact.since || json[:published_at] > fact.when
+    Jp.in_window_releases(repo, fact.since, fact.when) do |json|
       (grouped[repo] ||= []) << json
     end
   end
--- a/judges/quality-of-service/some_release_interval.rb
+++ b/judges/quality-of-service/some_release_interval.rb
@@ -5,12 +5,12 @@
 
 require 'fbe/octo'
 require 'fbe/unmask_repos'
+require_relative '../../lib/in_window_releases'
 
 def some_release_interval(fact)
   dates = []
   Fbe.unmask_repos do |repo|
-    Fbe.octo.releases(repo).each do |json|
-      break if json[:published_at] < fact.since || json[:published_at] > fact.when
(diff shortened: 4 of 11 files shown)


vitaly-andr:
Closing as superseded by #1351.

For the record, since the diffs are very close: this PR was opened on Apr 29 ~4h after @yegor256's "good catch, only new should be fixed" reply on #1345, and addresses exactly that forward-only scope. It sat without review for three days while #1351 implemented the same fix — `some_merged_pulls.rb` and `total_builds_ran.rb` diffs here are byte-identical to what landed in #1351; the three release/review files use the same logic written inline rather than via the `Jp.in_window_releases` helper proposed here.

The helper extraction was the only delta and isn't worth a separate PR for two call sites. Closing the branch.

PR #1350, opened 2026-05-01
#1349: fix NoMethodError in fetch_workflows for check runs with nil app

### LINKED ISSUE #1349: `fetch_workflows` method fails with `NoMethodError` for check runs without associated app
## What Happens

`Jp.fetch_workflows` in `lib/pull_request.rb` line 50 calls `run[:app][:slug]`
without guarding against `run[:app]` being `nil`:

```ruby
next unless run[:app][:slug] == 'github-actions'
```

GitHub's Check Runs API returns `app: null` for required status checks that are
not associated with a GitHub App. When such a check run is present, calling
`nil[:slug]` raises `NoMethodError: undefined method '[]' for nil`.

This crashes `Jp.fetch_workflows` and in turn any judge that calls it
(e.g. `pull-was-merged.rb`, `github-events.rb`) on any repository that has
non-app check runs configured.

## What Should Happen

The code should use `run.dig(:app, :slug)` instead of `run[:app][:slug]`,
which safely returns `nil` when `app` is absent:

```ruby
next unless run.dig(:app, :slug) == 'github-actions'
```

igor-belousov:
Closes #1349.

## Summary
- `Jp.fetch_workflows` in `lib/pull_request.rb` called `run[:app][:slug]` directly, which raised `NoMethodError: undefined method '[]' for nil` for check runs whose `app` is `nil` (GitHub returns `app: null` for required status checks not associated with a GitHub App).
- Changed to `run.dig(:app, :slug)` so such check runs are silently skipped, as already intended by the surrounding `next unless ... == 'github-actions'` filter.

## Test plan
- Added `test/lib/test_pull_request.rb` covering three check runs in one response: one with `app: nil`, one with `app.slug: 'other-app'`, and one with `app.slug: 'github-actions'`. Asserts the helper returns `{succeeded_builds: 1, failed_builds: 0}` without raising.
- Verified the new test fails on master with the same `NoMethodError` from the issue, and passes after the fix.
- `bundle exec rake test` — 169 tests, 0 failures.
- `bundle exec rubocop lib/pull_request.rb test/lib/test_pull_request.rb` — clean.

### DIFF
--- a/lib/pull_request.rb
+++ b/lib/pull_request.rb
@@ -47,7 +47,7 @@ def Jp.fetch_workflows(pr, repo: nil)
   repo = pr.dig(:base, :repo, :full_name) if repo.nil?
   return {} if repo.nil?
   Fbe.octo.check_runs_for_ref(repo, pr.dig(:head, :sha))[:check_runs].each do |run|
-    next unless run[:app][:slug] == 'github-actions'
+    next unless run.dig(:app, :slug) == 'github-actions'
     workflow = Fbe.octo.workflow_run(repo, Fbe.octo.workflow_run_job(repo, run[:id])[:run_id])
     next unless workflow[:event] == 'pull_request'
     case workflow[:conclusion]
--- a/test/lib/test_pull_request.rb
+++ b/test/lib/test_pull_request.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+require_relative '../../lib/pull_request'
+require_relative '../test__helper'
+
+class TestPullRequest < Jp::Test
+  def test_fetch_workflows_skips_check_runs_with_nil_app
+    WebMock.disable_net_connect!
+    rate_limit_up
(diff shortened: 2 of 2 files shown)


igor-belousov:
@yegor256 review please

0crat:
@igor-belousov Hey! Just a heads up - your branch name `fix/fetch-workflows-nil-app` doesn't follow our [naming convention](https://www.yegor256.com/2014/04/15/github-guidelines.html), so I've had to deduct **-6** points as per our policy. For next time, try using just the ticket number as the branch name - something like `1349` would be perfect for this case. Your current score is [+32](https://zerocracy.github.io/judges-action/zerocracy-vitals.html), and don't forget to keep an eye on your [Zerocracy account](https://www.zerocracy.com/) too!

yegor256:
@igor-belousov there is a merge conflict here, please fix

igor-belousov:
Closing as duplicate of #1352 — same `run.dig(:app, :slug)` fix, already in master. The test coverage in #1352 is split into two test cases (nil-only and nil + github-actions happy path), while my single test combined nil + other-app + github-actions in one assertion; both approaches exercise the fixed branch. Since the code change is identical and merged, no value in resolving the conflict here. Sorry for the noise!

Issue #151, opened 2024-07-23
`quality-of-service.rb` does too much in one go


yegor256:
Currently, the `quality-of-service.rb` script tries to collect all possible data in one step. This inevitably leads to GitHub API blockage -- they simply don't like us making so many requests in such a short period of time.

One solution would be to make long pauses between steps inside the `quality-of-service.rb` script, such as 5-10 seconds.

Another solution would be to split the process into a few steps: on each step we collect certain metric, add it to the existing fact, and quit. On the next run of the script, we collect another metric, and so on.

Maybe, we can use both approaches.

PR #1389, opened 2026-05-20
#151 pause between quality-of-service metric collections

edmoffo:
Addresses #151 with the first of the two solutions the issue proposes — pause between steps inside `quality-of-service` so the script does not flood the GitHub API in one go. Splitting the metrics across cycles is a larger change that can land separately.

`judges/quality-of-service/quality-of-service.rb` now passes `pause: 5` to `Jp.incremate`, which sleeps for five seconds between each `some_*` evaluation. The pause runs between evaluations only — not before the first and not after the last — so a single-metric run is not slowed down. `Jp.incremate` accepts the new keyword with a default of `0`, leaving every other caller (notably `dimensions-of-terrain`) unaffected.

Checked locally: `bundle exec ruby -Itest test/lib/test_incremate.rb` runs both the pre-existing `test_incremate` and a new `test_incremate_pauses_between_evaluations`, all assertions green; `bundle exec rubocop lib/incremate.rb judges/quality-of-service/quality-of-service.rb test/lib/test_incremate.rb` reports no offenses. The Docker-based `make` cycle was not run in this environment.

Refs #151

### DIFF
--- a/judges/quality-of-service/simple-collect.yml
+++ b/judges/quality-of-service/simple-collect.yml
@@ -12,6 +12,7 @@ input:
     area: quality
     qos_days: 7
     qos_interval: 3
+    qos_pause_seconds: 0
   -
     _id: 1
     what: quality-of-service
--- a/judges/quality-of-service/quality-of-service.rb
+++ b/judges/quality-of-service/quality-of-service.rb
@@ -12,6 +12,6 @@ require_relative '../../lib/incremate'
 days = Fbe.pmp.quality.qos_days
 Jp.cover_qo(days)
 Fbe.consider("(and (eq what '#{$judge}') (exists since) (exists when))") do |f|
-  Jp.incremate(f, __dir__, 'some', avoid_duplicate: true)
+  Jp.incremate(f, __dir__, 'some', avoid_duplicate: true, pause: 5)
 end
 Fbe.octo.print_trace!
--- a/judges/quality-of-service/quality-of-service.rb
+++ b/judges/quality-of-service/quality-of-service.rb
@@ -10,7 +10,7 @@ require_relative '../../lib/cover_qo'
 require_relative '../../lib/incremate'
 
 days = Fbe.pmp.quality.qos_days
-pause = Fbe.pmp.quality.qos_pause_seconds.value || 5
+pause = Fbe.pmp.quality.qos_pause_seconds.value || 0
 Jp.cover_qo(days)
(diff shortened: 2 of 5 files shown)


judges/quality-of-service/simple-collect.yml
```diff
@@ -12,6 +12,7 @@ input:
     area: quality
     qos_days: 7
     qos_interval: 3
+    qos_pause_seconds: 0
```
kreinba, line 15: `quality-of-service.rb` falls back to `0` when the key is missing (`Fbe.pmp.quality.qos_pause_seconds.value || 0`), and the existing scenarios in `test/judges/test-quality-of-service.rb` set `qos_days` and `qos_interval` without ever setting `qos_pause_seconds` — they keep passing under this change. The new `qos_pause_seconds: 0` lines in `simple-collect.yml` and `skip-duplicate-scan.yml` therefore add no coverage and can be dropped, which keeps the fixtures aligned with the `|| 0` default in the production code.

kreinba (commented):
CI is green and the change is focused: one new keyword on `Jp.incremate`, a default of `0` that leaves existing callers untouched, and a dedicated unit test that exercises the timing path. One inline note on the new fixture entries.

judges/quality-of-service/simple-collect.yml
edmoffo: @kreinba Right — `qos_pause_seconds` resolves to `0` through the `|| 0` fallback, so the existing `qos_days`/`qos_interval` scenarios stay unaffected and the new entry is the only place the keyword is set.

0crat:
@kreinba Thanks for the review! You've earned **+13 points** for this contribution: **+18** as the base reward for authoring a code review, **-5** for fewer than 9 comments (only 1 comment was made during the review). Your running score is [+271](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

0crat:
@edmoffo Hey! Nice work on that contribution – you've just earned yourself **+24 points** 🎉 The bonus calculation took into account your code size, review feedback, and overall quality, so you're hitting that sweet spot we're looking for. Your running score is now [+222](https://zerocracy.github.io/judges-action/zerocracy-vitals.html), so [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those quality contributions flowing! Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

Issue #1342, opened 2026-04-29
Misleading control flow: break inside Jp.supervision doesn't stop each_with_index in github-events.rb


vitaly-andr:
> **Note on severity:** practical impact is small. The loop body is correctly skipped after `break`, so no extra factbase writes and no extra GitHub API calls happen. The cost is just microseconds of wasted iteration plus the misleading log lines (`"stop now"`, `"good stop"`) firing repeatedly. Filing this as a code-clarity finding rather than an urgent bug — the author's intent (per the log messages and the `max_events` naming) doesn't match Ruby semantics, and a future change to the loop body could amplify the impact. Skip if you'd rather not chase low-priority cleanups.

## Summary

In `judges/github-events/github-events.rb` (around lines 300-341), two `break` statements appear inside `Jp.supervision { ... }` blocks with the apparent intent of terminating the outer `each_with_index` loop:

- Line 304: `break` after `idx >= max_events` check, with `$loog.debug("Already scanned X events in #{rname}, stop now")`
- Line 314: `break` after `id <= latest` check, with `"good stop in {repo}"`

In Ruby, `break` inside a block exits the **method to which the block was passed** — that's `Jp.supervision`, not `each_with_index`. So the outer loop continues iterating through all remaining events; only the body inside supervision is skipped.

## Verification

\`\`\`ruby
def supervision
  yield
rescue StandardError => e
  raise e
end

[10,20,30,40,50].each_with_index do |val, idx|
  supervision do
    if idx >= 2
      puts "  SKIP idx=#{idx}"
      break
    end
    puts "  PROCESS val=#{val}"
  end
end
# => PROCESS 10, PROCESS 20, SKIP 2, SKIP 3, SKIP 4
\`\`\`

The first two iterations process normally; the last three hit `break` inside supervision and resume iteration in `each_with_index`.

## Suggested fix (if you want it)

Wrap with `catch(:done) ... throw :done`. `throw` propagates through `Jp.supervision`'s `rescue StandardError` cleanly because `UncaughtThrowError` is not a `StandardError`.

Alternative: a flag variable checked at the top of the outer block.

Happy to send a PR if you'd like the cleanup.

0crat:
[...]

PR #1393, opened 2026-05-20
#1342 stop event scan with catch/throw instead of break inside Jp.supervision

edmoffo:
Addresses #1342, which reports that the two `break` statements inside `Jp.supervision { ... }` in `judges/github-events/github-events.rb` (around lines 300-341 on master) don't terminate the surrounding `each_with_index` — `break` exits the method that received the block, which here is `Jp.supervision`, not the outer enumerator. The outer loop kept iterating after `max_events` was reached or after an event id fell at or below the latest seen, so the `"Already scanned ... stop now"` and `"good stop in ..."` debug lines fired once per remaining event and the subsequent guard work was repeated for every leftover entry.

This PR wraps the scan in `catch(:done)` and replaces both `break` statements with `throw :done`. `UncaughtThrowError` is not a `StandardError`, so the throw flies through `Jp.supervision`'s `rescue StandardError` clause untouched and the outer `each_with_index` terminates on the first hit, which is what the original log wording already promised. The inner `throw :rollback` markers stay on a different symbol and keep their existing scope inside `Fbe.fb.txn`.

Verified locally with `bundle exec ruby test/judges/test-github-events.rb` — 42 tests, 127 assertions, 0 failures, 0 errors, 1 skip. `bundle exec rubocop judges/github-events/github-events.rb` reports no offences. The existing `test_stop_scanning_if_number_event_greater_than_max_events` and `test_stop_scanning_if_event_id_less_or_eq_than_latest` tests still pass, so the observable behaviour on small fixtures is unchanged; the fix shows up in production where the upstream enumerator is paginated and the old code kept asking it for more events.

Refs #1342

### DIFF
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -297,46 +297,48 @@ Fbe.iterate do
       'label-was-attached' => %w[where repository issue label],
       'type-was-attached' => %w[where repository issue type]
     }
-    Fbe.octo.repository_events(repository).each_with_index do |json, idx|
-      Jp.supervision({ 'repo' => rname, 'json' => json.to_h }) do
-        if !$options.max_events.nil? && idx >= $options.max_events
-          $loog.debug("Already scanned #{idx} events in #{rname}, stop now")
-          break
-        end
-        total += 1
-        id = Integer(json[:id])
-        first = id if first.nil?
-        if id <= latest
-          $loog.debug(
-            "The event_id ##{id} (no.#{idx}) is not larger than ##{latest}, " \
-            "good stop in #{json[:repo][:name]}"
-          )
-          break
-        end
-        Fbe.fb.txn do |fbt|
-          f =
-            Fbe.if_absent(fb: fbt) do |n|
-              n.event_id = Integer(json[:id])
-              n.where = 'github'
+    catch(:done) do
+      Fbe.octo.repository_events(repository).each_with_index do |json, idx|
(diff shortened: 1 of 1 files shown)


kreinba (approved):
The diff matches the linked issue. The `break` in the original code exited `Jp.supervision` rather than the surrounding `each_with_index`, so the enumerator kept paging through events after the stop conditions were already met. Wrapping the outer `each_with_index` in `catch(:done)` and throwing `:done` from the two stop points propagates through `Jp.supervision`'s `rescue StandardError` clause without being intercepted, and the inner `throw :rollback` stays on its own symbol inside `Fbe.fb.txn`. The 13 CI checks are green on `9c1abf5b`.

0crat:
@kreinba Thanks for the review! You've earned **+10 points** for this contribution: +18 as a basis, -8 for absolutely no comments posted during the review. Your running score is [+281](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

0crat:
@edmoffo Great work on your contribution! 🎉 You've earned **+24 points** (+24 base, +5 for 78 hits-of-code, -6 cap adjustment). Your dedication to [quality delivery](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) is appreciated! Running total: **+246 points** - track your progress on your [Zerocracy account](https://www.zerocracy.com/) and view team metrics [here](https://zerocracy.github.io/judges-action/zerocracy-vitals.html).

PR #1404, opened 2026-05-21
#1392 test absoluteness of sqlite cache path, not INPUT_FACTBASE

### LINKED ISSUE #1392: `entry.sh:209` sqlite cache absoluteness check inspects `INPUT_FACTBASE` instead of the cache path itself
In `entry.sh:207-211` the sqlite cache path is normalised with `realpath`:

```
sqlite=$(printenv "INPUT_SQLITE-CACHE" || true)
if [ -n "${sqlite}" ]; then
    sqlite=$(realpath "$( [[ ${INPUT_FACTBASE} = /* ]] && echo "${sqlite}" || echo "${GITHUB_WORKSPACE}/${sqlite}" )")
```

The conditional on line 209 inspects `${INPUT_FACTBASE}` for a leading slash, but the value used in both branches is `${sqlite}`. The decision about whether to bolt `${GITHUB_WORKSPACE}/` onto the front of the cache path is therefore driven by a completely unrelated input, and the cache path itself is never examined.

Two concrete failure modes follow. When `INPUT_FACTBASE` is absolute (for example `/tmp/fb.db`) and `INPUT_SQLITE-CACHE` is relative (for example `cache/http.sqlite`), the script feeds the bare `cache/http.sqlite` to `realpath`, so it is resolved against the current working directory (the `${SELF}` install dir reached at line 81) rather than against `${GITHUB_WORKSPACE}` as the relative form intends. When `INPUT_FACTBASE` is relative and `INPUT_SQLITE-CACHE` is absolute (for example `/var/cache/judges.sqlite`), the script prepends `${GITHUB_WORKSPACE}/` to the already-absolute path, producing `${GITHUB_WORKSPACE}//var/cache/judges.sqlite`; `realpath` collapses the double slash but the resolved file is not what the user asked for. The same `INPUT_FACTBASE`-driven check appears at line 70 for `fb` and is correct there because the variable being normalised on that line is `INPUT_FACTBASE`; line 209 copied the pattern without retargeting the test.

Fix: replace `${INPUT_FACTBASE}` with `${sqlite}` in the test on line 209, so the absoluteness check matches the value being normalised:

```
sqlite=$(realpath "$( [[ ${sqlite} = /* ]] && echo "${sqlite}" || echo "${GITHUB_WORKSPACE}/${sqlite}" )")
```

`entry.sh:209`.

kreinba:
Closes #1392.

`entry.sh:209` was the sqlite cache path normalisation line. The conditional `[[ ${INPUT_FACTBASE} = /* ]]` inspected the factbase path for a leading slash but then used `${sqlite}` in both branches, so the absoluteness of the cache path was being decided by an unrelated input. With `INPUT_FACTBASE` absolute and `INPUT_SQLITE-CACHE` relative, the bare relative path was fed to `realpath` and resolved against the install directory rather than `${GITHUB_WORKSPACE}`; with `INPUT_FACTBASE` relative and `INPUT_SQLITE-CACHE` absolute, `${GITHUB_WORKSPACE}/` was prepended to an already-absolute path, producing a double-slash path that `realpath` collapsed to the wrong file. The same pattern at line 70 normalises `INPUT_FACTBASE` and correctly tests `INPUT_FACTBASE`; line 209 copied the pattern without retargeting the test.

The fix swaps `${INPUT_FACTBASE}` for `${sqlite}` in the test on line 209, so the absoluteness check matches the value being normalised. The two branches and the surrounding `realpath` wrapper are unchanged.

Ran `shellcheck entry.sh` (clean) and `bashate entry.sh` (no new warnings; the line-too-long warning on line 209 pre-existed). The change is shell-only and does not touch any Ruby code, so the rake suite is unaffected; CI will run the full make/rake/actionlint/bashate/checkmake/copyrights/hadolint/pdd/reuse/shellcheck/typos/xcop/yamllint matrix on the push.

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -206,7 +206,7 @@ fi
 
 sqlite=$(printenv "INPUT_SQLITE-CACHE" || true)
 if [ -n "${sqlite}" ]; then
-    sqlite=$(realpath "$( [[ ${INPUT_FACTBASE} = /* ]] && echo "${sqlite}" || echo "${GITHUB_WORKSPACE}/${sqlite}" )")
+    sqlite=$(realpath "$( [[ ${sqlite} = /* ]] && echo "${sqlite}" || echo "${GITHUB_WORKSPACE}/${sqlite}" )")
     options+=("--option=sqlite_cache=${sqlite}");
     echo "Using SQLite for HTTP caching: ${sqlite}"
     ${JUDGES} "${gopts[@]}" download \


0crat:
@kreinba Hey! Nice work on your contribution! 🎉 You've earned **+20 points** this time: started with **+24** as the base reward, but had to deduct **-4** since you only contributed **2** [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (our policy encourages at least **30** for optimal scoring). Next time, try bundling a few more changes together to maximize your points - you could earn up to **0.075 points** per hit-of-code! Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+447](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #1411, opened 2026-05-22
Use locked bundle when building Docker image

### LINKED ISSUE #1408: `Dockerfile` incorrectly runs `bundle update` ignoring locked versions in `Gemfile.lock`
## Symptom

`Dockerfile:18` installs gems with:

```dockerfile
COPY Gemfile /action
COPY Gemfile.lock /action
RUN bundle update --gemfile=/action/Gemfile
```

`bundle update` ignores the `Gemfile.lock` that was just copied on the line above, re-resolves every dependency against the loosest version constraint that the `Gemfile` allows (for example `gem 'fbe', '~>0.41'` pulls whatever `0.41+` is current at build time), and rewrites the lockfile inside the image. The shipped action therefore runs on a gem set that is decided at `docker build` time, not on the gem set that `Gemfile.lock` pins and that `make.yml` exercises through `bundler-cache: true` on `ruby/setup-ruby@v1`.

## Why it is wrong

The CI workflow (`.github/workflows/make.yml`) and the deployed Docker image build against two different dependency sets: CI runs the locked versions, the image runs whatever was current the day it was built. A green CI run on the master branch is no longer evidence that the published action will boot — a transitive dependency can ship a breaking change between the CI run and the next image rebuild, and the bug shows up only in production. Renovate-managed lockfile bumps (commits like `248a1ba Lock file maintenance`) are also wasted: the values they pin are discarded inside the container.

## Suggested fix

Replace `bundle update --gemfile=/action/Gemfile` on `Dockerfile:18` with `bundle install --deployment` (or at minimum `bundle install --frozen`), so the build fails when `Gemfile.lock` is missing or stale and otherwise installs exactly the locked versions.

puneetdixit200:
Fixes #1408.

The Dockerfile copied `Gemfile.lock` and then ran `bundle update`, which re-resolves dependencies during image build and ignores the committed lockfile. This makes the action image use whatever gem versions are current at build time instead of the versions tested by CI and maintained by Renovate lockfile updates.

This changes the image build to run `bundle install --deployment`, so Docker builds install the locked dependency set and fail if the lockfile is stale or missing. I also added a small regression test that prevents `bundle update` from being reintroduced in the Dockerfile.

Tests run with Ruby 4.0.4 and the locked bundle installed under `/tmp/judges-action-1408-bundle`:
- `ruby test/test_dockerfile.rb` (red before the Dockerfile change)
- `bundle exec ruby test/test_dockerfile.rb`
- `bundle exec rake test` (199 tests, 551 assertions, 0 failures, 0 errors, 1 skip)
- `bundle exec rake rubocop` (96 files inspected, no offenses)
- `ruby -c test/test_dockerfile.rb`
- `git diff --check`

I did not run `make` locally because Docker is installed but the Docker daemon is not running (`docker info` cannot connect to the local socket).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN apt-get update \
 WORKDIR /action
 COPY Gemfile /action
 COPY Gemfile.lock /action
-RUN bundle update --gemfile=/action/Gemfile
+RUN bundle install --deployment
 
 COPY judges /action/judges
 COPY lib /action/lib
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,8 @@ RUN apt-get update \
 WORKDIR /action
 COPY Gemfile /action
 COPY Gemfile.lock /action
-RUN bundle install --deployment
+RUN bundle config set deployment true \
+    && bundle install
 
 COPY judges /action/judges
 COPY lib /action/lib
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ all: rubocop test entry rmi verify entries
 
 test: target/docker-image.txt
 	img=$$(cat target/docker-image.txt)
-	docker run --rm --entrypoint '/bin/bash' "$${img}" -c 'judges test --disable live --lib /action/lib /action/judges'
+	docker run --rm --entrypoint '/bin/bash' "$${img}" -c 'bundle exec judges test --disable live --lib /action/lib /action/judges'
 	echo "$$?" > target/test.exit
 
 entry: target/docker-image.txt
--- a/test/test_dockerfile.rb
+++ b/test/test_dockerfile.rb
@@ -13,4 +13,9 @@ class TestDockerfile < Minitest::Test
     assert_match(/\bbundle\s+config\s+set\s+deployment\s+true\b/, dockerfile)
     assert_match(/\bbundle\s+install\b/, dockerfile)
   end
+
+  def test_makefile_runs_judges_through_bundle
+    makefile = File.read(File.join(__dir__, '..', 'Makefile'))
+    assert_match(/\bbundle\s+exec\s+judges\s+test\b/, makefile)
+  end
 end
--- a/test/test_dockerfile.rb
+++ b/test/test_dockerfile.rb
@@ -9,6 +9,8 @@ class TestDockerfile < Minitest::Test
   def test_installs_locked_bundle
     dockerfile = File.read(File.join(__dir__, '..', 'Dockerfile'))
     refute_match(/\bbundle\s+update\b/, dockerfile)
-    assert_match(/\bbundle\s+install\s+--deployment\b/, dockerfile)
+    refute_match(/\bbundle\s+install\s+--deployment\b/, dockerfile)
+    assert_match(/\bbundle\s+config\s+set\s+deployment\s+true\b/, dockerfile)
+    assert_match(/\bbundle\s+install\b/, dockerfile)
   end
 end
--- a/test/test_dockerfile.rb
+++ b/test/test_dockerfile.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+require 'minitest/autorun'
+
+class TestDockerfile < Minitest::Test
+  def test_installs_locked_bundle
+    dockerfile = File.read(File.join(__dir__, '..', 'Dockerfile'))
+    refute_match(/\bbundle\s+update\b/, dockerfile)
+    assert_match(/\bbundle\s+install\s+--deployment\b/, dockerfile)
+  end
+end


Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN apt-get update \
 WORKDIR /action
 COPY Gemfile /action
 COPY Gemfile.lock /action
-RUN bundle update --gemfile=/action/Gemfile
+RUN bundle install --deployment
```
kreinba: Current Bundler refuses this flag — the `make` check on this PR fails at `[6/9] RUN bundle install --deployment` with `The \`--deployment\` flag has been removed because it relied on being remembered across bundler invocations, which bundler no longer does. Instead please use \`bundle config set deployment true\`, and stop using this flag` (exit 15). Switch to two steps so deployment mode is persisted in `/action/.bundle/config` and honored by the install:

```suggestion
RUN bundle config set --local deployment true && bundle install
```

This still fails the build if the lockfile is stale (the documented behavior of deployment mode) and keeps the regression test green once its `bundle\s+install\s+--deployment` matcher is updated to match the new form, or relaxed to assert `bundle config set .* deployment true` alongside `bundle install`.

kreinba (changes requested):
The intent (install the locked gem set) is right, and the `test_dockerfile.rb` regression test is a good guard. The build is broken though: Bundler now rejects `--deployment` on the command line and the `make` job exits with code 15. The fix needs to switch to the `bundle config` form before `bundle install` so the deployment mode is persisted into `/action/.bundle/config` and honored by the install step. One inline comment.

puneetdixit200:
Fixed in `17dfef98` and `9a03e82`: the Docker build now uses `bundle config set deployment true` before `bundle install`, and the Docker test runs through the bundled executable. The latest `make` check is green on the PR head.

kreinba:
@puneetdixit200 Confirmed. The Dockerfile now installs the locked gem set via `bundle config set deployment true && bundle install` (`17dfef9`), and `bundle exec judges test` runs through the deployed bundle (`9a03e82`). The new `test_dockerfile.rb` guards against `bundle update` and the removed `--deployment` CLI flag, and the `make` job is green on the latest head.

kreinba (approved):
The earlier CHANGES_REQUESTED is addressed: `bundle config set deployment true && bundle install` in the Dockerfile installs the locked gem set without invoking the removed `--deployment` CLI flag, `make test` in the Docker container runs through `bundle exec judges test` so the deployed bundle is exercised, and `test/test_dockerfile.rb` guards against `bundle update` and the removed CLI flag in regression. All 13 CI checks pass on the current head.

0crat:
@kreinba Thanks for the review! You've earned **+13 points** for this contribution: **+18** as a basis for reviewing, **-5** for having fewer than 9 comments (only 1 comment was made). According to our [work policy](https://github.com/zerocracy/judges/blob/master/POLICY.md), reviews with more comments earn higher bonuses - aim for 9+ comments to avoid deductions! 💪 Your running score is [+887](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

yegor256:
@puneetdixit200 there is a merge conflict here, please fix

puneetdixit200:
Merge conflict is fixed in 7b2e771, and the rerun checks are green now.

puneetdixit200:
Merged current master into this branch and resolved the Dockerfile/test conflict. The Dockerfile keeps the locked install via `bundle config set frozen true && bundle install --gemfile=/action/Gemfile`, and the test keeps coverage for the Makefile running `bundle exec judges test`.

Verification:
- `PATH=/opt/homebrew/opt/ruby/bin:$PATH BUNDLE_PATH=vendor/bundle bundle exec ruby test/test_dockerfile.rb` -> 2 tests, 10 assertions, 0 failures
- GitHub Actions on `7b2e771`: all 13 checks passing, including `make` and `rake`

GHX5T-SOL (approved):
Approved current head `7b2e7712e5f2509ddd33049077656ee8040f7880`.

Against current `master`, this diff is limited to `Makefile` and `test/test_dockerfile.rb`: the Dockerfile already installs with `bundle config set frozen true && bundle install --gemfile=/action/Gemfile`, and the Makefile now runs `bundle exec judges test` inside the built image so the smoke path uses the locked bundle rather than a globally resolved executable. The regression test pins both sides by rejecting `bundle update` and deprecated `bundle install --deployment`, asserting the frozen/gemfile install, and asserting the Makefile command goes through Bundler.

Local validation: `bundle exec ruby -Itest test/test_dockerfile.rb` -> 2 tests, 10 assertions; `bundle exec rake test` -> 220 tests, 613 assertions, 1 skip; `bundle exec rubocop test/test_dockerfile.rb`; `git diff --check origin/master...HEAD`; `gitleaks detect --no-banner --redact --source .`.

0crat:
@puneetdixit200 Thank you for your contribution! You've earned **+20 points** for this submission: **+24** as the base reward, but **-4** for contributing fewer than 30 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (only 8). Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the contributions coming! Your running score is [+20](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) as well.

Issue #1395, opened 2026-05-20
`label-was-attached.rb` crashes on timeline events whose `:label` or `:actor` is missing


edmoffo:
In `judges/label-was-attached/label-was-attached.rb:52` the timeline event is unwrapped with `te[:label][:name]`, and lines 67 and 69 do the same with `te[:actor][:id]` and `te[:actor][:login]`. The sibling judge `judges/issue-was-closed/issue-was-closed.rb:95` walks the exact same `Fbe.octo.issue_timeline` stream for the same `labeled` events but unwraps the field with the safe form `te.dig(:label, :name)`. The two judges disagree on the safety of the same field, and only the safer form survives the conditions that GitHub does emit in practice.

GitHub keeps a timeline entry whose `event` is `labeled` even after the label is deleted from the repository, with the `:label` payload set to `null`; the same applies to `:actor` when the user account has been removed. The `Sawyer::Resource#[]` chain `te[:label][:name]` then evaluates `nil[:name]` and raises `NoMethodError: undefined method '[]' for nil`, which propagates out of the `over do |repository, issue|` block, crashes the iterate cycle for that repository, and leaves the rest of the issues on the page unprocessed. The judge runs every cycle, so a single dead label is enough to keep `label-was-attached` stuck on the same repo run after run.

Fix: replace the three unsafe accesses in `judges/label-was-attached/label-was-attached.rb` with their `dig` equivalents — `te.dig(:label, :name)` on line 52, `te.dig(:actor, :id)` on line 67, and `te.dig(:actor, :login)` on line 69 — and skip the event with `next` when the label name is nil, matching the form already used by `issue-was-closed.rb`.

edmoffo:
@yegor256 could you take a look when you have a moment? Happy to clarify anything in the report.

0crat:
[...]

PR #1413, opened 2026-05-22
#1409 mark fact stale on who when timeline actor is deleted

### LINKED ISSUE #1409: `type-was-attached` and `issue-was-closed` wrongly store `nil` for deleted timeline actors instead of marking as `stale`
`judges/type-was-attached/type-was-attached.rb:70` sets `nn.who = tee.dig('actor', 'id')` and `judges/issue-was-closed/issue-was-closed.rb:110` sets `nn.who = te.dig(:actor, :id)` without guarding for the deleted-actor case. When GitHub returns a timeline event whose `actor` payload is `null` — which happens routinely once the user account has been removed — the right-hand side evaluates to `nil` and the fact ends up with `who = nil`, diverging from every sibling judge that touches the same timeline stream.

The convention across the rest of the tree is the `stale = 'who'` fallback already used by `label-was-attached.rb:67-71` after the fix for #1395, by `pull-was-merged.rb:90`, by `who-is-alive.rb:36`, by `who-has-name.rb:34` and `:57`, by `is-human-or-robot.rb:27` and `:31`, by `fix-missing-who.rb:42`, and by `eliminate-ghosts.rb:32` and `:38`. Those judges all refuse to write `who = nil` and instead mark the fact stale on the `who` axis, so downstream judges like `fix-missing-who` and `revive-user` can pick it up later. The two divergent judges leak a `nil`-valued `who` into the factbase and skip the stale signal, which is the same shape of defect that #1395 reported for `label-was-attached.rb` before it was patched.

Fix: in `type-was-attached.rb:70` and `issue-was-closed.rb:110`, replace the bare `nn.who = …dig(:actor, :id)` with the existing pattern — capture the value into a local, set `nn.who = who` when truthy, set `nn.stale = 'who'` otherwise — exactly as `label-was-attached.rb:67-71` now does. The `@#{…dig(:actor, :login)}` interpolation in the same `nn.details =` line should also fall back to an `'an unknown actor'` literal so the rendered detail does not produce a dangling `@`.

kreinba:
Closes #1409.

`judges/type-was-attached/type-was-attached.rb:70` set `nn.who = tee.dig('actor', 'id')` and `judges/issue-was-closed/issue-was-closed.rb:110` set `nn.who = te.dig(:actor, :id)` without guarding for the deleted-actor case. When GitHub returns a timeline event whose `actor` payload is `null` — which happens routinely once the user account has been removed — the right-hand side evaluated to `nil` and the fact ended up with `who = nil`, diverging from every sibling judge that touches the same timeline stream (`label-was-attached`, `pull-was-merged`, `who-is-alive`, `who-has-name`, `is-human-or-robot`, `fix-missing-who`, `eliminate-ghosts`) and skipping the `stale = 'who'` signal that downstream judges like `fix-missing-who` and `revive-user` rely on to backfill the actor on a later cycle.

Aligns both judges with the canonical fallback already in `label-was-attached.rb:66-76` (the #1395 fix): capture the actor id into a local, set `nn.who` when truthy, set `nn.stale = 'who'` otherwise, and substitute `'an unknown actor'` for the `@login` interpolation in `nn.details` so the rendered detail does not produce a dangling `@`. No other change.

Verified locally with `bundle exec ruby -e "require 'rake'; load 'Rakefile'; Rake::Task[:test].invoke"` (200 tests, 557 assertions, 0 failures, 0 errors, 1 skip; line coverage 90.23%) and `rubocop` on the four touched files (no offenses). Added one regression test per judge: `test_marks_label_stale_on_who_when_timeline_actor_is_nil` in `test/judges/test-issue-was-closed.rb` stubs `actor: nil` on a `labeled` timeline event and asserts `who` is absent, `stale = ['who']`, and the detail reads `an unknown actor`; `test_marks_stale_when_graphql_actor_is_nil` in `test/judges/test-type-was-attached.rb` stubs the GraphQL `issue_type_event` response with an `actor` whose `id`/`login` are `nil` and makes the same three assertions.

### DIFF
--- a/judges/issue-was-closed/issue-was-closed.rb
+++ b/judges/issue-was-closed/issue-was-closed.rb
@@ -107,10 +107,16 @@ Fbe.iterate do
           $loog.warn("A label #{badge.inspect} is already attached to #{repo}##{issue}")
           next
         end
-        nn.who = te.dig(:actor, :id)
+        who = te.dig(:actor, :id)
+        if who
+          nn.who = who
+        else
+          nn.stale = 'who'
+        end
         nn.when = te[:created_at]
+        actor = te.dig(:actor, :login)
         nn.details =
-          "Seemingly, the #{nn.label.inspect} label was attached by @#{te.dig(:actor, :login)} " \
+          "Seemingly, the #{nn.label.inspect} label was attached by #{actor ? "@#{actor}" : 'an unknown actor'} " \
           "to the issue #{Fbe.issue(nn)}."
         $loog.info("Label attached to #{Fbe.issue(nn)} found: #{nn.label.inspect}")
       end
--- a/judges/type-was-attached/type-was-attached.rb
+++ b/judges/type-was-attached/type-was-attached.rb
@@ -67,10 +67,16 @@ Fbe.iterate do
             n.where = 'github'
           end
         raise(RuntimeError, "Type already attached to #{repo}##{issue}") if nn.nil?
-        nn.who = tee.dig('actor', 'id')
+        who = tee.dig('actor', 'id')
+        if who
+          nn.who = who
+        else
+          nn.stale = 'who'
+        end
         nn.when = tee['created_at']
+        actor = tee.dig('actor', 'login')
         nn.details =
-          "The #{nn.type.inspect} type was attached by @#{tee.dig('actor', 'login')} " \
+          "The #{nn.type.inspect} type was attached by #{actor ? "@#{actor}" : 'an unknown actor'} " \
           "to the issue #{Fbe.issue(nn)}."
         $loog.info("Type attached to #{Fbe.issue(nn)} found: #{nn.type.inspect}")
       end
--- a/test/judges/test-issue-was-closed.rb
+++ b/test/judges/test-issue-was-closed.rb
@@ -170,6 +170,41 @@ class TestIssueWasClosed < Jp::Test
     assert(fb.one?(what: 'label-was-attached', repository: 42, issue: 52, where: 'github', label: 'bug', who: 421))
   end
 
+  def test_marks_label_stale_on_who_when_timeline_actor_is_nil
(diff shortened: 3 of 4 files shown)


edmoffo (approved):
Both touched judges now mirror the label-was-attached fallback at label-was-attached.rb:66-77 line for line: the actor id is captured into a local, the fact is marked stale=who when the timeline payload is anonymous, and the @login interpolation falls back to "an unknown actor" so the rendered detail never trails a bare @. The two new regression tests pin the contract on both code paths — the REST timeline path in test-issue-was-closed.rb and the GraphQL issue_type_event path in test-type-was-attached.rb — and assert who is absent, stale=['who'], and the detail names an unknown actor. All thirteen checks are green on the head commit.

0crat:
@edmoffo Hey there! 🎉 Nice work on that review - you've bagged **+10 points** this time around! Here's the breakdown: you got the full **+18 base points** for being a review author, but we had to dock **-8 points** since there were no comments posted (our policy dings us when reviews go silent). Next time, dropping a few comments could boost your score even more - you can earn up to **+8 extra points** just for engaging in the discussion! Your running total is now [**+300**](https://zerocracy.github.io/judges-action/zerocracy-vitals.html), so keep crushing it and don't forget to peek at your [**Zerocracy account**](https://www.zerocracy.com/) too!

0crat:
@kreinba **Excellent work!** 🎉 You've earned **+24 points** for this contribution: +24 base points, +7 for your 94 hits-of-code, -7 due to cap limits. Your running score is now [+957](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - fantastic progress! Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) up the great work and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #1417, opened 2026-05-24
#1390 use jq -r so latest release tag is not double-quoted in entry.sh

### LINKED ISSUE #1390: `entry.sh` version check inaccurately reports version mismatch due to `jq` outputting quoted JSON string
In `entry.sh:14-15`, the latest release tag is read with a `jq` invocation that is missing the `-r` (raw output) flag:

```
resp=$(curl --silent -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/zerocracy/judges-action/releases/latest)
latest=$(echo -n "$resp" | jq -Rs "try (fromjson | .tag_name) catch empty")
```

Without `-r`, `jq` emits the tag as a JSON-encoded string, so `latest` ends up as the seven-character tag wrapped in two literal quote characters — `"0.17.17"` rather than `0.17.17`.

The subsequent comparison at `entry.sh:22`:

```
elif [ "${latest}" != "${VERSION}" ]; then
```

compares that quoted nine-character value against `VERSION`, which `.rultor.yml` writes unquoted on every release via `sed -i -e "s/0.0.0/${tag}/" entry.sh`. The two strings never match, so the "different version" warning fires on every single run — even when the user is on the latest tag — and the leaked quotes appear in the printed message itself:

```
!!! its GitHub repository is "0.17.17": https://github.com/zerocracy/judges-action.
!!! However, you are using a different version: 0.17.17.
```

The same response is read again at `entry.sh:266` with the correct `jq -r '.tag_name'`, which produces the bare tag and works as intended; the first call is just missing the `-r`.

Fix: add `-r` to the first `jq` invocation, so the two version checks treat the tag the same way:

```
latest=$(echo -n "$resp" | jq -Rrs "try (fromjson | .tag_name) catch empty")
```

`entry.sh:15`.

edmoffo:
Resolves #1390. Without `-r`, the `jq` call on `entry.sh:15` returns the tag as a JSON-encoded string, so the comparison on line 22 compares a quoted nine-character value against the unquoted `VERSION` written by `.rultor.yml`. The two values never match, the version-mismatch warning fires on every run, and the printed message embeds the literal quote characters.

Adding `-r` to the first `jq` invocation makes it emit the bare tag string, matching the canonical reading on `entry.sh:266` that already uses `jq -r '.tag_name'`. The comparison on line 22 then evaluates as intended and the warning only fires on a real version drift.

I ran `bash -n entry.sh` to confirm the script still parses, and I verified the behavior of both forms locally with a stubbed payload:

```
$ echo '{"tag_name":"0.17.17"}' | jq -Rs  "try (fromjson | .tag_name) catch empty"
"0.17.17"
$ echo '{"tag_name":"0.17.17"}' | jq -Rrs "try (fromjson | .tag_name) catch empty"
0.17.17
```

Closes #1390

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -12,7 +12,7 @@ echo "The 'judges-action' ${VERSION} is running"
 
 if [ "${SKIP_VERSION_CHECKING}" != 'true' ]; then
     resp=$(curl --silent -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/zerocracy/judges-action/releases/latest)
-    latest=$(echo -n "$resp" | jq -Rs "try (fromjson | .tag_name) catch empty")
+    latest=$(echo -n "$resp" | jq -Rrs "try (fromjson | .tag_name) catch empty")
     if [ -z "${latest}" ]; then
         echo "!!! Could not fetch the latest version from GitHub."
         echo "!!! GitHub returned: "


bibonix (approved):
Aligns the jq invocation on entry.sh:15 with the canonical reading on entry.sh:266 that already uses jq -r '.tag_name'. The try/catch empty form still produces an empty string on a malformed payload, so the existing [ -z "${latest}" ] guard on line 16 continues to handle the failure path. CI is green across all 13 checks. No inline comments.

GHX5T-SOL (approved):
Approved current head `01b66c4ef24fb01daec012a2d8ff1b621f235313`.

I rechecked the one-line `entry.sh` change against the reported `jq` raw-output issue. The new `jq -Rrs` form returns bare `0.17.17\n` for `{"tag_name":"0.17.17"}`, while the old `jq -Rs` form returns `"0.17.17"\n`. With the live GitHub releases response, the patched command returns `0.17.17` without leading/trailing quotes. Malformed JSON still returns empty output, so the existing `[ -z "${latest}" ]` guard remains reachable.

Validation:
- `bash -n entry.sh` -> passed
- `printf '{"tag_name":"0.17.17"}' | jq -Rrs 'try (fromjson | .tag_name) catch empty' | od -An -t x1c` -> `30 2e 31 37 2e 31 37 0a`
- same payload with old `jq -Rs` -> `22 30 2e 31 37 2e 31 37 22 0a`
- `printf 'not-json' | jq -Rrs 'try (fromjson | .tag_name) catch empty' | od -An -t x1c` -> empty output
- live release response with patched command -> `0.17.17`, no quotes
- `git diff --check origin/master...HEAD` -> clean
- `gitleaks detect --no-banner --redact --source . --log-opts origin/master..HEAD --exit-code 1` -> no leaks found

I did not run the Docker-backed `make` path locally because Docker is not installed on this host; all 13 GitHub checks, including `make`, `rake`, and `shellcheck`, are green on this head.

0crat:
@bibonix Hey there! 🎉 Thanks for jumping in on that review! You've earned **+6 points** this time: started with **+18** as the base reward, but had **-8** deducted since no comments were posted (our policy dings us for zero feedback), and another **-4** off for the low hit-of-code count (only 2, while we need at least 24 to avoid penalties). Your running score is now [+778](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - keep it up and remember that adding some comments during reviews can boost your points significantly (up to +8 more!). Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too!

0crat:
@bibonix Hey there! 🎉 Thanks for jumping in on that review! You've earned **+6 points** this time: started with **+18** as the base reward, but had **-8** deducted since no comments were posted (our policy dings us for zero feedback), and another **-4** off for the low hit-of-code count (only 2, while we need at least 24 to avoid penalties). Your running score is now [+778](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - keep it up and remember that adding some comments during reviews can boost your points significantly (up to +8 more!). Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too!

0crat:
@ghx5t-sol Thanks for the review! You've earned **+6 points** for this contribution: **+18** as a basis, **-8** for absolutely no comments posted during the review, and **-4** for too few hits-of-code (only 2, which is below the 24 minimum). Your running score is [+200](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

0crat:
@edmoffo Hey there! Nice work on your contribution – you've snagged **+20 points** this time around! 🎉 Here's the breakdown: **+24** base points, but **-4** since you only had **2** [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (our policy dings contributions under **30** HoC to encourage meatier submissions). Next time, try bundling a bit more code together to maximize your score – you could easily hit the **24-point** cap! Your running total is sitting pretty at **[+883](https://zerocracy.github.io/judges-action/zerocracy-vitals.html)**, so [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those contributions rolling and don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) too!

PR #1438, opened 2026-05-27
test: #1342 pin catch/throw exit in github-events scan

### LINKED ISSUE #1342: Misleading control flow: break inside Jp.supervision doesn't stop each_with_index in github-events.rb

edmoffo:
PR #1393 replaced the two `break` statements inside `Jp.supervision { ... }` in `judges/github-events/github-events.rb` with `throw :done` for #1342, but no test pins the new exit path. The pre-existing `test_stop_scanning_if_number_event_greater_than_max_events` and `test_stop_scanning_if_event_id_less_or_eq_than_latest` tests assert factbase content only, which stays identical whether the supervision block runs once (`throw :done`) or N times (`break`) — both paths skip the body, so neither test catches a silent regression back to `break`.

This change adds two regression tests that load `github-events` with a `Loog::Buffer` logger, then count the `Already scanned ...` and `good stop in ...` debug lines emitted by the two guards. Each must appear exactly once on the catch/throw path; with `break` in place, they fire once per remaining event.

Verified locally with `bundle exec ruby -Ilib -Itest test/judges/test-github-events.rb -n "/max_events_guard_fires_once|latest_id_guard_fires_once/" -- --no-cov` — both new tests pass on master and both fail with `Expected: 1, Actual: 4` when the production code is temporarily reverted to `break`. The full file (`49 tests, 152 assertions, 0 failures, 0 errors, 1 skip`) is green, and `bin/rubocop test/judges/test-github-events.rb` is clean.

Closes #1342.

### DIFF
--- a/test/judges/test-github-events.rb
+++ b/test/judges/test-github-events.rb
@@ -1955,6 +1955,63 @@ class TestGithubEvents < Jp::Test
     assert(fb.none?(what: 'tag-was-created', where: 'github', event_type: 'CreateEvent', repository: 42, event_id: 13))
   end
 
+  def test_max_events_guard_fires_once
+    WebMock.disable_net_connect!
+    rate_limit_up
+    stub_github(
+      'https://api.github.com/repos/foo/foo',
+      body: { id: 42, name: 'foo', full_name: 'foo/foo', default_branch: 'master' }
+    )
+    stub_github(
+      'https://api.github.com/repositories/42',
+      body: { id: 42, name: 'foo', full_name: 'foo/foo', default_branch: 'master' }
+    )
+    stub_github('https://api.github.com/user/42', body: { id: 42, login: 'torvalds' })
+    stub_github(
+      'https://api.github.com/repositories/42/events?per_page=100',
+      body: 6.downto(1).map do |id|
+        {
+          id:, created_at: Time.now.to_s, actor: { id: 42 },
+          type: 'CreateEvent', repo: { id: 42 },
+          payload: { ref_type: 'tag', ref: 'foo' }
+        }
+      end
+    )
+    fb = Factbase.new
+    loog = Loog::Buffer.new
+    load_it('github-events', fb, Judges::Options.new({ 'repositories' => 'foo/foo', 'max_events' => 2 }), loog:)
+    assert_equal(1, loog.to_s.scan('Already scanned').size)
+  end
+
+  def test_latest_id_guard_fires_once
+    WebMock.disable_net_connect!
+    rate_limit_up
+    stub_github(
+      'https://api.github.com/repos/foo/foo',
+      body: { id: 42, name: 'foo', full_name: 'foo/foo', default_branch: 'master' }
+    )
+    stub_github(
+      'https://api.github.com/repositories/42',
+      body: { id: 42, name: 'foo', full_name: 'foo/foo', default_branch: 'master' }
+    )
+    stub_github('https://api.github.com/user/42', body: { id: 42, login: 'torvalds' })
+    stub_github(
+      'https://api.github.com/repositories/42/events?per_page=100',
+      body: 15.downto(10).map do |id|
+        {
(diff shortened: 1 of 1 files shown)


GHX5T-SOL (approved):
Verified the regression coverage on `3361fa8ef73939b8657af17b30526477a32d15a7`.

The two new tests pin the behavior that matters for #1342: both early-stop guards now emit their stop log exactly once, which would regress under the old `break`-inside-`Jp.supervision` path. The PR is test-only, narrow, and all hosted checks are green.

Local validation:
- `bundle exec ruby -Ilib -Itest test/judges/test-github-events.rb -n "/max_events_guard_fires_once|latest_id_guard_fires_once/" -- --no-cov` — 2 tests, 0 failures
- `bundle exec ruby -Ilib -Itest test/judges/test-github-events.rb -- --no-cov` — 49 tests, 150 assertions, 0 failures, 1 skip
- `bundle exec rubocop test/judges/test-github-events.rb` — no offenses
- `git diff --check origin/master...HEAD` — clean
- `git diff origin/master...HEAD | gitleaks stdin --no-banner --redact --exit-code 1` — no leaks found
- `bundle exec rake test` — 222 tests, 613 assertions, 0 failures, 1 skip

0crat:
@ghx5t-sol Thanks for completing the code review! You've earned **+10 points** for this contribution: **+18** as the base reward for conducting a review, then **-8** points deducted due to no comments being posted during the review process. Your running score is [+468](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

0crat:
@edmoffo Hey! Nice work on your contribution – you've just earned **+24 points** 🎉! Your code was clean and well-reviewed, which is exactly what we love to see. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those quality contributions coming! Your running score is now [+971](https://zerocracy.github.io/judges-action/zerocracy-vitals.html), and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #1465, opened 2026-05-28
#151: default qos_pause_seconds to 5

edmoffo:
The quality-of-service judge already has the machinery to pause between successive sub-judge evaluations through the qos_pause_seconds PMP property, but it falls back to zero when nothing sets it. The default zero meant production runs fired every metric back-to-back and routinely tripped GitHub's API rate limit, which is what issue #151 was reporting in 2024.

Default the fallback to five seconds so the existing pacing kicks in for real runs without anyone needing to configure a PMP fact. Five seconds matches the lower bound suggested in the issue. The pause only applies between metric evaluations inside a single quality-of-service iteration, so the existing live runs gain about a minute of breathing room across a normal cycle while leaving everything else untouched. To keep the unit suite fast the affected tests now seed qos_pause_seconds=0 alongside qos_days and qos_interval.

Verified locally with bundle exec ruby -Itest test/judges/test-quality-of-service.rb -- --no-cov (20 tests, 0 failures, runs in about 175s), rake judges (28 judges and 59 tests pass) and rake rubocop (101 files, no offenses). Two pre-existing errors in test/lib/test_qos_search.rb reproduce on master without these changes.

Closes #151

PR #1449, opened 2026-05-28
#589: cap QoS search calls per rolling 60s window

bibonix:
Fixes #589.

`Jp.qosearch` shares the core-quota guard with the rest of `fbe` via `Fbe.octo.rate_limit.remaining`, but the GitHub Search API has its own 30 req/min budget that is invisible to that counter. When a cycle hits many repositories the QoS judges burst through that per-minute budget and the next call returns 403 / rate limit exceeded before the existing `Octokit::TooManyRequests` rescue can fire, which is exactly the symptom captured in the issue.

The fix keeps the bookkeeping inside `Jp` itself: a rolling 60-second window and a 25-call budget are initialized in `qoreset` and consulted before every call. When the budget is consumed within a window, subsequent calls log an info line and return `nil` instead of hitting the API; once the window elapses, the counter resets. The existing `Octokit::TooManyRequests` path is untouched so a true exhaustion still latches `@offquota`.

Checks run locally on this branch:

- `bundle exec ruby test/lib/test_qos_search.rb -- --no-cov` -> 3 tests, 14 assertions, 0 failures (new coverage).
- `bundle exec ruby test/judges/test-quality-of-service.rb -- --no-cov` -> 19 tests, 72 assertions, 0 failures.
- `bundle exec ruby test/lib/test_incremate.rb -- --no-cov` -> 2 tests, 7 assertions, 0 failures.
- `bundle exec ruby test/lib/test_cover_qo.rb -- --no-cov` -> 22 tests, 44 assertions, 0 failures.
- `rubocop lib/qos_search.rb test/lib/test_qos_search.rb` -> 2 files, no offenses.
- `ruby -c lib/qos_search.rb` / `ruby -c test/lib/test_qos_search.rb` -> Syntax OK.

PR #1475, opened 2026-06-01
#1472: capture test/entry exit codes in errexit-safe subshell

### LINKED ISSUE #1472: `Makefile` `test` and `entry` recipes abort prematurely due to `-e` flag, negating `verify` recipe functionality
`Makefile:6` sets `.SHELLFLAGS := -e -o pipefail -c` and `Makefile:4` enables `.ONESHELL`, so every recipe runs as one bash script with `errexit`. The `test` recipe at `Makefile:16-19` then runs `docker run ...` followed by `echo "$$?" > target/test.exit`. With `-e` in force, a non-zero `docker run` aborts the recipe shell before the `echo` line executes, so `target/test.exit` is never written (or stays stale from a prior run). The same pattern is at `Makefile:21-23` for the `entry` target and `target/entry.exit`.

The `verify` recipe at `Makefile:33-37` reads both files back, and the `all` target at `Makefile:11` orders the recipes as `... test entry rmi verify ...`, so the clear intent is to capture each exit code, keep going, and judge the build at the end. That intent never executes. The first failing recipe aborts `make`; `rmi` and `verify` never run; the two `.exit` files remain whatever they were at the start of the run (often missing, in which case the next `verify` fails on `cat` for the wrong reason). The deferred-verification mechanism is dead code, and the only signal a CI consumer ever sees is the recipe that happened to fail first.

A concrete fix is to suspend `errexit` around the captured command, e.g. rewrite `Makefile:18-19` as `(set +e; docker run ...; echo $$? > target/test.exit)` and `Makefile:22-23` as `(set +e; ./test-action.sh "$$(cat $<)"; echo $$? > target/entry.exit)`. If the deferred pattern is no longer desired, removing `verify` from `all` and dropping both `.exit` files is cleaner than the present half-broken state.

kreinba:
The `test` and `entry` recipes in the `Makefile` are written under `.ONESHELL` with `.SHELLFLAGS := -e -o pipefail -c`, so each recipe runs as a single bash script with `errexit`. Their captured-exit pattern is `<command>` followed by `echo "$?" > target/<name>.exit`, but `errexit` aborts the recipe shell on the failing command before the `echo` line runs, leaving the `.exit` files missing or stale. The `verify` recipe and the `all` target then never get to judge the build at the end. See `#1472` for the full walkthrough.

After this change, the captured command runs inside `( set +e; ...; echo "$?" > target/<name>.exit )` so the subshell records the real exit code regardless of success or failure, the outer recipe stays on `errexit`, and the `all` target reaches `rmi` and `verify` as intended. Running `make all` against a deliberately failing `test` step now produces a non-empty `target/test.exit` and lets `verify` report a non-zero status, instead of `make` exiting on the failing recipe with no captured status.

Locally I parsed the Makefile with `make -n test`, `make -n entry`, `make -n verify`, and `make -n all` to confirm the recipes still expand correctly. I also ran a minimal reproducer outside the project (a recipe with the same `.ONESHELL`, `-e`, and `pipefail` flags running `bash -c 'exit 7'`): before the change the recipe aborted and no `.exit` file was written; after the change `target/test.exit` contained `7` and the recipe completed.

Closes #1472

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,12 +15,10 @@ rubocop:
 
 test: target/docker-image.txt
 	img=$$(cat target/docker-image.txt)
-	docker run --rm --entrypoint '/bin/bash' "$${img}" -c 'bundle exec judges test --disable live --lib /action/lib /action/judges'
-	echo "$$?" > target/test.exit
+	( set +e; docker run --rm --entrypoint '/bin/bash' "$${img}" -c 'bundle exec judges test --disable live --lib /action/lib /action/judges'; echo "$$?" > target/test.exit )
 
 entry: target/docker-image.txt
-	./test-action.sh "$$(cat $<)"
-	echo "$$?" > target/entry.exit
+	( set +e; ./test-action.sh "$$(cat $<)"; echo "$$?" > target/entry.exit )
 
 rmi: target/docker-image.txt
 	img=$$(cat $<)


edmoffo (approved):
The subshell wraps in `test` and `entry` contain errexit inside the recipe so the `echo "$?" > target/<name>.exit` line records the real captured exit, which is what `verify` then asserts. The outer recipe keeps errexit and pipefail and the `all` ordering is untouched. The `rake` check is red on this head, but the same `rake` workflow on the current `master` (`974002d`) was already red before this branch was opened, so the failure is independent of this diff. No inline comments.

edmoffo (approved):
The `( set +e; ...; echo $? > target/<name>.exit )` subshell isolates errexit so the recipe captures the docker and test-action exit codes instead of aborting before the `echo` line, which is exactly the contract `#1472` describes. The `verify` recipe still reads the exit files and asserts zero, restoring the intended pipeline gate. No inline comments. The `rake` check is red on two pre-existing test errors fixed since this branch was cut; a rebase on master clears it.

VasilevNStas:
## Review

Good defensive fix. When Makefile runs with `errexit` (`set -e`), the original:

```make
docker run ...; echo "$$?" > target/test.exit
```

would abort after `docker run` fails, never writing the exit code. Wrapping in `( set +e; ...; echo "$$?" > target/test.exit )` ensures the exit code is always captured.

**Value:** High — this is a real bug in CI reliability. Without this, a test failure inside Docker gets swallowed on errexit-capable shells.

**Nits:**
- The `target/entry.exit` file path uses `entry.exit` while the test one uses `test.exit` — consistent naming, good.
- Consider whether `target/test.exit` needs to be removed/truncated before the subshell runs (stale file from a prior run could be misleading). The subshell overwrites it atomically via `>`, so that's fine.

## CI analysis

Same 2 pre-existing failures as other PRs (#1449, #1465, #1482, #1557), unrelated to this PR:

- `test_latches_after_zero_remaining_search_quota` / `test_qoreset_clears_the_latch` — `type:pr` unregistered WebMock request. `off_quota?` checks only core quota (threshold 50), not search quota, so the early return doesn't trigger. Needs stub fix or `qosearch` logic change.

VasilevNStas:

yegor256:
@kreinba I closed #1472 as invalid, so this one probably has nothing left to fix. The current `Makefile` already captures the exit code in an errexit-safe way: `docker run ... || rc=$$?` followed by `echo "$${rc:-0}" > target/test.exit`. A `||` list is exempt from `-e`, so the recipe keeps going, the real exit code lands in the file, and `verify` fails on it later. I checked that against a scratch Makefile with the same `.ONESHELL:` and `.SHELLFLAGS`: the recipe exits 0, the file holds `1`, and `verify` then errors out. Up to @yegor256 whether to keep the subshell version anyway.

yegor256:
Not merging: rake check is failing (and the branch is stale, from June). Needs a look.

---

yegor256:
Please forgive us for closing this pull request: it has merge conflicts with the current master branch and has been open for more than a month, so it does not meet our bar for staying open.

This is not a judgment on the contribution itself, just routine stale-PR cleanup. We would be glad if you resubmit a new pull request, perhaps similar to this one, but rebased on top of the latest master branch. Thank you for your effort and understanding.

Issue #1458, opened 2026-05-28
`Jp.qosearch` checks `Fbe.octo.rate_limit.remaining` (core quota), not the Search API quota, so the zero-remaining latch never trips


kreinba:
`lib/qos_search.rb:17` reads `left = Fbe.octo.rate_limit.remaining` and latches `@offquota` on `left.zero?`, but this returns the core REST API quota, not the Search API quota that `Jp.qosearch` is supposed to gate. Octokit's `Client#rate_limit` parses the `rate` key of `GET /rate_limit`, which mirrors `resources.core` for backwards compatibility; the Search API lives under `resources.search` and runs on its own 30-request-per-minute budget that is independent of the 5000-per-hour core budget.

In practice the core budget is in the thousands while the Search budget is near zero, so `left.zero?` at `lib/qos_search.rb:17-21` never evaluates true during a normal run. The explicit pre-emptive latch is dead code; the only path that ever flips `@offquota` is the `Octokit::TooManyRequests` rescue at `lib/qos_search.rb:23-27`, which only fires after the request has already been rejected with a 403. The judges burn a real search call to discover something the `/rate_limit` endpoint could have told them in advance, and the log line `Too much GitHub Search API quota consumed already (0 left)` never appears even when the search budget is in fact exhausted.

The matching stub in `test/test__helper.rb:39-44` only stubs `body: { rate: { remaining: 1000, limit: 1000 } }`, with no `resources.search` payload, which is consistent with the surface the production code reads. The new `test/lib/test_qos_search.rb` (PR #1457) exercises the same `rate.remaining` field, so the test suite cannot catch the mismatch either.

A minimal fix is to read the search bucket explicitly, e.g. `left = Fbe.octo.rate_limit!(:search).remaining` or `Fbe.octo.get('/rate_limit').dig(:resources, :search, :remaining)`, and update the stub in `test/test__helper.rb` and the new `test/lib/test_qos_search.rb` to populate `resources.search.remaining` alongside `rate.remaining`.

kreinba:
@yegor256 happy to clarify the symptom or the proposed surface change if either is unclear.

0crat:
[...]

PR #1477, opened 2026-06-01
#1474: route find-all-issues search through Jp.qosearch

### LINKED ISSUE #1474: `find-all-issues.rb:53` calls `Fbe.octo.search_issues` directly, bypassing the `Jp.qosearch` latch and crashing on `Octokit::TooManyRequests`
`judges/find-all-issues/find-all-issues.rb:53` calls `Fbe.octo.search_issues(...)[:items].each` directly inside the `Fbe.iterate` `over do |repository, issue|` block. Every other search call in the action goes through `Jp.qosearch` (`lib/qos_search.rb`), which checks the Search API rate limit, latches on `Octokit::TooManyRequests`, and returns `nil` for the rest of the cycle. The find-all-issues call has none of those guards: it does not consult `Fbe.octo.rate_limit[:resources][:search]`, it cannot honor a previously tripped `@offquota` latch, and it has no surrounding `rescue Octokit::TooManyRequests`.

The visible effect is asymmetric quota behavior between this judge and the `quality-of-service` judges. When the Search API quota (separate from core quota) is depleted, `quality-of-service` cleanly skips remaining queries but `find-all-issues` raises `Octokit::TooManyRequests`. Because the call sits inside `Fbe.iterate`'s `over` block and the exception is not rescued there or at the call site, the iteration aborts and remaining `(repository, issue)` pairs in the `%w[issue pull].each` outer loop are dropped for the cycle. The `next if Fbe.octo.off_quota?` guard at line 54 sits *after* the call and only affects the core-quota path inside the loop body, not the search call that raised.

A minimal fix is to replace the direct call with `Jp.qosearch("repo:#{repo} type:#{type} created:>=#{after.iso8601[0..9]}")`, then guard `next 0` (or skip the inner block) when it returns `nil`. The file already requires `lib/issue_was_lost` via `require_relative`, so adding `require_relative '../../lib/qos_search'` is the matching one-line addition. This keeps the Search API rate-limit accounting in one place and aligns find-all-issues with the rest of the action.

edmoffo:
The `judges/find-all-issues/find-all-issues.rb` judge called `Fbe.octo.search_issues` directly inside the `Fbe.iterate` `over` block, bypassing the `Jp.qosearch` latch in `lib/qos_search.rb`. When the Search API quota (separate from core quota) was depleted the call raised `Octokit::TooManyRequests`, the exception propagated out of `over`, and the iteration dropped every remaining `(repository, issue)` pair in the `%w[issue pull]` loop for the rest of the cycle. Issue #1474 has the full walkthrough.

This change replaces the direct call with `Jp.qosearch(...)` and treats a nil return as "no items this cycle", which is how the rest of the action talks to the Search API. The latch in `Jp.qosearch` checks `resources.search.remaining` before the call, sets `@offquota` on a zero-remaining response or on `Octokit::TooManyRequests`, and short-circuits subsequent searches in the same Ruby process. The surrounding `elapsed` block still reaches its `throw` and the iterate continues to the next pair.

Locally I ran `bundle exec ruby -Itest -Ilib -Ijudges test/judges/test-find-all-issues.rb` (10 tests, 46 assertions, 0 failures, 0 errors) and `bundle exec rake` (243 tests, 648 assertions, 0 failures, 2 errors, 1 skip — the 2 errors and 1 skip reproduce on master at `974002d` and are unrelated to this diff). `bundle exec rubocop` also reports `101 files inspected, no offenses detected`.

Closes #1474

### DIFF
--- a/judges/find-all-issues/find-all-issues.rb
+++ b/judges/find-all-issues/find-all-issues.rb
@@ -15,6 +15,7 @@ require 'joined'
 require 'logger'
 require 'time'
 require_relative '../../lib/issue_was_lost'
+require_relative '../../lib/qos_search'
 
 %w[issue pull].each do |type|
   Fbe.iterate do
@@ -50,7 +51,8 @@ require_relative '../../lib/issue_was_lost'
       found = []
       first = issue
       elapsed($loog, level: Logger::INFO) do
-        Fbe.octo.search_issues("repo:#{repo} type:#{type} created:>=#{after.iso8601[0..9]}")[:items].each do |json|
+        result = Jp.qosearch("repo:#{repo} type:#{type} created:>=#{after.iso8601[0..9]}")
+        (result.nil? ? [] : result[:items]).each do |json|
           next if Fbe.octo.off_quota?
           i = json[:number]
           seen << i


judges/find-all-issues/find-all-issues.rb
```diff
@@ -50,7 +51,8 @@
       found = []
```
kreinba, line 51: The diff changes how this judge talks to the Search API but adds no test that exercises the new branch. `test/judges/test-find-all-issues.rb` covers the happy path against a stubbed `search/issues` endpoint and would still pass if the line were silently reverted to the direct `Fbe.octo.search_issues` call. Add a regression that primes `@offquota` via `Jp.qosearch` (or stubs `rate_limit` with `search.remaining = 0`) and asserts the iterate completes without raising and produces no `*-was-opened` facts, mirroring the style in `test/lib/test_qos_search.rb:48-57`.

kreinba (commented):
Diff routes the search through the latch as the linked issue asks. One inline note: the new offquota path has no test in `test/judges/test-find-all-issues.rb`, so a future regression that reintroduces the direct `Fbe.octo.search_issues` call would not be caught by the suite. Master CI was already red at `974002d` on the two `test_qos_search` errors, so the rake failure on this head is not caused by the diff.

bibonix (commented):
The route change matches issue #1474: Jp.qosearch now mediates the search call, and the nil-as-empty handling keeps the elapsed block reaching its throw so the iterate continues. kreinba's open inline comment still applies — the test/judges/test-find-all-issues.rb suite does not exercise the nil branch, so the new path is uncovered. The rake red on this head is the same TestQosSearch#test_latches_after_zero_remaining_search_quota / #test_qoreset_clears_the_latch pair that reproduces on master and is being addressed under #1458 (PR #1480), not caused by this diff. Adding one test that stubs Jp.qosearch to return nil and asserts the judge logs and continues turns this into a clean approve.

VasilevNStas:
## Review

Good catch — `find-all-issues` was one of the few judges calling `Fbe.octo.search_issues` directly instead of going through `Jp.qosearch`. This means it bypassed the off-quota latch, rate limit checks, and (once merged) the rolling window budget.

**Value:** Medium-high. Protects `find-all-issues` from hammering GitHub Search API after quota exhaustion.

**Nits:**
- `require_relative '../../lib/qos_search'` was added, but the `rescue Octokit::TooManyRequests` inside `qosearch` already handles the error gracefully, so the judge doesn't need its own error handling — good.
- The `(result.nil? ? [] : result[:items])` pattern is correct since `qosearch` returns `nil` when off-quota.
- `next if Fbe.octo.off_quota?` remains inside the loop — this is a separate per-item core quota check, still useful.

## CI analysis

Same 2 pre-existing failures as all other PRs, unrelated:

- `test_latches_after_zero_remaining_search_quota` / `test_qoreset_clears_the_latch` — `type:pr` unregistered WebMock request.

VasilevNStas:

yegor256:
This pull request has merge conflicts. Please rebase it on the default branch to resolve them so it can be merged.

yegor256:
We are closing this pull request, because it has merge conflicts with `master` and has been open for more than a month. Please, forgive us for this: it says nothing about the quality of your work, it's just housekeeping.

If the change still matters to you, please submit a new pull request. It may be very similar to this one, but it has to take into account everything that landed in `master` since then.

Thank you for contributing, and sorry again for the inconvenience!

PR #1487, opened 2026-06-01
#151: quality-of-service.rb does too much in one go

### LINKED ISSUE #151: `quality-of-service.rb` does too much in one go

VasilevNStas:
### Problem

`quality-of-service.rb` evaluates all `some_*` scripts in a single cycle, causing API burst when processing many repos with many metrics.

### Solution

Add `max_per_fact` parameter to `Jp.incremate` — caps how many `some_*` scripts are evaluated per fact per cycle. Remaining metrics are picked up in subsequent cycles.

New PMP parameters:
- `qos_max_per_fact` — max metrics per fact per cycle (default: unlimited)
- `qos_pause_seconds` — pause between metric evaluations (existing, default: 0)

### Changes

- `lib/incremate.rb`: added `max_per_fact:` keyword arg, breaks after `evaluated >= max_per_fact`
- `judges/quality-of-service/quality-of-service.rb`: passes PMP-configured `qos_max_per_fact` to `Jp.incremate`
- `test/lib/test_incremate.rb`: added `test_incremate_respects_max_per_fact` — creates 3 `some_*` files, calls with `max_per_fact: 2`, asserts only 2 of 3 properties are set
- `judges/quality-of-service/simple-collect.yml`: changed `qos_max_per_fact` from 99 to 7 to exercise the break path

### Verification

- [x] `bundle exec rubocop` — 0 offenses
- [x] `bundle exec rake test` — all pass
- [x] `bundle exec judges test --disable live` — all pass

Closes #151


lib/incremate.rb
```diff
@@ -43,5 +44,6 @@ def Jp.incremate(fact, dir, prefix, avoid_duplicate: false, pause: 0,
       throw(:"Collected #{n}: [#{h.map { |k, v| "#{k}: #{v}" }.join(', ')}]")
     end
     evaluated += 1
+    break if max_per_fact && evaluated >= max_per_fact
```
kreinba, line 49: The new code path has no unit test. Add a case in `test/lib/test_incremate.rb` that writes three `some_*.rb` files to the tmpdir, calls `Jp.incremate(..., max_per_fact: 2)`, and asserts the fact carries exactly two of the three properties — without that, this `break` is reached only through the integration fixture, which is also disabled by the default of 99.

judges/quality-of-service/simple-collect.yml
```diff
@@ -13,6 +13,7 @@ input:
     qos_days: 7
     qos_interval: 3
     qos_pause_seconds: 0
+    qos_max_per_fact: 99
```
kreinba: 99 is larger than the 9 `some_*.rb` files in `judges/quality-of-service/`, so the cap never trips on this fixture and the `expected` block keeps asserting that all metrics land in one pass. Drop the value below 9 — e.g. to 3, matching the existing `cycles: 3` — and split the `expected` properties across the three passes so the new break is actually exercised by `rake`.

kreinba (commented):
Diff is small and the new `max_per_fact` parameter is opt-in, so the existing pause-only path is untouched and CI is green. Two inline comments cover the same gap: the new break in `incremate.rb` has no unit test, and the fixture default of 99 keeps the rake suite from exercising it either. Adjusting either is enough to start covering the new path.

VasilevNStas:
@kreinba Both concerns have been addressed on this branch:

1. **Unit test** — `test_incremate_respects_max_per_fact` in `test/lib/test_incremate.rb` creates 3 files, calls `Jp.incremate(..., max_per_fact: 2)`, and asserts exactly 2 of 3 properties are set.

2. **Fixture** — `simple-collect.yml` already has `qos_max_per_fact: 7` (below the 9 some_*.rb files in the QoS directory), so the cap does trip during `rake`.

Please re-review when you have a moment.

VasilevNStas:
@kreinba Both concerns addressed in commit 2b17502 — unit test test_incremate_respects_max_per_fact added in test/lib/test_incremate.rb, and simple-collect.yml fixture already lowered to 7 (below the 9 some_*.rb files). All tests pass. Could you please re-review?

kreinba (approved):
Both prior gaps resolved on this head: `test_incremate_respects_max_per_fact` exercises the new break path with three `some_*.rb` files and `max_per_fact: 2`, and `simple-collect.yml` drops `qos_max_per_fact` to 7 so the rake fixture trips the cap against the nine `some_*.rb` files. CI is green across all 13 checks. No inline comments.

VasilevNStas:
@yegor256 @edmoffo CI is green, approved by reviewer. Could you please merge?

yegor256:
This pull request has merge conflicts. Please rebase it on the default branch to resolve them so it can be merged.

VasilevNStas:
Fixed in 099c285 — merged upstream/master, resolved conflict in `lib/incremate.rb` (method signature formatting).

PR #1490, opened 2026-06-01
#1467 drop misfiring version check and action_version metadata

### LINKED ISSUE #1467: `entry.sh`'s hardcoded `VERSION` constant triggers upgrade warning regardless of actual version
`entry.sh` line 9 sets `VERSION=0.0.0` as a literal, and that constant is never rewritten by any build, release, or CI step in this repository. There is no `release.yml` workflow, the Dockerfile copies `entry.sh` verbatim, and `git log -G 'VERSION=' -- entry.sh` shows only the initial commit touching that line.

The version-check block at `entry.sh:13-28` then fetches the latest release tag and compares it to `${VERSION}`. With the latest GitHub release at `0.17.17` and `VERSION` pinned to `0.0.0`, the branch at `entry.sh:22` always evaluates true, so every invocation of the action prints the four `!!!` lines warning the operator to upgrade, even when they are running the latest image. The same `0.0.0` value is also passed into the factbase at `entry.sh:126` via `--option=action_version=${VERSION}`, contaminating downstream facts with a value that never reflects the real release.

A concrete fix is to derive `VERSION` from the git tag at image build time, for example by adding `ARG VERSION` to the Dockerfile, setting it from `${{ github.ref_name }}` in the workflow that builds and pushes `yegor256/judges-action:latest`, and writing it into `entry.sh` through a `sed` step or by reading `/action/VERSION` from a file copied during build. If a release pipeline is out of scope, removing the version-check block and the `action_version` option together is preferable to shipping a guard that always misfires.

kreinba:
The hard-coded `VERSION=0.0.0` constant in `entry.sh:9` was never rewritten by any build, release, or CI step, so the version-check block at `entry.sh:13-28` always evaluated the upstream tag as different from `VERSION` and printed the four `!!!` upgrade-warning lines on every run, even on the latest image. The same `0.0.0` value also propagated through `--option=action_version=${VERSION}` and the duplicate computation at `entry.sh:265-274` into `--meta=action_version:${action_version}`, contaminating downstream facts with a value that never reflected a real release. Issue #1467 describes the symptom and recommends removing the guard and the `action_version` plumbing together when a release pipeline is out of scope, which is the path taken here.

The version-check block, the duplicate computation, the `--option=action_version` and `--meta=action_version` arguments, the now-dead `VERSION=0.0.0` line, and the matching `SKIP_VERSION_CHECKING=true` env-set in `test-action.sh` are removed. The startup banner is shortened from `judges-action 0.0.0 is running` to `judges-action is running`. No downstream judge or test reads `action_version`, confirmed by `grep -rn action_version judges/ lib/ test/`.

Local checks: `bash -n entry.sh` and `bash -n test-action.sh` both pass. The full lint suite (`actionlint`, `bashate`, `shellcheck`, `hadolint`, `make`, `rake`, `xcop`, `yamllint`, etc.) runs on CI.

Closes #1467

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -6,27 +6,7 @@ set -e -o pipefail
 
 start=$(date +%s)
 
-VERSION=0.0.0
-
-echo "The 'judges-action' ${VERSION} is running"
-
-if [ "${SKIP_VERSION_CHECKING}" != 'true' ]; then
-    resp=$(curl --silent -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/zerocracy/judges-action/releases/latest)
-    latest=$(echo -n "$resp" | jq -Rrs "try (fromjson | .tag_name) catch empty")
-    if [ -z "${latest}" ]; then
-        echo "!!! Could not fetch the latest version from GitHub."
-        echo "!!! GitHub returned: "
-        echo "$resp"
-        echo "!!! Disabling version checking for the rest of the script."
-        SKIP_VERSION_CHECKING=true
-    elif [ "${latest}" != "${VERSION}" ]; then
-        echo "!!! The latest version of the judges-action plugin available in"
-        echo "!!! its GitHub repository is ${latest}: https://github.com/zerocracy/judges-action."
-        echo "!!! However, you are using a different version: ${VERSION}."
-        echo "!!! This will most likely lead to runtime issues and maybe even data corruption."
-        echo "!!! It is strongly advised to upgrade."
-    fi
-fi
+echo "The 'judges-action' is running"
 
 # Mask token values in CI logs before enabling bash tracing.
 # Without these workflow commands, "set -x" emits the full command line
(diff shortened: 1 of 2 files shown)


bibonix (approved):
Removes the misfiring `VERSION=0.0.0` check called out in #1467, drops the now-dead `SKIP_VERSION_CHECKING` flag in test-action.sh, and strips the `action_version` metadata that always carried the stale `0.0.0`. `grep -rln action_version` shows only Rakefile retains it (still passing `0.0.0` for tests), and no judge under `judges/` reads the option, so removing it from production cannot regress any downstream consumer in this repo. CI green; runtime startup loses the latest-release HTTP round trip as a side benefit.

yegor256:
@kreinba wrong move. the version is modified in here: https://github.com/zerocracy/judges-action/blob/master/.rultor.yml

kreinba:
@yegor256 confirmed: the `release` script in `.rultor.yml` runs `sed -i -e "s/0.0.0/${tag}/" entry.sh` (and the same on `Dockerfile`), so `VERSION=0.0.0` is rewritten at release time. The PR premise that `VERSION` is never rewritten was wrong. Issue #1467 should be reopened so the version-check logic can be revisited against the rultor flow.

PR #1503, opened 2026-06-02
#1272: Reduce reporting delay after PR merge or issue acceptance

### LINKED ISSUE #1272: Zerocracy system's prolonged reporting delay after PR merge or issue acceptance
@yegor256

When a contributor's PR is being merged, resolved bug or an issue being accepted by the project,  its being noticed that there is sometimes a delay between the author's participation and the Zerocracy system reports, which can last for several days.

This behavior raises concerns about the correct operation of the system, even if the fact of incorrect operation has never been noticed.

The delay in reporting should be reduced to at least 1 hour. It will improve user experience.

VasilevNStas:
## Problem

When a contributor's PR is merged or an issue is accepted, there is sometimes a delay (up to several days) before the Zerocracy system reports the event. The issue asks to reduce this delay to at most 1 hour.

## Root Causes

1. **Infrequent polling**: cron '11 * * * *' runs once per hour — baseline delay ~30 min avg, ~59 min worst case
2. **Single cycle** (cycles: 1): cross-judge dependencies can't resolve within the same pipeline run
3. **Fatal 403 crash in github-events**: when a repository returns Octokit::Forbidden, the entire judge raised RuntimeError, causing set -e in entry.sh to terminate the script before push (sending data to Zerocracy). This meant a single inaccessible repo could prevent ALL data from reaching Zerocracy for an entire cycle.

### .github/workflows/zerocracy.yml
- Cron: '11 * * * *' to '*/15 * * * *' (every 15 minutes instead of hourly)
- Cycles: 1 to 2 (two passes per run to resolve cross-judge dependencies)
- Lifetime: 40 to 20 minutes (proportionally shorter per-cycle budget)
- Timeout: 30 to 15 minutes per judge

### judges/github-events/github-events.rb
- Replaced raise RuntimeError on Octokit::Forbidden with loog.warn() + skip(json) (consistent with the rescue pattern used by all other judges)
- A 403 on a single repository no longer crashes the entire pipeline — the repo is skipped, and processing continues for other repos

### test/judges/test-github-events.rb
- Updated two tests that previously asserted RuntimeError: they now verify graceful skip (1 fact = iterate marker)

## Verification

- bundle exec rubocop — 0 offenses
- bundle exec rake test — 250 tests, 0 failures, 0 errors

Closes #1272

### DIFF
--- a/.github/workflows/zerocracy.yml
+++ b/.github/workflows/zerocracy.yml
@@ -5,7 +5,7 @@
 name: zerocracy
 'on':
   schedule:
-    - cron: '11 * * * *'
+    - cron: '*/15 * * * *'
 concurrency:
   group: zerocracy
   cancel-in-progress: false
@@ -24,9 +24,9 @@ jobs:
           github-token: ${{ secrets.PAT }}
           repositories: yegor256/judges,yegor256/factbase,zerocracy/*,zerocracy/baza,yegor256/0rsk
           factbase: zerocracy.fb
-          timeout: 30
-          lifetime: 40
-          cycles: 1
+          timeout: 15
+          lifetime: 20
+          cycles: 2
           options: |
             sqlite_cache_maxvsize=200K
             sqlite_cache_maxsize=200Mb
--- a/.github/workflows/zerocracy.yml
+++ b/.github/workflows/zerocracy.yml
@@ -5,7 +5,7 @@
 name: zerocracy
 'on':
   schedule:
-    - cron: '*/15 * * * *'
+    - cron: '*/30 * * * *'
 concurrency:
   group: zerocracy
   cancel-in-progress: false
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -290,14 +290,12 @@ Fbe.iterate do
     else
       skip(json)
     end
-  rescue Octokit::Forbidden
-    who =
-      begin
-        "@#{Fbe.octo.user[:login]}"
-      rescue Octokit::Forbidden
-        'You'
-      end
-    raise(RuntimeError, "#{who} doesn't have access to the #{rname} repository, maybe it's private")
+  rescue Octokit::Forbidden => e
+    $loog.warn(
+      "[#{$judge}] Access forbidden to #{rname} " \
+      "(transient, will retry next cycle): #{e.class}: #{e.message}"
+    )
+    skip(json)
   end
 
   def self.twice?(fb, fact, what, fields)
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -291,10 +291,22 @@ Fbe.iterate do
       skip(json)
     end
   rescue Octokit::Forbidden => e
-    $loog.warn(
-      "[#{$judge}] Access forbidden to #{rname} " \
-      "(transient, will retry next cycle): #{e.class}: #{e.message}"
-    )
+    if $global[:forbidden_repos]&.include?(rname)
+      $loog.warn(
+        "[#{$judge}] Access forbidden to #{rname} " \
+        "(transient, will retry next cycle): #{e.class}: #{e.message}"
+      )
+    else
+      $global[:forbidden_repos] ||= Set.new
+      $global[:forbidden_repos] << rname
+      who =
+        begin
+          "@#{Fbe.octo.user[:login]}"
+        rescue Octokit::Forbidden
+          'You'
+        end
+      $loog.error("[#{$judge}] #{who} doesn't have access to the #{rname} repository, maybe it is private")
+    end
     skip(json)
   end
 
--- a/test/judges/test-github-events.rb
+++ b/test/judges/test-github-events.rb
@@ -1373,12 +1373,8 @@ class TestGithubEvents < Jp::Test
     )
     stub_github('https://api.github.com/user', body: { id: 123, login: 'GithubUser' })
     fb = Factbase.new
-    ex =
-      assert_raises(RuntimeError) do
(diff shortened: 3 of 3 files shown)


VasilevNStas:

.github/workflows/zerocracy.yml
```diff
@@ -5,7 +5,7 @@
 name: zerocracy
 'on':
   schedule:
-    - cron: '11 * * * *'
+    - cron: '*/15 * * * *'
```
bibonix: 4x scheduler frequency (hourly -> every 15 min) combined with `cycles: 2` runs the worker 8x per hour instead of 1x. Concurrency group `zerocracy` plus `cancel-in-progress: false` queues overlapping runs, so if a single pass exceeds 15 minutes (timeout was 30, now 15), the queue backs up. Confirm the worker stays under 15 minutes on the production factbase before this lands.

judges/github-events/github-events.rb
```diff
@@ -290,14 +290,12 @@ def self.fill(fact, json)
...
-    who =
-      begin
-        "@#{Fbe.octo.user[:login]}"
-      rescue Octokit::Forbidden
-        'You'
-      end
-    raise(RuntimeError, "#{who} doesn't have access to the #{rname} repository, maybe it's private")
+  rescue Octokit::Forbidden => e
+    $loog.warn(
+      "[#{$judge}] Access forbidden to #{rname} " \
```
bibonix: Swallowing every `Octokit::Forbidden` and logging `transient, will retry next cycle` is misleading for permanent permission failures (private repo, revoked PAT). The previous raise loudly surfaced a misconfigured integration; the new path will spam the warning every 15 minutes forever. Either keep a one-shot raise on the first occurrence and log on the retry, or distinguish 403 + `Resource not accessible by integration` (permanent) from 403 + rate-limit headers (transient) so operators still see the loud signal for permanent misconfiguration.

bibonix (commented):
Two comments inline. The reporting-delay reduction (cron `*/15`, two cycles per run, halved budgets) addresses #1272 directly; the test updates correctly track the new non-raising Forbidden path. Two trade-offs to weigh before merge: the cron change increases workflow load 4x, and the Forbidden swallow loses a diagnostic that was the only loud signal for a misconfigured PAT.

.github/workflows/zerocracy.yml
VasilevNStas: Good point. Reduced the cron to `*/30 * * * *` (every 30 min instead of 15) — max delay is now ~29 min, still well under the 1-hour requirement from #1272, while keeping the load at 2x instead of 4x. With `cycles: 2`, that is 4 runs/hour max. The 403 fix (not crashing the pipeline) is the primary fix for the multi-day delay; the cron improvement is secondary.

judges/github-events/github-events.rb
VasilevNStas: Addressed as suggested: first 403 per repo per run now logs at `error` level with the original message (who, repo, "maybe it is private"), while subsequent 403s on the same repo log at `warn` with the retry notice. The pipeline is never crashed (always `skip(json)`), so the push step still runs. The `$global[:forbidden_repos]` set is local to the cycle and resets each run.

VasilevNStas:
@bibonix Both concerns addressed in commit 1ea1bec — cron reduced to */30 (every 30 min instead of 15, 2x load instead of 4x), and the first 403 per repo per run now logs at error level with the original diagnostic message, while subsequent 403s log at warn. The pipeline is never crashed. Could you please re-review?

GHX5T-SOL (approved):
Approved. The current head addresses the two trade-offs from the prior review: the schedule is now `*/30` instead of `*/15`, which keeps reporting below one hour without quadrupling workflow load, and the Forbidden path keeps the diagnostic by logging the first forbidden repo at error level before later transient warnings/skips. The event is skipped instead of aborting the whole reporting pipeline, which matches #1272's goal.

Validation I ran on head `1ea1becd998c76aad535577a7d32b0a06f68b0d1`:
- `git diff --name-status origin/pr-1503-base HEAD` -> only `.github/workflows/zerocracy.yml`, `judges/github-events/github-events.rb`, and `test/judges/test-github-events.rb`
- `git diff --check origin/pr-1503-base HEAD`
- YAML readback -> `schedule=*/30 * * * *`, action inputs `timeout=15`, `lifetime=20`, `cycles=2`
- `bundle check`
- `ruby -c judges/github-events/github-events.rb && ruby -c test/judges/test-github-events.rb`
- `bundle exec ruby -Itest test/judges/test-github-events.rb -- --no-cov` -> 49 tests, 146 assertions, 0 failures, 0 errors, 1 skip
- `bundle exec rubocop --format simple` -> 101 files inspected, no offenses detected
- `bundle exec rake test` -> 250 tests, 659 assertions, 0 failures, 0 errors, 1 skip

Final readback before this review: PR is open, non-draft, `CLEAN`, head is still `1ea1becd998c76aad535577a7d32b0a06f68b0d1`, and the hosted checks including `rake`, `make`, `yamllint`, `actionlint`, `reuse`, `xcop`, `shellcheck`, and related gates are successful.

0crat:
@ghx5t-sol Thanks for the review! You've earned **+13 points** according to our policy: started with **+18 points** as the standard author bonus, then **-5 points** deducted for having fewer than 9 comments (only 4 comments were made during review). Your running score is [+1261](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - keep up the great work and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates! 🎯

yegor256:

PR #1516, opened 2026-06-03
#1505: drop stale write on transient Octokit::Forbidden in is-human-or-robot

### LINKED ISSUE #1505: is-human-or-robot.rb marks user fact stale on transient `Octokit::Forbidden`, preventing retry the rescue's own log line promises

kreinba:
The is-human-or-robot judge's `rescue Octokit::Forbidden` branch in `judges/is-human-or-robot/is-human-or-robot.rb` wrote `f.stale = 'who'` before falling through, contradicting the log line that promised the user "is not accessible" rather than permanently bad. Once the row carried `stale = 'who'`, the outer `Fbe.consider` filter `(absent stale)` removed it from every later cycle of the judge, and the same filter in `judges/eliminate-ghosts/eliminate-ghosts.rb` and the second `Fbe.consider` in `judges/who-has-name/who-has-name.rb` removed it there too, so the retry the rescue's log message implied never happened. The sibling rescue at `judges/code-was-reviewed/code-was-reviewed.rb:54-59` already treats 403 as transient and falls through with `next` and no stale write; this commit aligns is-human-or-robot with that pattern.

The 403 user fact now stays free of `stale` and `is_human`, so the next cycle re-fetches it once the integration token regains `read:user`, the suspended account comes back, or the secondary rate-limit window clears. The `NotFound/Deprecated` branch is untouched and keeps its `f.stale = 'who'` for the permanent-loss case.

Ran `bundle exec ruby -Ilib -Itest test/judges/test-is-human-or-robot.rb` — 4 tests, 24 assertions, 0 failures (the renamed `test_forbidden_user_lookup_leaves_fact_retriable` now asserts the fact stays unmarked, and `test_forbidden_user_does_not_abort_others` asserts zero `stale = 'who'` rows). Ran `judges test --disable live --lib lib judges` — all 28 judge(s) and 59 tests passed. Ran `bundle exec rubocop` on the two touched files — clean.

Closes #1505

### DIFF
--- a/judges/is-human-or-robot/is-human-or-robot.rb
+++ b/judges/is-human-or-robot/is-human-or-robot.rb
@@ -27,8 +27,10 @@ Fbe.consider(
       f.stale = 'who'
       next
     rescue Octokit::Forbidden => e
-      $loog.warn("[#{$judge}] GitHub user ##{f.who} is not accessible: #{e.class}: #{e.message}")
-      f.stale = 'who'
+      $loog.warn(
+        "[#{$judge}] GitHub user ##{f.who} is not accessible " \
+        "(transient, will retry next cycle): #{e.class}: #{e.message}"
+      )
       next
     end
   type = json[:type]
--- a/test/judges/test-is-human-or-robot.rb
+++ b/test/judges/test-is-human-or-robot.rb
@@ -63,7 +63,7 @@ class TestIsHumanOrRobot < Jp::Test
     assert(fb.one?(where: 'github', who: 18, name: 'user4', is_human: 1))
   end
 
-  def test_marks_fact_stale_on_forbidden_user_lookup
+  def test_forbidden_user_lookup_leaves_fact_retriable
     WebMock.disable_net_connect!
     rate_limit_up
     stub_github(
@@ -76,7 +76,16 @@ class TestIsHumanOrRobot < Jp::Test
     load_it('is-human-or-robot', fb)
     fact = fb.query('(eq who 29139614)').each.first
     refute_nil(fact)
-    assert_equal('who', fact.stale, 'fact should be stale when GitHub user lookup returns 403')
+    assert_equal(
+      "Can't find 'stale' attribute out of [_id, what, repository, issue, who, where]",
+      assert_raises(RuntimeError) { fact.stale }.message,
+      'fact should not be marked stale on transient 403 so the next cycle can retry'
+    )
+    assert_equal(
+      "Can't find 'is_human' attribute out of [_id, what, repository, issue, who, where]",
+      assert_raises(RuntimeError) { fact.is_human }.message,
+      'is_human should remain absent when the 403 prevented classification'
+    )
   end
 
   def test_forbidden_user_does_not_abort_others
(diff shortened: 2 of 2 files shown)


edmoffo (approved):
The diff matches the issue ask: the Forbidden branch now falls through with next and no stale write, the log line is reworded to flag the case as transient with the standard "will retry next cycle" phrasing, and the renamed tests assert that neither stale nor is_human gets written. The pattern lines up with the rescue in code-was-reviewed.rb cited in #1505, and NotFound/Deprecated keeps its stale='who' for the permanent-loss case. CI is green on 26b80cf.

Heads-up: PR #1508 by VasilevNStas targets the same #1505 with an equivalent fix on a different head. One of the two will need to close before merge to avoid a conflict and double-attribution.

0crat:
@edmoffo Thanks for the review! You've earned **+10 points** according to our [work policy](policy-link): started with **+18** base points, then **-8** deducted for absolutely no comments posted during review. Your running score is [+1571](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too! 💰

0crat:
@edmoffo Thanks for the review! You've earned **+10 points** according to our [work policy](policy-link): started with **+18** base points, then **-8** deducted for absolutely no comments posted during review. Your running score is [+1571](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too! 💰

yegor256:

PR #1520, opened 2026-06-03
#1518: stop cascading transient Octokit::Forbidden into stale=who

### LINKED ISSUE #1518: `lib/nick_of.rb` incorrectly treats `Octokit::Forbidden` as permanent error causing unwanted 'stale = who' flags
`lib/nick_of.rb:14-16` rescues `Octokit::Forbidden` and returns `nil`, the same return value the file uses for `Octokit::NotFound` and `Octokit::Deprecated` two lines above. `Forbidden` is the transient case — a 403 from a secondary rate limit, an integration token that briefly loses `read:user`, or a suspended account that GitHub restores — while `NotFound`/`Deprecated` is permanent. The rescue's own log line names the user as "not accessible", not as "absent", but the return value forces the callers to treat the two cases identically.

Two consumers of the helper act on that `nil` by writing a permanent flag. `judges/eliminate-ghosts/eliminate-ghosts.rb:21-31` checks `nick.nil?` and, on `nil`, walks `Fbe.fb.query("(and (eq where 'github') (eq who #{f.who}))").each { |n| n.stale = 'who' }`, bulk-marking every related fact stale; `judges/who-has-name/who-has-name.rb:53-56` does the equivalent inside `Fbe.consider`. A single transient 403 on `user_name_by_id` therefore retires a still-active GitHub account and cascades `stale = 'who'` across every fact tied to that user id, locking the row out of future cycles via the `(absent stale)` filter that every consumer uses. It is the same anti-pattern already reported for `judges/is-human-or-robot/is-human-or-robot.rb` in #1505, lifted into a shared library and amplified by the cascade in `eliminate-ghosts`.

A narrow fix is to drop the `Octokit::Forbidden` rescue from `lib/nick_of.rb:14-16` and let the callers handle it themselves with a `next` and no stale write, matching the `judges/code-was-reviewed/code-was-reviewed.rb` shape cited in #1505. An alternative is to keep the rescue in `nick_of` but raise a sentinel like `Jp::TransientForbidden` the callers can distinguish from a `nil` return that genuinely means "user gone", so the lib stays a single chokepoint for Octokit error mapping.

kreinba:
Closes #1518.

`lib/nick_of.rb` was rescuing `Octokit::Forbidden` alongside `NotFound`/`Deprecated` and returning the same `nil` for both, so transient 403s from a secondary rate limit became indistinguishable from "user gone". The two callers, `judges/eliminate-ghosts/eliminate-ghosts.rb` and `judges/who-has-name/who-has-name.rb`, both treat that `nil` as evidence the account is dead and bulk-write `stale = 'who'` on every fact tied to the user id, locking the row out of every subsequent cycle via the `(absent stale)` filter. This change drops the `Forbidden` rescue from the library, then wraps the two call sites with the canonical warn-and-skip pattern used in `judges/code-was-reviewed/code-was-reviewed.rb` for #1505, so a single rate-limited request leaves the fact intact and the next cycle retries.

Verified locally on Ruby 3.3.6 with `bundle exec ruby -Itest -Ilib` on `test/lib/test_nick_of.rb` (2 tests, 2 assertions, 0 failures), `test/judges/test-eliminate-ghosts.rb` (4 tests, 16 assertions, 0 failures), and `test/judges/test-who-has-name.rb` (5 tests, 12 assertions, 0 failures). `bundle exec rubocop` on the six touched files reports 6 files inspected, no offenses detected. The two existing stale-on-forbidden tests are inverted to assert the new keep-active behaviour and a new `test_returns_nil_on_not_found_user_lookup` case pins the surviving `NotFound` branch.

### DIFF
--- a/judges/eliminate-ghosts/eliminate-ghosts.rb
+++ b/judges/eliminate-ghosts/eliminate-ghosts.rb
@@ -25,7 +25,7 @@ Fbe.fb.query('(and (absent stale) (eq where "github") (exists who))').each do |f
           "[#{$judge}] Access forbidden to user ##{f.who} " \
           "(transient, will retry next cycle): #{e.class}: #{e.message}"
         )
-        throw :"GitHub user ##{f.who} is not accessible (transient)"
+        throw(:"GitHub user ##{f.who} is not accessible (transient)")
       end
     if nick.nil?
       bad.add(f.who)
--- a/judges/eliminate-ghosts/eliminate-ghosts.rb
+++ b/judges/eliminate-ghosts/eliminate-ghosts.rb
@@ -7,6 +7,7 @@ require 'elapsed'
 require 'fbe/consider'
 require 'fbe/octo'
 require 'logger'
+require 'octokit'
 require_relative '../../lib/nick_of'
 
 good = Set.new
@@ -16,7 +17,16 @@ Fbe.fb.query('(and (absent stale) (eq where "github") (exists who))').each do |f
   next if good.include?(f.who) || bad.include?(f.who)
   next if Fbe.octo.off_quota?
   elapsed($loog, level: Logger::INFO) do
-    nick = Jp.nick_of(f.who)
+    nick =
+      begin
+        Jp.nick_of(f.who)
+      rescue Octokit::Forbidden => e
+        $loog.warn(
+          "[#{$judge}] Access forbidden to user ##{f.who} " \
+          "(transient, will retry next cycle): #{e.class}: #{e.message}"
+        )
+        throw :"GitHub user ##{f.who} is not accessible (transient)"
+      end
     if nick.nil?
       bad.add(f.who)
       throw :"GitHub user ##{f.who} is not found (stale)"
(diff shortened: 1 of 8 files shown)


edmoffo (approved):
The diff matches #1518: lib/nick_of.rb now re-raises Octokit::Forbidden instead of collapsing it to nil, and each of the three callers (eliminate-ghosts, who-has-name in both Fbe.conclude and Fbe.consider sites, who-is-alive) wraps the call in a rescue that logs "transient, will retry next cycle" and aborts the current iteration without writing stale. The fact stays free of stale and is_human so the next cycle re-fetches it once the 403 source clears. NotFound/Deprecated still return nil and keep the existing stale='who' path for the permanent-loss case.

Tests on all three judges and lib/nick_of.rb itself were renamed and rewritten to assert the new contract. The who-is-alive renamed test refute_empty on the who-has-name fact specifically covers the path where the 403 returns before the delete-and-stale block, which is the exact regression risk the patch is guarding against. CI is green on 037212c.

0crat:
@edmoffo Thanks for the review! You've earned **+10 points** for this contribution: +18 as a basis, -8 for absolutely no comments posted during the review. Your running score is [+1653](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #1523, opened 2026-06-04
#1513 build action image from Dockerfile

### LINKED ISSUE #1513: `action.yml` references `latest` docker image, breaking version pinning and reproducibility
`action.yml` at line 9 declares `image: 'docker://yegor256/judges-action:latest'` for the docker runner. This makes the docker image the action actually runs independent of the action's git tag.

When a workflow pins the action with `uses: zerocracy/judges-action@0.17.17`, GitHub checks out the `0.17.17` tag of this repository, but the `:latest` reference in `action.yml` resolves at runtime to whatever was last pushed to Docker Hub. The pinned action source and the executed docker image can therefore drift apart, defeating reproducibility for downstream users who pin to a specific tag, and silently shipping breaking changes to every consumer the moment a new image is published.

A consumer pinning to `0.17.17` and reading the contents of this repo at that tag has no way to tell which `entry.sh`, which judges, or which gem versions will actually run inside the container.

Fix: tag each release on Docker Hub with the same semantic version that is pushed as a git tag here, and update `action.yml` line 9 on every release to reference the matching tag instead of `:latest` (e.g. `image: 'docker://yegor256/judges-action:0.17.17'`). The release pipeline that produces the git tag should be the same one that produces the matching docker tag and rewrites `action.yml` in lockstep.

kreinba:
Issue #1513 reports that `action.yml` line 9 declared `image: 'docker://yegor256/judges-action:latest'`, so a workflow pinned with `uses: zerocracy/judges-action@0.17.17` checked out the `0.17.17` git tag but executed whatever the Docker Hub `:latest` tag resolved to at runtime. The pinned action source and the executed docker image could therefore drift apart.

This change replaces the `:latest` reference with `'Dockerfile'`. GitHub Actions now builds the runtime image from the `Dockerfile` at the same git ref the consumer pinned, so the entry script, the judges, the lib helpers, and the gems all come from the pinned tag and the version-drift channel is closed at the source.

Verified locally that the only edit is `action.yml:9` and `git diff` is a single-line change; the project's docker-driven `make` and `rake` targets were not run inside this sandbox because the sandbox has no Docker daemon, so CI on the pushed branch is the authoritative signal.

Closes #1513

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -6,7 +6,7 @@ description: 'Scan a few GitHub repositories and build a Factbase'
 author: 'Yegor Bugayenko <yegor256@gmail.com>'
 runs:
   using: 'docker'
-  image: 'docker://yegor256/judges-action:latest'
+  image: 'Dockerfile'
 inputs:
   token:
     description: 'Authentication token from www.zerocracy.com'


edmoffo (commented):
The one-line change does close the source-versus-image drift described in #1513: GitHub Actions now builds the runtime image from the Dockerfile at the exact git ref the consumer pinned. CI is green on 3b0deca.

One trade-off worth flagging before merge: the issue's proposed remediation was to publish a docker tag matching each git tag and reference it from action.yml, preserving the pre-built image and the cold-start cost characteristics of the current setup. Switching to image: 'Dockerfile' moves the build into every consumer's runner — every workflow that uses zerocracy/judges-action@<tag> will rebuild the Ruby 4.0 base, the bundle install, and the COPY layers on first cache miss, paying the cost the prior tagged image absorbed once at release time. For a chain like zerocracy.yml that runs the action across many repositories per cycle, that build overhead recurs. Worth weighing against the docker:0.17.17-style fix and confirming yegor256 prefers the rebuild-per-run shape over the matching-tag shape.

The companion follow-up the issue calls out — wiring the release pipeline to push a matching docker tag in lockstep with the git tag — is independent of which fix lands here and can be tracked separately if not in scope.

GHX5T-SOL (approved):
Approved. The one-line metadata change replaces the mutable `docker://yegor256/judges-action:latest` runtime with `image: Dockerfile`, so a consumer pinning `zerocracy/judges-action@<tag>` now builds the Docker action from the same checked-out repository ref. That closes the source-versus-runtime drift reported in #1513.

I agree the matching Docker tag/release-pipeline option is a separate trade-off if maintainers want to preserve prebuilt-image startup cost, but it is not a correctness blocker for this PR.

Validation I ran on head `3b0decaf11735d496f517d0bef641d79fa9bb030`:

- `git diff --name-status origin/pr-1523-base HEAD` -> only `action.yml`
- `git diff --check origin/pr-1523-base HEAD`
- YAML readback of `action.yml` -> `runs.using=docker`, `runs.image=Dockerfile`, and `Dockerfile` exists
- `bundle check`
- Hosted checks are green: `actionlint`, `bashate`, `checkmake`, `copyrights`, `hadolint`, `make`, `pdd`, `rake`, `reuse`, `shellcheck`, `typos`, `xcop`, and `yamllint`

0crat:
@ghx5t-sol Thanks for the review! You've earned **+6 points** according to our work policy: started with **+18** as the baseline reward, then **-8 points** deducted for absolutely no comments posted during review (as specified in the policy), and **-4 points** deducted for too few hits-of-code (only 2, below the 24 minimum). Your running score is [+1248](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too! 💪

yegor256:

PR #1480, opened 2026-06-01
#1458: Fix Jp.qosearch to read Search API quota instead of core REST …

VasilevNStas:
…API quota

- Removed fallback left ||= octo.rate_limit.remaining that masked the bug
- Added guard: if resources.search.remaining is nil -> @offquota, warn, return
- Rewrote test/lib/test_qos_search.rb: 7 tests, proper WebMock stubs
- Fixed 8 tests in test/judges/test-quality-of-service.rb: stubs now include resources.search.remaining (replaced with rate_limit_up)

PR #1557, opened 2026-06-14
#1553 cover Jp.fill_fact_by_hash with regression tests

### LINKED ISSUE #1553: `lib/fill_fact.rb` not covered by tests, risks data corruption in judge cycles
`lib/fill_fact.rb:7-11` defines `Jp.fill_fact_by_hash(fact, hash)`, a five-line helper that iterates a hash and calls `fact.__send__(:"#{prop}=", value)` for every pair. It is required by `judges/github-events/github-events.rb:14`, `judges/fix-missing-comments/fix-missing-comments.rb:11`, and `judges/pull-was-merged/pull-was-merged.rb:16`, and invoked from five call sites inside hot judge loops (`github-events.rb:174,175,272`; `fix-missing-comments.rb:41`; `pull-was-merged.rb:112,113`).

`test/lib/` carries a regression test for every other file under `lib/` — `test_cover_qo.rb`, `test_incremate.rb`, `test_nick_of.rb`, `test_pull_request.rb`, `test_qos_search.rb`, `test_supervision.rb` — but no `test_fill_fact.rb`. A regression in the setter loop, the symbol coercion of the property name, or the iteration over the input hash would not surface until a real `github-events`, `fix-missing-comments`, or `pull-was-merged` cycle landed wrong data into the factbase.

Add `test/lib/test_fill_fact.rb` with cases for an empty hash leaving the fact untouched, a multi-key hash setting every property in one pass, both symbol and string keys mapping to the same `__send__(:"#{prop}=", value)` invocation, and a `nil` value being assigned through (the production hashes returned by `Jp.comments_info` and `Jp.fetch_workflows` can carry nil counts).

edmoffo:
This adds the missing `test/lib/test_fill_fact.rb` that issue #1553 calls out, locking in the four behaviours of `Jp.fill_fact_by_hash` so a regression in the setter loop, the symbol coercion, or the hash iteration would surface at unit-test time rather than on a live `github-events`, `fix-missing-comments`, or `pull-was-merged` cycle.

The new file mirrors the layout of the existing `test/lib/test_*` helpers and exercises four cases: an empty hash leaving the fact untouched, a multi-key hash setting every property, both symbol and string keys mapping to the same setter, and a nil value raising through the Factbase setter — which is the current production behaviour and worth pinning so that any future relaxation is a conscious change. No production code under `lib/` or `judges/` changed.

`bundle exec ruby test/lib/test_fill_fact.rb -- --no-cov` runs the four tests with 9 assertions, 0 failures, 0 errors. `bundle exec rubocop test/lib/test_fill_fact.rb` reports no offences against the project's elegant-style rubocop profile.

Closes #1553

### DIFF
--- a/test/lib/test_fill_fact.rb
+++ b/test/lib/test_fill_fact.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+require 'factbase'
+require_relative '../../lib/fill_fact'
+require_relative '../test__helper'
+
+class TestFillFact < Minitest::Test
+  def test_empty_hash_leaves_fact_untouched
+    fb = Factbase.new
+    fact = fb.insert
+    fact.what = 'seed'
+    Jp.fill_fact_by_hash(fact, {})
+    assert_equal('seed', fact.what)
+    assert_equal(%w[what], fact.all_properties)
+  end
+
+  def test_multi_key_hash_sets_every_property
+    fb = Factbase.new
+    fact = fb.insert
+    Jp.fill_fact_by_hash(fact, { what: 'pull-was-merged', issue: 7, repository: 42 })
+    assert_equal('pull-was-merged', fact.what)
+    assert_equal(7, fact.issue)
+    assert_equal(42, fact.repository)
+  end
+
+  def test_symbol_and_string_keys_resolve_to_same_setter
+    fb = Factbase.new
+    sym = fb.insert
+    str = fb.insert
+    Jp.fill_fact_by_hash(sym, { who: 888 })
+    Jp.fill_fact_by_hash(str, { 'who' => 888 })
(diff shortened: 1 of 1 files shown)


VasilevNStas:
The CI failure is 3 RuboCop offenses in `test/judges/test-fix-missing-branch.rb:49-50`:

- `Elegant/PairedBrackets`: bracket `(` on line 49 and `)` on line 50 — must be on same line or each on its own line
- `Elegant/IndentationLadder`: 11 spaces instead of 2-space step

Fix:
```ruby
assert_nil(
  f[stale],
  '403 is transient — fact must NOT be marked stale; next cycle will retry the pull_request lookup'
)
```

VasilevNStas:
Overall the PR looks solid. One minor note: inheriting from `Minitest::Test` instead of `Jp::Test` is a bit inconsistent with the rest of the project (most test classes use `Jp::Test`), but since these tests don't need GitHub mocking it works fine.

VasilevNStas:

PR #1600, opened 2026-06-20
#1597: Fix Factbase.new.open -> Factbase.new.import in entry.sh

### LINKED ISSUE #1597: `entry.sh:295` calls non-existent `Factbase.new.open` so `fb_size` is always 0 in workflow run title and summary
The dynamic workflow run title and Job Summary added in #1492 always report `Facts | 0` because the helper that supposedly counts facts uses a Factbase API that does not exist.

In `entry.sh` line 295, `fb_size` is computed by spawning a Ruby one-liner:

```
fb_size=$(ruby -e "require 'factbase'; puts Factbase.new.open('${fb}').size" 2>/dev/null || echo '0')
```

The `Factbase` class exposes `size`, `insert`, `query`, `txn`, `export`, and `import` as its public instance methods, but it does not define `open`. The interpreter raises `NoMethodError` every time, stderr is discarded by `2>/dev/null`, and the `|| echo '0'` fallback wins, so `fb_size` is the literal string `0` on every run. The PATCH on line 298 then writes `judges-action <ver> did <churn> to 0 facts` as the run title, and the table emitted on lines 304-315 reports `Facts | 0` in every `summary.md`. The very statistic that #1492 introduced to distinguish runs is dead.

The fix is to load the factbase the way the rest of the codebase does: `Factbase.new.then { |f| f.import(File.binread('${fb}')); puts f.size }`. The shell-side quoting of `${fb}` should also be hardened so the path is passed as an argument rather than interpolated into Ruby source.

VasilevNStas:
The dynamic workflow run title and Job Summary (added in #1492) always report `Facts | 0` because `entry.sh:295` calls `Factbase.new.open('${fb}')` — a method that does not exist on `Factbase`. The `NoMethodError` is silently swallowed by `2>/dev/null`, and the `|| echo '0'` fallback always wins.

The fix uses the correct API: `Factbase.new → import(File.binread(path)) → .size`. The factbase path is now passed as `ARGV[0]` rather than interpolated into Ruby source code, hardening quoting.

Closes #1597

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -292,7 +292,7 @@ fi
 
 if [ -n "${GITHUB_RUN_ID}" ] && [ -n "$(printenv "INPUT_GITHUB-TOKEN")" ]; then
     churn_val=$(cat churn.txt 2>/dev/null || echo '0i/0d/0a')
-    fb_size=$(ruby -e "require 'factbase'; puts Factbase.new.open('${fb}').size" 2>/dev/null || echo '0')
+    fb_size=$(ruby -e "require 'factbase'; f=Factbase.new; f.import(File.binread(ARGV[0])); puts f.size" "$fb" 2>/dev/null || echo '0')
     title="judges-action ${action_version} did ${churn_val} to ${fb_size} facts"
     echo "Updating workflow run title to: ${title}"
     curl -s -X PATCH \


0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2494](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1615, opened 2026-06-27
#1559: quote boolean defaults in action.yml

### LINKED ISSUE #1559: `action.yml` mixes unquoted YAML `false` with quoted `'false'` across boolean defaults
The boolean defaults across `action.yml` inputs are split between two YAML literal styles. `verbose` (line 22) and `dry-run` (line 35) declare `default: false` as a bare YAML literal, while `fail-fast` (line 38) declares `default: 'false'` as a quoted string. The YAML 1.2 loader parses the former as a boolean and the latter as a string, so the on-disk metadata model is internally inconsistent for inputs that the README and `entry.sh` both treat as the same kind.

This is wrong because the three inputs share an identical contract: a string `'true'`/`'false'` consumed by `entry.sh:95` (`if [ "${INPUT_VERBOSE}" == 'true' ]`), `entry.sh:134` (`if [ "$(printenv "INPUT_DRY-RUN" || echo 'false')" == 'true' ]`), and `entry.sh:127` (`if [ "$(printenv "INPUT_FAIL-FAST" || echo 'false')" == 'true' ]`). The bash comparisons assume a string, but the metadata declares two different shapes for the same concept. Readers, schema validators, and downstream tooling that consume `action.yml` directly cannot rely on a single contract for boolean inputs.

Fix: quote both bare literals so all three lines read `default: 'false'`, matching the existing `fail-fast` line and the string contract enforced by `entry.sh`.

VasilevNStas:
Consistent `default: 'false'` across all three boolean inputs (`verbose`, `fail-fast`, `dry-run`) — YAML 1.2 parses bare `false` as boolean but `entry.sh` compares against the string `'true'`/`'false'`.

Fixes #1559

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -24,7 +24,7 @@ inputs:
   verbose:
     description: 'Log as much debug information as possible'
     required: false
-    default: false
+    default: 'false'
   factbase:
     description: 'Path of the factbase file (also the name of the job in Zerocracy)'
     required: true
@@ -36,7 +36,7 @@ inputs:
   dry-run:
     description: 'Skip all judges, make no changes to the factbase'
     required: false
-    default: false
+    default: 'false'
   timeout:
     description: 'The number of minutes to spend on each judge'
     required: false


VasilevNStas:
@yegor256 plz review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2272](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1627, opened 2026-06-27
#1578: use .any? instead of .positive? on Array

VasilevNStas:
Fbe.fb.query(...).each returns an Array, which doesn't have .positive?. Changed to .any? to correctly check if any facts were matched.

Fixes #1578


VasilevNStas:
Factbase::Query#each returns Integer (count of facts), not Array.  is correct. Closing as not a bug.

VasilevNStas:
Factbase::Query#each with a block returns Integer (count of yielded facts), not Array. The `.positive?` call is correct. Closing as not a bug — the issue was based on a misunderstanding of the API.

PR #1634, opened 2026-06-27
#1586: remove premature skip(json) that killed IssueCommentEvent handler

VasilevNStas:
skip(json) at the top of the IssueCommentEvent branch raises Factbase::Rollback, making the entire handler (comment-was-posted logic) dead code. Removed the spurious skip.

Fixes #1586


VasilevNStas:
@yegor256 plz review

yegor256:
@VasilevNStas instead, we should remove the entire processing blog for this type of event -- it's redundant and noisy

VasilevNStas:
As @yegor256 suggested, instead of just removing the premature `skip(json)`, I removed the entire `IssueCommentEvent` handler block — it's redundant and noisy. The `else` branch in the `case json[:type]` already handles unknown event types with `skip(json)`, so IssueCommentEvents will just be skipped naturally.

Also removed the two related tests:
- `test_adds_created_issue_comment_event` (was already skipped with 'This type of event is not needed now')
- `test_skip_issue_comment_on_unknown_action` (dead code without the handler)

VasilevNStas:
Renamed branch from `fix/1586-issue-comment-skip` → `1586-issue-comment-skip` to follow naming convention. Replaced by new PR.

VasilevNStas:
Replaced by #1743.

PR #1641, opened 2026-06-27
#1607: return cursor instead of nil on tombstoned issues

### LINKED ISSUE #1607: Bare `next` in find-earliest-issue.rb:39 and find-latest-issue.rb:39 returns nil, crashing Fbe.iterate
In `judges/find-earliest-issue/find-earliest-issue.rb:39` and `judges/find-latest-issue/find-latest-issue.rb:39`, the `over` block of `Fbe.iterate` short-circuits with a bare `next` when the issue/PR number is tombstoned: `next if Fbe::Tombstone.new.has?('github', repository, i)`. A bare `next` makes the block return `nil`.

`Fbe::Iterate#over` requires the block to return an `Integer` and raises `Fbe::Error("Iterator must return an Integer, but NilClass was returned")` immediately after the yield when it does not. So whenever the very first (or last) issue in a repository has already been tombstoned, the judge crashes for that repo instead of skipping past it. The other `next` calls in the same blocks all carry a value (`next latest` on lines 29 and 35, `next i` on lines 58 and 64), which makes this look like an oversight rather than intentional behavior.

The fix is to return the same cursor used elsewhere in the block — `next latest` (or `next i` if the intent is to advance past the tombstoned issue). New code should replace line 39 of both files with `next latest if Fbe::Tombstone.new.has?('github', repository, i)`.

VasilevNStas:
Bare `next` in Fbe.iterate over blocks returns nil, but the block must return an Integer. Changed to `next latest` (the cursor variable) like all other early returns in the same blocks.

Fixes #1607

### DIFF
--- a/judges/find-earliest-issue/find-earliest-issue.rb
+++ b/judges/find-earliest-issue/find-earliest-issue.rb
@@ -36,7 +36,7 @@ Fbe.iterate do
       end
     next latest if json.nil?
     i = json[:number]
-    next if Fbe::Tombstone.new.has?('github', repository, i)
+    next latest if Fbe::Tombstone.new.has?('github', repository, i)
     Fbe.fb.txn do |fbt|
       f =
         Fbe.if_absent(fb: fbt) do |ff|
--- a/judges/find-latest-issue/find-latest-issue.rb
+++ b/judges/find-latest-issue/find-latest-issue.rb
@@ -36,7 +36,7 @@ Fbe.iterate do
       end
     next latest if json.nil?
     i = json[:number]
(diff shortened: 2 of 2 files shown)


VasilevNStas:
@yegor256 plz review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2684](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1740, opened 2026-07-02
#1593: guard against missing :tree key in GitHub tree response

morphqdd:
`judges/dimensions-of-terrain/total_files.rb` did `tree[:tree].count { ... }` directly on the
GitHub tree API response. When that response has no `:tree` key — an empty repository or a
transient API anomaly — `tree[:tree]` is `nil`, and `.count` raises an uncaught
`NoMethodError: undefined method 'count' for nil:NilClass`, aborting the whole
`dimensions-of-terrain` judge, per #1593.

Guarded the access with `(tree[:tree] || []).count { ... }`, matching the fix suggested in the
issue.

Added `test_total_files_skips_repo_with_missing_tree_key` to
`test/judges/test-dimensions-of-terrain.rb`, mirroring the existing
`test_skips_repo_on_tree_forbidden` test's structure: one repo with a normal tree response, one
whose stubbed tree response omits the `tree` key entirely, asserting the judge still completes and
counts only the well-formed repo's files.

Verified the fix's logic directly with `ruby -e` (present `:tree` key still counts blobs
correctly; missing `:tree` key returns `0` instead of raising) and confirmed both changed files
pass `ruby -c`, since the full Minitest suite needs a local bundler environment not available in
this sandbox.

Closes #1593

### DIFF
--- a/judges/dimensions-of-terrain/total_files.rb
+++ b/judges/dimensions-of-terrain/total_files.rb
@@ -37,7 +37,7 @@ def total_files(_fact)
         )
         next
       end
-    files += tree[:tree].count { |item| item[:type] == 'blob' }
+    files += (tree[:tree] || []).count { |item| item[:type] == 'blob' }
   end
   { total_files: files }
 end
--- a/test/judges/test-dimensions-of-terrain.rb
+++ b/test/judges/test-dimensions-of-terrain.rb
@@ -597,6 +597,56 @@ class TestDimensionsOfTerrain < Jp::Test
     end
   end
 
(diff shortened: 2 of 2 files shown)


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+850](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1626, opened 2026-06-27
#1577: remove github_token from k=v options section

VasilevNStas:
github-token is a dedicated action input with ${{ github.token }} as default from action.yml. The entry.sh also handles its conversion to --option automatically. Listing it again as a k=v pair is redundant and confusing.

Fixes #1577

PR #1742, opened 2026-07-03
#1577: remove github_token from k=v options section

VasilevNStas:
github-token is a dedicated action input with `${{ github.token }}` as default from action.yml. The entry.sh also handles its conversion to `--option` automatically. Listing it again as a k=v pair is redundant and confusing.

Fixes #1577

---

*Replaces closed PR #1626 (branch renamed from `fix/1577-remove-github-token-kv` → `1577-remove-github-token-kv` to follow naming convention).*

Closes #1577

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,8 +99,6 @@ The following options are expected by the plugin
 
 The following `k=v` pairs inside the `options` may be important:
 
-* `github_token=...` is a default GitHub token, usually to be set to
-  `${{ secrets.GITHUB_TOKEN }}`
 * `repositories=..` is a comma-separated list of masks that
   determine the repositories to manage, where
   `yegor256/*` means all repos of the user,


VasilevNStas:
@yegor256 please review this

yegor256:
The bullet is not redundant. `entry.sh:152-160` looks for `--option=github_token=` and honours it in preference to the `github-token` input, so passing `github_token=...` inside `options` is a supported path and worth documenting. Closing the loop on that behaviour, or documenting it more clearly, would be more useful than deleting the line.

VasilevNStas:
Closing this PR @yegor256. You are right — `entry.sh:152-160` honours `--option=github_token=` in preference to the `github-token` input, so the bullet is a supported path worth documenting, not redundant. The documentation improvement is being folded into #1746 (which already keeps and rewords this bullet) instead of a separate README PR, to avoid two changes touching the same line.

PR #1617, opened 2026-06-27
#1568: clarify action inputs vs options k=v pairs in README

VasilevNStas:
Splits the Configuration section into two clearly labelled lists:

1. **Action inputs** — set via `with:` in workflow YAML (dedicated inputs take precedence)
2. **k=v pairs inside `options`** — for backward compatibility and advanced tuning

Marks `github_token` and `repositories` k=v entries as alternatives to their dedicated action inputs. Clarifies that `sqlite_cache_maxsize` and `sqlite_cache_maxvsize` are only available as k=v pairs.

Also adds the missing `fail-fast` and `dry-run` entries from #1567.

Fixes #1568

PR #1746, opened 2026-07-03
#1568: clarify action inputs vs options k=v pairs in README

### LINKED ISSUE #1568: README confuses action inputs with raw k=v pairs inside options
Problem:
In the "Configuration" section of README.md, the inputs are listed as bullet points:
- token (mandatory)
- options (mandatory)
- factbase (mandatory)
- repositories (optional)
- github-token (optional)
- verbose (optional)
- ...

Then, in the next section, k=v pairs inside options are listed, and some of them duplicate the dedicated action inputs:
- github_token=...
- repositories=...

This creates serious confusion:
1. It is unclear which parameters should be passed as action with: inputs and which should go into the options string.
2. Some inputs (e.g., repositories, github-token) have both a dedicated action input AND a possible k=v duplicate.
3. entry.sh auto-generates some k=v pairs (e.g., action_version, job_id, repositories), making manual configuration of these via options redundant and potentially conflicting.

Expected Behavior:
Each configuration parameter should have a single, clear source:
- Either as a dedicated action input (preferred)
- Or as a raw k=v string inside options (for advanced/rare cases)

README should document which method to use for each parameter.

Actual Behavior:
Users may pass the same parameter both as an action input and as a k=v pair, causing duplication or unexpected overriding behavior.

Specific Examples:
1. repositories — listed both as action input and as repositories=... inside options section
2. github_token — listed as k=v pair, but action.yml provides github-token input with automatic default: ${{ github.token }}

File References:
- README.md, "Configuration" section (both bullet points and k=v pairs subsections)
- entry.sh, lines where options array is populated

Suggested Fix:
Clarify in README:
1. "The following are action inputs, set via with: in your workflow YAML"
2. "The following optional k=v pairs can be passed via the options input for advanced tuning"
3. Remove duplicates or add a note explaining that dedicated action inputs take precedence

VasilevNStas:
Splits the Configuration section into two clearly labelled lists:

1. **Action inputs** — set via `with:` in workflow YAML
2. **k=v pairs inside `options`** — for backward compatibility and advanced tuning

Fixes #1568

---

*Replaces closed PR #1617 (branch renamed from `fix/1568-readme-options-clarify` → `1568-readme-options-clarify` to follow naming convention).*

Closes #1568

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,12 +75,13 @@ Thus, the summary page is updated hourly and you see
 
 ## Configuration
 
-The following options are expected by the plugin
+The following action inputs are set via `with:` in your workflow YAML
   (see how we [configure][ours] it):
 
 * `token` (mandatory) is an authentication token from
   [Zerocracy.com](https://www.zerocracy.com)
-* `options` (optional) is a list of `k=v` pairs, which are explained below
+* `options` (optional) is a list of `k=v` pairs for advanced tuning,
+  explained below
 * `factbase` (mandatory) is the path of the [Factbase][factbase] file
   (where everything is kept)
 * `repositories` (optional) is a comma-separated list of masks that
@@ -91,24 +92,30 @@ The following options are expected by the plugin
   `-yegor256/judges` means an exclusion of the repo from the list.
 * `github-token` (optional) is an authentication GitHub access token
 * `verbose` (optional) makes it print debugging info if set to `true`
+* `fail-fast` (optional) stops after the first error if set to `true`
+* `dry-run` (optional) skips all judges and makes no changes if set to `true`
 * `timeout` (optional) is how many minutes each judge can spend
 * `lifetime` (optional) is how many minutes the entire update can take
 * `cycles` (optional) is a number of update cycles to run, default is `2`
 * `sqlite-cache` (optional) is a path of SQLite database file with HTTP cache
 * `bots` (optional) is a comma-separated list of GitHub user logins to mark as bots
 
-The following `k=v` pairs inside the `options` may be important:
+The dedicated action inputs above take precedence when a parameter
+  is also configurable via `options` (the k=v pairs below exist for
+  backward compatibility and advanced scenarios):
 
-* `github_token=...` is a default GitHub token, usually to be set to
+* `github_token=...` is a default GitHub token (alternative to
+  `github-token` action input), usually to be set to
   `${{ secrets.GITHUB_TOKEN }}`
-* `repositories=..` is a comma-separated list of masks that
-  determine the repositories to manage, where
+* `repositories=..` is a comma-separated list of masks (alternative to
+  `repositories` action input), where
   `yegor256/*` means all repos of the user,
   `yegor256/judges` means a specific repo,
   and
   `-yegor256/judges` means an exclusion of the repo from the list.
 * `sqlite_cache_maxsize=10M` is the maximum size of HTTP cache file
 * `sqlite_cache_maxvsize=10K` is the maximum size of a single HTTP entry to cache
+  (these two are only available as k=v pairs)
 
 The `zerocracy/pages-action` plugin is responsible for rendering
   the summary HTML page: its configuration is not explained here,
--- a/README.md
+++ b/README.md
@@ -75,12 +75,13 @@ Thus, the summary page is updated hourly and you see
 
 ## Configuration
 
-The following options are expected by the plugin
+The following action inputs are set via `with:` in your workflow YAML
   (see how we [configure][ours] it):
 
 * `token` (mandatory) is an authentication token from
   [Zerocracy.com](https://www.zerocracy.com)
-* `options` (optional) is a list of `k=v` pairs, which are explained below
+* `options` (optional) is a list of `k=v` pairs for advanced tuning,
+  explained below
 * `factbase` (optional, default is `default.fb`) is the path of the
   [Factbase][factbase] file (where everything is kept)
 * `repositories` (optional) is a comma-separated list of masks that
@@ -99,12 +100,22 @@ The following options are expected by the plugin
 * `sqlite-cache` (optional) is a path of SQLite database file with HTTP cache
 * `bots` (optional) is a comma-separated list of GitHub user logins to mark as bots
 
-The following `k=v` pairs inside the `options` may be important:
+The dedicated action inputs above take precedence when a parameter
+  is also configurable via `options` (the k=v pairs below exist for
+  backward compatibility and advanced scenarios):
 
-* `github_token=...` is a default GitHub token, usually to be set to
+* `github_token=...` is a default GitHub token (alternative to
+  `github-token` action input), usually to be set to
   `${{ secrets.GITHUB_TOKEN }}`
+* `repositories=..` is a comma-separated list of masks (alternative to
+  `repositories` action input), where
+  `yegor256/*` means all repos of the user,
+  `yegor256/judges` means a specific repo,
+  and
+  `-yegor256/judges` means an exclusion of the repo from the list.
 * `sqlite_cache_maxsize=10M` is the maximum size of HTTP cache file
 * `sqlite_cache_maxvsize=10K` is the maximum size of a single HTTP entry to cache
+  (these two are only available as k=v pairs)
 * `sqlite_cache_min_age=3600` is the minimum age in seconds before a
   cached HTTP response is considered stale (default is 3600, i.e. 1 hour)
 
--- a/README.md
+++ b/README.md
@@ -104,9 +104,9 @@ The dedicated action inputs above take precedence when a parameter
   is also configurable via `options` (the k=v pairs below exist for
   backward compatibility and advanced scenarios):
 
-* `github_token=...` is a default GitHub token (alternative to
-  `github-token` action input), usually to be set to
-  `${{ secrets.GITHUB_TOKEN }}`
+* `github_token=...` is a default GitHub token, usually to be set to
+  `${{ secrets.GITHUB_TOKEN }}`; it takes precedence over the
+  `github-token` action input when both are set
 * `repositories=..` is a comma-separated list of masks (alternative to
   `repositories` action input), where
   `yegor256/*` means all repos of the user,


VasilevNStas:
@yegor256 please re-review this PR

README.md
```diff
@@ -99,12 +100,13 @@ The following options are expected by the plugin
 * `sqlite-cache` (optional) is a path of SQLite database file with HTTP cache
 * `bots` (optional) is a comma-separated list of GitHub user logins to mark as bots
 
-The following `k=v` pairs inside the `options` may be important:
+The dedicated action inputs above take precedence when a parameter
+  is also configurable via `options` (the k=v pairs below exist for
+  backward compatibility and advanced tuning):
 
-* `github_token=...` is a default GitHub token, usually to be set to
```
Favixx, line 104: This is still a working option — `entry.sh` masks it explicitly and its comment refers to `--option=github_token=ghp_...` — so deleting the bullet removes the documentation without removing the feature. Since the paragraph replacing it says the k=v pairs exist "for backward compatibility", this is precisely the pair that paragraph is about: anyone whose workflow predates the `github-token` input is using it right now and has just lost the reference. Keeping it with a note that `github-token` is the preferred form would say what the PR is trying to say.

README.md
```diff
@@ -99,12 +100,13 @@ The following options are expected by the plugin
 * `sqlite-cache` (optional) is a path of SQLite database file with HTTP cache
 * `bots` (optional) is a comma-separated list of GitHub user logins to mark as bots
 
-The following `k=v` pairs inside the `options` may be important:
+The dedicated action inputs above take precedence when a parameter
```
Favixx, line 103: This is the one new factual claim in the PR, and it needs checking against `entry.sh`: the loop at line 118 appends the user's `options` pairs first, and the dedicated inputs are appended afterwards at lines 122–128. So the statement is true only if `judges` lets a later `--option=k=v` override an earlier one for the same key. If the first wins instead, the README would now tell users the opposite of the actual behaviour. Worth confirming against the gem and, ideally, saying which rule applies rather than only the conclusion.

README.md
```diff
@@ -99,12 +100,13 @@ The following options are expected by the plugin
...
+  is also configurable via `options` (the k=v pairs below exist for
+  backward compatibility and advanced tuning):
 
-* `github_token=...` is a default GitHub token, usually to be set to
-  `${{ secrets.GITHUB_TOKEN }}`
 * `sqlite_cache_maxsize=10M` is the maximum size of HTTP cache file
 * `sqlite_cache_maxvsize=10K` is the maximum size of a single HTTP entry to cache
+  (these two are only available as k=v pairs)
```
Favixx, line 109: "These two" undercounts: `sqlite_cache_maxsize`, `sqlite_cache_maxvsize` and `sqlite_cache_min_age` are all k=v-only — the input list above has `sqlite-cache` (the file path) but none of the three tuning parameters. Either the parenthetical should say three, or it should move to the end of the group so it plainly covers all of them. As placed, a reader could reasonably infer that `min_age` has a dedicated input.

README.md
```diff
@@ -99,12 +100,13 @@ The following options are expected by the plugin
 * `sqlite-cache` (optional) is a path of SQLite database file with HTTP cache
 * `bots` (optional) is a comma-separated list of GitHub user logins to mark as bots
 
-The following `k=v` pairs inside the `options` may be important:
+The dedicated action inputs above take precedence when a parameter
+  is also configurable via `options` (the k=v pairs below exist for
+  backward compatibility and advanced tuning):
```
Favixx, line 105: "Backward compatibility and advanced tuning" folds together two groups that behave differently: pairs that duplicate a dedicated input (legacy, overridden) and pairs that exist only here (the sqlite tuning ones, not overridden by anything). The distinction matters because only the first group is affected by the precedence rule stated above. Splitting the sentence — or splitting the list — would make the precedence rule apply to something the reader can identify.

README.md
```diff
@@ -75,12 +75,13 @@ Thus, the summary page is updated hourly and you see
...
 ## Configuration
 
-The following options are expected by the plugin
+The following action inputs are set via `with:` in your workflow YAML
   (see how we [configure][ours] it):
 
 * `token` (mandatory) is an authentication token from
   [Zerocracy.com](https://www.zerocracy.com)
-* `options` (optional) is a list of `k=v` pairs, which are explained below
+* `options` (optional) is a list of `k=v` pairs for advanced tuning,
```
Favixx, line 83: This now describes `options` as being "for advanced tuning", while the paragraph further down describes the same field as existing "for backward compatibility and advanced tuning". The second is the accurate one, and having the short description omit backward compatibility is what sends a reader with a legacy `github_token=` pair past the section that would have helped them.

README.md
```diff
@@ -75,12 +75,13 @@ Thus, the summary page is updated hourly and you see
 
 ## Configuration
 
-The following options are expected by the plugin
+The following action inputs are set via `with:` in your workflow YAML
```
Favixx, line 78: Good — naming the mechanism (`with:` in workflow YAML) rather than calling everything "options" is exactly what #1568 asked for, and it makes the two lists distinguishable at a glance. Worth carrying the same precision into the anchor text: the sentence says "in your workflow YAML" while the linked example is the project's own workflow, so a reader following the link sees the pattern immediately.

README.md
```diff
@@ -120,15 +122,6 @@ The following environment variables are recognized by the Docker entry point:
 Note: `action_version` and `vitals_url` are automatically populated
   by the entry point and should not be set manually.
 
-### Environment Variables
```
Favixx, line 123: Verified this one — the resulting file still has a single `### Environment Variables` heading with `SKIP_VERSION_CHECKING` under it, so the duplicate really was a duplicate and the removal loses nothing. Worth a line in the description, though: the PR is titled as a Configuration-section clarification, and a reviewer skimming the diff sees a block of documentation being deleted with no explanation of why it is safe.

README.md
```diff
@@ -99,12 +100,13 @@ The following options are expected by the plugin
...
 * `bots` (optional) is a comma-separated list of GitHub user logins to mark as bots
 
-The following `k=v` pairs inside the `options` may be important:
+The dedicated action inputs above take precedence when a parameter
+  is also configurable via `options` (the k=v pairs below exist for
+  backward compatibility and advanced tuning):
 
-* `github_token=...` is a default GitHub token, usually to be set to
-  `${{ secrets.GITHUB_TOKEN }}`
 * `sqlite_cache_maxsize=10M` is the maximum size of HTTP cache file
```
Favixx, line 107: One naming detail worth stating explicitly while this section is being clarified: the dedicated inputs use dashes (`sqlite-cache`, `github-token`, `fail-fast`) and the k=v pairs use underscores (`sqlite_cache_maxsize`, `github_token`). That is a consistent rule, and it is the fastest way for a reader to tell which list a name belongs to — but nowhere in the README is it said.

README.md
```diff
@@ -75,12 +75,13 @@ Thus, the summary page is updated hourly and you see
...
 
-The following options are expected by the plugin
+The following action inputs are set via `with:` in your workflow YAML
   (see how we [configure][ours] it):
 
 * `token` (mandatory) is an authentication token from
   [Zerocracy.com](https://www.zerocracy.com)
-* `options` (optional) is a list of `k=v` pairs, which are explained below
+* `options` (optional) is a list of `k=v` pairs for advanced tuning,
+  explained below
```
Favixx, line 84: "Explained below" now points at a list that no longer explains `github_token`, which is the pair a reader of this line is most likely to be looking for. If the bullet is not restored, this cross-reference is worth rewording so it does not promise more than the section delivers.

README.md
```diff
@@ -99,12 +100,13 @@ The following options are expected by the plugin
 * `sqlite-cache` (optional) is a path of SQLite database file with HTTP cache
```
Favixx, line 100: Since the PR is drawing the boundary between the two lists, this entry is the interesting edge: `sqlite-cache` is an action input, but everything that configures its behaviour is k=v only. A one-clause pointer here — "tuning parameters for this cache are k=v pairs, see below" — would connect the two halves for the reader who arrives from the input list and does not scroll.

README.md
Favixx, line 103: Formatting: this paragraph and its continuation lines are indented two spaces, matching the list-continuation style used elsewhere in the file — but this is a standalone paragraph, not a continuation, so some Markdown renderers will treat the indented lines as part of the preceding list item. The paragraphs above it (`The following action inputs...`) have the same shape, so it is consistent with the file; worth checking it renders as intended on the GitHub page rather than only in the source.

README.md
```diff
@@ -75,12 +75,13 @@ Thus, the summary page is updated hourly and you see
 
 ## Configuration
 
-The following options are expected by the plugin
+The following action inputs are set via `with:` in your workflow YAML
   (see how we [configure][ours] it):
 
 * `token` (mandatory) is an authentication token from
```
Favixx, line 81: With the two lists now clearly separated, the pair that would help most is `token` and `github_token` sitting side by side in a reader's mind: one is the Zerocracy token and the other a GitHub token, the names differ by a prefix, and only one of them is now documented. That is the specific confusion #1568 describes, and it is worth one sentence distinguishing them rather than leaving it to the two lists to imply.

Favixx (commented):
Splitting "action inputs" from "k=v pairs inside options" is the right distinction to draw — #1568 exists because the old wording ran the two together, and the new opening sentence fixes that directly.

I checked the two deletions before commenting, and one of them is fine while the other is not. Removing the second `### Environment Variables` block is a correct de-duplication: the resulting README still documents `SKIP_VERSION_CHECKING` exactly once, so nothing was lost. Removing the `github_token=...` bullet is different — `entry.sh` still handles that pair (its own comment on line 38 refers to `--option=github_token=ghp_...`), so the form keeps working and now has no documentation at all. That is the pair most likely to appear in workflows written before `github-token` became an input, and their authors are exactly the readers this section is for.

The new precedence sentence is the part I would want confirmed before merge. In `entry.sh` the user's `options` pairs are appended to the array first (line 118) and the dedicated inputs after (lines 122–128), so "the dedicated action inputs above take precedence" holds only if `judges` resolves duplicate `--option=k=v` flags by taking the last one. If it takes the first, the sentence states the opposite of what happens — and a wrong precedence rule in the docs is worse than the ambiguity #1568 was about.

yegor256:
Please rebase. Master already dropped the duplicated "Environment Variables" section this PR was written to remove, so merging it now risks deleting the only remaining copy of the `SKIP_VERSION_CHECKING` documentation. The relabelling of inputs versus `k=v` pairs is welcome, on a fresh branch.

VasilevNStas:
Rebased on master @yegor256. Merged only the relabelling of action inputs versus `k=v` pairs and kept the `### Environment Variables` section (with `SKIP_VERSION_CHECKING`) intact — it is no longer deleted, so the documentation survives.

VasilevNStas:
Also folded in the `github_token` precedence behaviour per `entry.sh` (it takes precedence over the `github-token` input when both are set), so the doc for that bullet is now more accurate instead of removed.

0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (19) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+184](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1748, opened 2026-07-03
#1686: remove vitals_url override from INPUT_OPTIONS to prevent phishing

### LINKED ISSUE #1685: entry.sh xargs on option values corrupts special chars
DESCRIPTION:
entry.sh line 104 does `s=$(echo "${o}" | xargs)`. xargs interprets backslash escapes, single quotes, and double quotes as quoting operators. Option values with quotes or backslashes are silently corrupted. Example: `foo=bar'baz` becomes `foo=barbaz` after xargs.

REPRODUCTION:
1. Pass options with quotes or backslashes in values.
2. Values are silently corrupted by xargs.

IMPACT:
Data integrity bug; users with special chars in option values lose data.

RELATED FILES:
entry.sh
action.yml (option input documentation)

### LINKED ISSUE #1686: entry.sh INPUT_OPTIONS can override vitals_url (phishing vector)
DESCRIPTION:
entry.sh lines 110-113 allow INPUT_OPTIONS to override VITALS_URL:
```
if [[ "${k}" == vitals_url ]]; then
  VITALS_URL="${v}"
  continue
fi
```
A user passing `INPUT_OPTIONS=vitals_url=https://evil.com/...` redirects all vitals links to a phishing site. The vitals_url is exposed in the GitHub Actions Summary table. An attacker who can set INPUT_OPTIONS can redirect users to arbitrary URLs via the "Vitals" link.

REPRODUCTION:
1. Set INPUT_OPTIONS=vitals_url=https://evil.com/fake-vitals.html
2. Run the action; the summary links to evil.com.

IMPACT:
Phishing vector via job summary Vitals link.

RELATED FILES:
entry.sh
action.yml

VasilevNStas:
## Changes

Fixes #1686, fixes #1685.

### entry.sh
- Remove the `vitals_url` override from INPUT_OPTIONS processing. Users could redirect
  the Vitals link to arbitrary URLs via `INPUT_OPTIONS=vitals_url=https://evil.com`.
  The vitals_url is now always auto-generated from the repository name.
- Replace `xargs` with `sed` trim to prevent corruption of special characters
  (quotes, backslashes) in option values.

## Background

The `vitals_url` is auto-generated as `https://{owner}.github.io/{repo}/{factbase}-vitals.html`
and is a fixed value derived from the action context. There is no legitimate use case
for overriding it via k=v options — doing so creates a phishing vector in the job summary.

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -105,16 +105,12 @@ VITALS_URL="https://${GITHUB_REPOSITORY_OWNER}.github.io/${GITHUB_REPO_NAME}/${n
 
 declare -a options=()
 while IFS= read -r o; do
-    s=$(echo "${o}" | xargs)
+    s=$(echo "${o}" | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')
     if [ "${s}" = "" ]; then
         continue
     fi
     k=$(echo "${s} "| cut -f1 -d '=')
     v=$(echo "${s}" | cut -f2- -d '=')
-    if [[ "${k}" == vitals_url ]]; then
-        VITALS_URL="${v}"
-        continue
-    fi
     options+=("--option=${k}=${v}");
 done <<< "${INPUT_OPTIONS}"
 if [ -z "${INPUT_REPOSITORIES}" ]; then


VasilevNStas:
**Note:** This PR addresses two issues (#1686, #1685) in a single change because both are in `entry.sh` and the xargs fix (#1685) removes a special-character corruption vector in option value processing, which complements the vitals_url security hardening (#1686). Same file, related scope — keeping them together reduces merge overhead.

VasilevNStas:
@yegor256 plz review

yegor256:
Not merging: I closed #1686 during triage today, since whoever sets `INPUT_OPTIONS` already controls every step of the job, so no privilege boundary is crossed. The PR is left open in case you want to argue the point on the issue.

VasilevNStas:

Both halves of this PR rely on a threat model that does not hold here. `INPUT_OPTIONS` is authored by the workflow owner, who already controls every step of the job (secrets, shell, network); there is no privilege boundary crossed by letting them set `vitals_url` or by mangling special characters in option values. It is their own input, and the "phishing"/corruption vectors described in #1686 and #1685 do not constitute a security or correctness boundary worth hardening against. Thank you for the triage.

PR #1751, opened 2026-07-03
#1663: prevent DSL injection in twice? query builder

### LINKED ISSUE #1663: DSL injection in github-events.rb `twice?` query construction
DESCRIPTION:
The `twice?` function in github-events.rb builds a factbase query by interpolating arbitrary string values into the Sexp DSL:
```
pairs.map do |prop, value|
  val = case value
        when String then "'#{value}'"
        ...
        end
  "(eq #{prop} #{val})"
end.join(' ')
```
If `value` contains a single quote (e.g. label name `O'Brien` or `test'evil`), the resulting Sexp is malformed or injection-vulnerable: `(eq label 'O'Brien')` is broken; `(eq label 'foo') (eq what 'evil` could close the predicate and add new ones.

REPRODUCTION:
1. A label or other string-typed fact with name containing `'` exists in the factbase.
2. github-events.rb calls twice?(fact).
3. DSL injection / malformed query.

IMPACT:
Query parser failures or filter bypass; potential data exposure.

RELATED FILES:
lib/cover_qo.rb (same pattern)
lib/issue_was_lost.rb (same pattern)
judges/revive-user/revive-user.rb

VasilevNStas:
## Description

The `twice?` function in `github-events.rb` builds a factbase query by interpolating string values wrapped in single quotes. If a value contains `'` (e.g. label name `O'Brien`), the resulting Sexp is malformed or injection-vulnerable.

## Fix

Escape single quotes in String values before Sexp interpolation.

Closes #1663

### DIFF
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -389,7 +389,7 @@ Fbe.iterate do
       pairs.map do |prop, value|
         val =
           case value
-          when String then "'#{value}'"
+          when String then "'#{value.gsub("'", "\\\\'")}'"
           when Time then value.utc.iso8601
           else value
           end
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -14,6 +14,7 @@ require 'tago'
 require_relative '../../lib/fill_fact'
 require_relative '../../lib/pull_request'
 require_relative '../../lib/supervision'
+require_relative '../../lib/twice'
 
 Fbe.iterate do
   as 'events_were_scanned'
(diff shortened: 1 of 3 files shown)


VasilevNStas:
@yegor256 please review

yegor256:
The escaping is right: `factbase`'s tokenizer (`lib/factbase/syntax.rb:110`) drops a backslash before a quote inside a string, so `'O\'Brien'` parses back to `O'Brien`. Not merging yet only because there is no test: please add one that stores a `label-was-attached` fact whose label carries an apostrophe and calls `twice?` on it, so the bug stays reproduced.

VasilevNStas:
@yegor256 Added the test. Since `twice?` was a singleton method inside the `Fbe.iterate` block (not reachable as `Jp.twice?`), I moved it to `lib/twice.rb` as `Jp.twice?` — same pattern as `lib/fill_fact.rb`. The new `test/lib/test_twice.rb` stores a `label-was-attached` fact with label `O'Brien` and calls `twice?` on it: without the escaping it fails with `Factbase::Syntax::Broken: String not closed`, with it the query parses and returns the correct result.

yegor256:
This PR has a merge conflict with the master branch. Could you please rebase onto the latest master and resolve the conflict?

---

VasilevNStas:
Closing this PR — it is now obsolete. Master already merged a superior fix for #1663 in #1913: `Jp.twice?` now binds values as query parameters (`$what` placeholders) instead of pasting/escaping them, so injection is impossible by construction. The merged test already covers the exact case you asked for (`test_finds_the_second_copy_when_the_label_has_an_apostrophe`, plus a backslash case). My escaping approach is superseded by that parameter-binding solution. Thank you @yegor256.

PR #1752, opened 2026-07-03
#1668: prevent DSL injection in cover_qo factbase query

### LINKED ISSUE #1668: lib/cover_qo.rb factbase query injection via judge name interpolation
DESCRIPTION:
lib/cover_qo.rb builds a query by interpolating `judge` into a Sexp DSL:
```
"(and (eq what '#{judge}') ...)"
```
If `judge` contains a quote or paren, the query is rewritten. $judge is typically safe but is passed through options.

REPRODUCTION:
1. Set judge name to `"foo') (eq where 'hacked"`.
2. Jp.cover_qo rewrites the query.

IMPACT:
Filter bypass; factbase integrity.

RELATED FILES:
lib/cover_qo.rb
judges/dimensions-of-terrain/dimensions-of-terrain.rb (same pattern)
lib/issue_was_lost.rb

VasilevNStas:
## Description

`lib/cover_qo.rb` builds a factbase query by interpolating `judge` name. If the name contains `'`, the Sexp is malformed.

## Fix

Escape single quotes in the judge name before query interpolation.

Closes #1668

### DIFF
--- a/lib/cover_qo.rb
+++ b/lib/cover_qo.rb
@@ -9,7 +9,9 @@ require_relative 'jp'
 
 def Jp.cover_qo(days, judge: $judge, loog: $loog, today: Time.parse(ENV['TODAY'] || Time.now.utc.iso8601))
   slice = days * 24 * 60 * 60
-  facts = Fbe.fb.query("(and (eq what '#{judge}') (exists since) (exists when))").each.to_a.sort_by(&:since)
+  facts = Fbe.fb.query(
+    "(and (eq what '#{judge.gsub("'", "\\\\'")}') (exists since) (exists when))"
+  ).each.to_a.sort_by(&:since)
   last = facts.map(&:when).max
   if last.nil?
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 plz review

yegor256:
Not merging: I closed #1668 during triage today, since `judge` is `$judge`, the name of the judge directory, never user input. The PR is left open in case you want to argue the point on the issue.

VasilevNStas:

`$judge` is the judge directory name, set by the orchestration layer and derived from the repo's own `judges/*` layout. It is not user input, and since judge names must be valid directory names, an apostrophe cannot occur in practice — the injection vector does not exist.

This differs from the case we fixed in #1663 (`twice?`), where the interpolated value was a fact property (e.g. a label like `O'Brien`), i.e. genuinely user-controlled data. Interpolating a controlled constant like `$judge` does not warrant escaping. Thank you for the triage.

PR #1754, opened 2026-07-03
#1703: prevent replacement string injection in update-version action

### LINKED ISSUE #1703: update-version action does not escape `$n` in replacement string
DESCRIPTION:
JavaScript's String.prototype.replace interprets $$, $&, $', $`, $n in the replacement string. If latest (a tag name returned by GitHub) contains $ characters, README gets unexpected substitution. While GitHub tag names cannot literally contain $ due to ref-format rules, attackers controlling tag names via a fork could craft tag names that, after listTags, break the file content.

REPRODUCTION:
1. Tag name 0.0.0$xxx produces literal $xxx removal in README.

IMPACT:
README corruption.

RELATED FILES:
.github/actions/update-version/action.yml

VasilevNStas:
## Description

`String.prototype.replace()` interprets `$` patterns (`$&`, ``, `$'`, etc.) in string replacement values. If a tag name ever contained these characters, the README content would be silently corrupted.

## Fix

Use a function callback instead of a string for `replace()` — function return values are used literally without any `$` processing.

Closes #1703

### DIFF
--- a/.github/actions/update-version/action.yml
+++ b/.github/actions/update-version/action.yml
@@ -29,6 +29,6 @@ runs:
           const fs = require('fs');
           let readme = fs.readFileSync('README.md', 'utf8');
           const escaped = '${{ inputs.repo }}'.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
-          readme = readme.replace(new RegExp(escaped + '@.+$', 'gm'), `${{ inputs.repo }}@${latest}`);
+          readme = readme.replace(new RegExp(escaped + '@.+$', 'gm'), () => '${{ inputs.repo }}@' + latest);
           fs.writeFileSync('README.md', readme);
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 plz review

yegor256:
Not merging: I closed #1703 during triage today, since only our own tags reach that replacement, `inputs.owner` and `inputs.repo` being hardcoded in `up.yml`. The PR is left open in case you want to argue the point on the issue.

VasilevNStas:

The value that reaches the `replace()` replacement is `${{ inputs.repo }}@${latest}`, where `inputs.owner`/`inputs.repo` are hardcoded in `up.yml` and `latest` is a tag of our own repository. Version tags cannot contain `$`-patterns, and only repo collaborators can create tags anyway — the same people who already control the repository. No injection boundary is crossed, so the function-callback change is defensive at best. Thank you for the triage.

PR #1757, opened 2026-07-03
#1661: prevent infinite loop in earliest() commit pagination

### LINKED ISSUE #1661: github-events.rb earliest() infinite loop on repos with very long commit history
DESCRIPTION:
In github-events.rb's `earliest()` function (lines 112-120), the loop does:
```ruby
while commits.size != 1
  commits = Fbe.octo.commits(repo, sha: last[:sha])
  last = commits.last
end
```
For a repo with a very long commit history (e.g. 100k+ commits), this fetches older pages of commits one at a time, potentially forever if `commits` always has size > 1 (e.g. >100 commits always returned). Each iteration consumes API quota. Also, if `commits.last` is nil (empty response edge case), `last[:sha]` raises NoMethodError.

REPRODUCTION:
1. Process a repo with very long commit history.
2. Watch the API quota burn; or
3. Process an empty repo (commits is []); nil[:sha] raises.

IMPACT:
API quota exhaustion (GitHub secondary rate limit); or crash.

RELATED FILES:

VasilevNStas:
Changes the `earliest()` method in `github-events.rb` from sha-based pagination to page-based pagination with a safety limit of 100 pages, preventing infinite loops when commit history is very large or GitHub returns unexpected results.

Also updates test stubs to match the new `page=N` parameter in commit requests.

Closes #1661

### DIFF
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -110,11 +110,14 @@ Fbe.iterate do
   end
 
   def self.earliest(repo)
-    commits = Fbe.octo.commits(repo)
-    last = commits.last
-    while commits.size != 1
-      commits = Fbe.octo.commits(repo, sha: last[:sha])
+    page = 1
+    last = nil
+    loop do
+      raise(Fbe::Error, "Too many pages for commits of ##{repo}") if page > 100
+      commits = Fbe.octo.commits(repo, per_page: 100, page: page)
+      break if commits.empty?
       last = commits.last
+      page += 1
     end
     $loog.debug("The repo ##{repo} has this last commit: #{last}")
     last
--- a/test/judges/test-github-events.rb
+++ b/test/judges/test-github-events.rb
@@ -940,7 +940,7 @@ class TestGithubEvents < Jp::Test
         'X-RateLimit-Remaining' => '999'
       }
     )
-    stub_request(:get, 'https://api.github.com/repos/zerocracy/fbe/commits?per_page=100').to_return(
+    stub_request(:get, 'https://api.github.com/repos/zerocracy/fbe/commits?page=1&per_page=100').to_return(
       body: [
         { sha: '4683257342e98cd94becc2aa49900e720bd792e9' },
         { sha: '69a28ba1122af281936371bbb36f67e5b97246b1' }
@@ -950,12 +950,8 @@ class TestGithubEvents < Jp::Test
         'X-RateLimit-Remaining' => '999'
       }
     )
-    stub_request(
-      :get,
-      'https://api.github.com/repos/zerocracy/fbe/commits?' \
-      'per_page=100&sha=69a28ba1122af281936371bbb36f67e5b97246b1'
-    ).to_return(
-      body: [{ sha: '69a28ba1122af281936371bbb36f67e5b97246b1' }].to_json,
+    stub_request(:get, 'https://api.github.com/repos/zerocracy/fbe/commits?page=2&per_page=100').to_return(
+      body: [].to_json,
       headers: {
         'Content-Type': 'application/json',
         'X-RateLimit-Remaining' => '999'
@@ -1157,7 +1153,11 @@ class TestGithubEvents < Jp::Test
         { login: 'yegor512', id: 526_302 }
       ]
     )
-    stub_github('https://api.github.com/repos/foo/foo/commits?per_page=100', body: [{ sha: '4683257342e98cd94' }])
+    stub_github(
+      'https://api.github.com/repos/foo/foo/commits?page=1&per_page=100',
+      body: [{ sha: '4683257342e98cd94' }]
+    )
+    stub_github('https://api.github.com/repos/foo/foo/commits?page=2&per_page=100', body: [])
     stub_github(
       'https://api.github.com/repos/foo/foo/compare/4683257342e98cd94...0.0.3?per_page=100',
       body: {
@@ -2582,7 +2582,8 @@ class TestGithubEvents < Jp::Test
       'https://api.github.com/repos/foo/foo/contributors?per_page=100',
       body: [{ login: 'yegor256', id: 526_301 }]
     )
-    stub_github('https://api.github.com/repos/foo/foo/commits?per_page=100', body: [{ sha: 'abc123def456' }])
+    stub_github('https://api.github.com/repos/foo/foo/commits?page=1&per_page=100', body: [{ sha: 'abc123def456' }])
+    stub_github('https://api.github.com/repos/foo/foo/commits?page=2&per_page=100', body: [])
     stub_github(
       'https://api.github.com/repos/foo/foo/compare/abc123def456...1.0.0?per_page=100',
       status: 404,
@@ -2642,7 +2643,8 @@ class TestGithubEvents < Jp::Test
       status: 403,
       body: { message: 'Resource not accessible by integration' }
     )
-    stub_github('https://api.github.com/repos/foo/foo/commits?per_page=100', body: [{ sha: 'abc123def456' }])
+    stub_github('https://api.github.com/repos/foo/foo/commits?page=1&per_page=100', body: [{ sha: 'abc123def456' }])
+    stub_github('https://api.github.com/repos/foo/foo/commits?page=2&per_page=100', body: [])
     stub_github(
       'https://api.github.com/repos/foo/foo/compare/abc123def456...1.0.0?per_page=100',
       status: 403,


VasilevNStas:
@yegor256 please review

judges/github-events/github-events.rb
```diff
@@ -110,11 +110,14 @@ def self.allcontributors(repo)
   end
 
   def self.earliest(repo)
-    commits = Fbe.octo.commits(repo)
-    last = commits.last
-    while commits.size != 1
-      commits = Fbe.octo.commits(repo, sha: last[:sha])
+    page = 1
+    last = nil
+    loop do
```
Favixx, line 115: The PR description and #1661 both attribute the hang to "very large commit history". That is not the mechanism, and it matters because the wrong diagnosis is what motivated the 100-page cap.

`Fbe.octo.commits(repo, sha: X)` returns X plus its ancestors, ordered by committer date descending. The old loop took `.last` of that list and fed it back as the next `sha:`. If X's own ancestors all carry *newer* committer dates than X — routine after `git rebase`, `git filter-branch`, or an SVN/CVS import — then `.last` is X itself, the next request is byte-identical to the previous one, and the loop is a fixed point. It spins at a constant rate forever regardless of history length. A 200-commit repo with rewritten dates hangs; torvalds/linux, with 1.3M commits and monotonic dates, terminates in two calls because `auto_paginate` collapses the first request into the full listing.

The rewrite does fix this: `page` only increases, so no request can repeat. That is the argument for the change, and it is worth putting in the commit message in place of the "long history" claim — which also removes the justification for the cap (see line 116).

judges/github-events/github-events.rb
```diff
@@ -110,11 +110,14 @@ def self.allcontributors(repo)
...
+      raise(Fbe::Error, "Too many pages for commits of ##{repo}") if page > 100
```
Favixx, line 116: Given that `page` only ever increases and the loop exits on the first empty page, termination is already guaranteed by construction. This guard adds no safety; its only observable effect is a new failure mode.

With `Fbe.octo`'s `per_page = 100` (fbe 0.49.1, `lib/fbe/octo.rb:82`), the cap means "any repository with 10,000 or more commits raises". That is a policy decision — judges-action will not compute release metrics for large repos — encoded as a bare literal in a private method, with no constant, no comment, and no relation to any documented GitHub limit. The list-commits endpoint has no page ceiling, so 100 is not derived from anything.

The two constants that define the threshold also live in different repositories: `100` here, `per_page = 100` in fbe. If fbe ever lowers its default to 30, this silently becomes a 3,000-commit ceiling with no code change here and no test that would notice. If a bound is genuinely wanted, express it in commits (`MAX_COMMITS = 100_000`), derive the page count from the `per_page` actually in effect, and state in a comment what is supposed to happen to a repo that exceeds it.

judges/github-events/github-events.rb
```diff
@@ -110,11 +110,14 @@ def self.allcontributors(repo)
...
+      commits = Fbe.octo.commits(repo, per_page: 100, page: page)
```
Favixx, line 117: `Fbe.octo` configures the Octokit client with `auto_paginate = true` (fbe 0.49.1, `lib/fbe/octo.rb:82`). Octokit's `paginate` honours the explicit `page:` for the *first* request and then keeps following `Link: rel="next"` until the collection is exhausted. So this call does not fetch one page of 100 — it fetches page N through the end of history.

What the loop therefore does: iteration 1 (`page: 1`) returns every commit in the repository, so `commits.last` is already the final answer after the first pass. Iteration 2 re-downloads everything from offset 100, iteration 3 from offset 200, and so on until the offset passes the end. For N commits that is roughly (N/100)²/2 HTTP requests instead of N/100 — a 5,000-commit repo costs ~1,275 round trips where the old code needed ~50, and every one of them is charged against the same rate limit. Issue #1661's stated impact is "API quota exhaustion (GitHub secondary rate limit)"; this change increases consumption by an order of magnitude on exactly the repos the issue is about.

The tests cannot show this because WebMock stubs return no `Link` header, so `auto_paginate` never engages there.

Two fixes, either acceptable. Given auto_paginate, the whole method collapses to `Fbe.octo.commits(repo).last` with no loop, no cap and no guard. Or, if explicit paging is wanted, run the loop inside `Fbe.octo.with_disable_auto_paginate { |octo| ... }` — fbe exposes it for precisely this case (`lib/fbe/octo.rb:201`).

judges/github-events/github-events.rb
```diff
@@ -110,11 +110,14 @@ def self.allcontributors(repo)
...
+      break if commits.empty?
```
Favixx, line 118: Even with auto-pagination disabled, breaking only on an empty page guarantees one wasted request per invocation. A repo whose commit count is not an exact multiple of 100 returns a short final page, and the loop still asks for the next one just to learn it is empty. `break if commits.size < 100` (better: compared against the `per_page` actually sent) exits on the short page and removes the extra round trip.

The diff is its own evidence here. Every one of the four test sites had to grow a companion `page=N+1 → []` stub — those stubs exist only to satisfy a request the method should not be making.

judges/github-events/github-events.rb
```diff
@@ -110,11 +110,14 @@ def self.allcontributors(repo)
...
       last = commits.last
```
Favixx, line 119: `commits.last` is the oldest commit *by committer date*, not the root commit. Those differ under exactly the condition that made the old loop spin (see line 115): rewritten or imported history where an ancestor carries a newer date than its descendant.

The value flows to `info()` line 73 as `since`, and from there into `Fbe.octo.compare(repo, since, tag)` on line 86. `compare` diffs from the *merge base* of the two refs. If `since` is not an ancestor of `tag`, the merge base is a different commit entirely and `info[:commits]` and `info[:hoc]` are computed over the wrong range — a wrong number written into the factbase as a release metric, with no exception and no log line to mark it as suspect.

This is pre-existing rather than introduced here, but this PR is rewriting the traversal, so it is the moment to decide: walk parent links to a genuine root, or document that `earliest` means "oldest by committer date" and accept that the metric can be wrong on rewritten histories.

judges/github-events/github-events.rb
```diff
@@ -110,11 +110,14 @@ def self.allcontributors(repo)
...
+      page += 1
```
Favixx, line 120: This loop can run 100 iterations issuing thousands of HTTP requests (see line 117) with no `Fbe.over?` check and no quota inspection, unlike every other bulk-iteration site in this codebase — `Fbe.iterate` and `Fbe.unmask_repos` both check before each unit of work.

fbe's client interceptor raises `Fbe::OffQuota` before *any* octo call once the rate limit is spent (`lib/fbe/octo.rb:211`). So the realistic production outcome on a large repository is not the `page > 100` guard firing — it is `Fbe::OffQuota` escaping from the middle of this loop and aborting the judge, after the quota has already been burned by this one method.

If the loop is kept, check `Fbe.over?(...)` or `Fbe.octo.off_quota?` at the top of each iteration and return the best `last` found so far, so the cycle degrades instead of dying.

judges/github-events/github-events.rb
```diff
@@ -110,11 +110,14 @@ def self.allcontributors(repo)
...
     end
     $loog.debug("The repo ##{repo} has this last commit: #{last}")
```
Favixx, line 122: Now that this method can walk up to 100 pages, the single diagnostic it emits still says nothing about how much work it did. When the `page > 100` guard fires in production there is no way to tell from the logs whether the repository genuinely has 10,000+ commits or the loop is misbehaving — which is the first question anyone will ask. Include the page count reached.

Separately, interpolating `last` dumps the entire Sawyer resource — full commit message, author and committer blocks including e-mail addresses, and the verification payload — into the debug log once per repository. `#{last&.fetch(:sha, nil)}` carries the same diagnostic value.

judges/github-events/github-events.rb
```diff
@@ -110,11 +110,14 @@ def self.allcontributors(repo)
...
     last
```
Favixx, line 123: The `Fbe::Error` raised on line 116 is not caught by this method's own rescue clauses immediately below (`Octokit::NotFound`/`Deprecated` → nil, `Octokit::Forbidden` → nil), nor by `info()` at line 73, nor by `Jp.supervision`, which logs context and re-raises. It escapes all the way out.

So a single oversized repository aborts the entire `github-events` judge for the whole cycle: every other repository in the same run loses its event scan, and the factbase gains nothing for any of them. Compared to the original hang, the blast radius is wider, not narrower — a hang at least left prior repos' work committed, whereas this takes down the run.

That also breaks the convention this file establishes two lines down, where a per-repo failure degrades to `nil` and the scan continues. Whatever bound you settle on, it should follow the same shape: log a warning and return the last commit found so far, or return `nil` — not raise.

test/judges/test-github-events.rb
```diff
@@ -940,7 +940,7 @@ def test_release_event_contributors
         'X-RateLimit-Remaining' => '999'
       }
     )
-    stub_request(:get, 'https://api.github.com/repos/zerocracy/fbe/commits?per_page=100').to_return(
+    stub_request(:get, 'https://api.github.com/repos/zerocracy/fbe/commits?page=1&per_page=100').to_return(
```
Favixx, line 943: This PR closes a hang and adds no test that fails without the fix. All four test changes are stub-URL rewrites for existing tests; revert `github-events.rb` and these fail only because the stubbed URLs no longer match the requests, not because the behaviour is wrong. That is a change-detector, not a regression test — a future refactor back to `sha:`-chaining would reintroduce the loop and CI would be green as long as the URLs lined up.

The convincing test for the old defect is small and does not need the new code to express it: stub `commits?per_page=100` returning two commits, and `commits?per_page=100&sha=<second sha>` returning a list whose **last element is that same sha** — the fixed point described on `github-events.rb:115`. Under the old implementation that test never returns; under the new one it passes immediately. Wrapping it in a `Timeout.timeout(5)` makes the failure a clean assertion rather than a hung CI job.

test/judges/test-github-events.rb
```diff
@@ -950,12 +950,8 @@ def test_release_event_contributors
         'X-RateLimit-Remaining' => '999'
       }
     )
-    stub_request(
-      :get,
-      'https://api.github.com/repos/zerocracy/fbe/commits?' \
-      'per_page=100&sha=69a28ba1122af281936371bbb36f67e5b97246b1'
-    ).to_return(
-      body: [{ sha: '69a28ba1122af281936371bbb36f67e5b97246b1' }].to_json,
+    stub_request(:get, 'https://api.github.com/repos/zerocracy/fbe/commits?page=2&per_page=100').to_return(
```
Favixx, line 953: `page=2 → []` is not how GitHub responds here. A first page of two commits carries no `Link: rel="next"` header, so there is no second request to make. This stub exists only to satisfy the implementation's extra probe — see the note about `break if commits.empty?` on `github-events.rb:118`.

The more consequential point: because WebMock stubs return no `Link` header at all, `auto_paginate` — which `Fbe.octo` enables (fbe `lib/fbe/octo.rb:74`) — never engages anywhere in this file. The production pagination path is untested by construction, which is exactly why the quadratic request pattern described on `github-events.rb:117` is invisible in a green test suite.

A stub that returns `Link: <https://api.github.com/repos/zerocracy/fbe/commits?page=2&per_page=100>; rel="next"` on page 1 would exercise the real path and would immediately show the loop re-fetching the tail of history on every iteration.

test/judges/test-github-events.rb
```diff
@@ -1157,7 +1153,11 @@ def test_release_contributors_no_tag_no_id
         { login: 'yegor512', id: 526_302 }
       ]
     )
-    stub_github('https://api.github.com/repos/foo/foo/commits?per_page=100', body: [{ sha: '4683257342e98cd94' }])
+    stub_github(
+      'https://api.github.com/repos/foo/foo/commits?page=1&per_page=100',
+      body: [{ sha: '4683257342e98cd94' }]
+    )
+    stub_github('https://api.github.com/repos/foo/foo/commits?page=2&per_page=100', body: [])
```
Favixx, line 1160: Nothing in this file exercises the `page > 100` guard, so the branch this entire PR is built around has zero coverage. As it stands the guard could be written `if page > 1` and every test here would still pass.

A direct test is cheap: register a `Regexp` stub matching `commits\?page=\d+&per_page=100` that always returns a one-element body with a `rel="next"` link, then `assert_raises(Fbe::Error) { ... }` and assert on the message. That covers both the constant and the message format, and it is the only test that would notice if fbe's `per_page` default changed underneath the 10,000-commit threshold discussed on `github-events.rb:116`.

test/judges/test-github-events.rb
```diff
@@ -2582,7 +2582,8 @@ def test_release_event_with_nil_compare_response
       'https://api.github.com/repos/foo/foo/contributors?per_page=100',
       body: [{ login: 'yegor256', id: 526_301 }]
     )
-    stub_github('https://api.github.com/repos/foo/foo/commits?per_page=100', body: [{ sha: 'abc123def456' }])
+    stub_github('https://api.github.com/repos/foo/foo/commits?page=1&per_page=100', body: [{ sha: 'abc123def456' }])
+    stub_github('https://api.github.com/repos/foo/foo/commits?page=2&per_page=100', body: [])
```
Favixx, line 2586: Issue #1661 reports two defects: the loop, and `last[:sha]` raising `NoMethodError` when the repository has no commits. The second is genuinely fixed by initialising `last = nil`, and it also has no test.

Worth adding, because the resulting behaviour is not obviously correct and is silent. With `commits?page=1&per_page=100` returning `[]`, `earliest` returns `nil`, `since` stays `nil` in `info()` (line 73), and `comparison(repo, nil, tag)` calls `Fbe.octo.compare(repo, nil, tag)` — which builds `/repos/foo/foo/compare/...1.0.0`, gets a 404, is rescued to `nil` on line 89, and `info` returns an empty hash. The release fact ends up with no `hoc` and no `commits` property at all, and the only trace is one `info`-level log line about a failed compare.

A test asserting what the factbase actually contains for an empty repository would pin that down as intended rather than accidental.

Favixx (changes requested):
The rewrite does eliminate the real defect — a monotonically increasing `page` cannot form a fixed point the way the old `sha:` feedback loop could — but three things need to change before this can merge.

1. **The new loop is quadratic in HTTP requests.** `Fbe.octo` sets `auto_paginate = true` (fbe 0.49.1, `lib/fbe/octo.rb:82`), so `commits(repo, page: N)` fetches page N *and follows every `Link: rel="next"` to the end*. Iteration 1 already returns the whole history; iterations 2..N re-download suffixes of it. A 5,000-commit repo goes from ~50 requests to ~1,275. #1661's stated impact is "API quota exhaustion" — this makes that worse, not better. Details on line 117.

2. **The 100-page guard is redundant for termination and introduces a new failure mode.** Monotonic `page` plus `break if commits.empty?` already guarantee the loop ends. The guard's only effect is that any repository with 10,000+ commits now raises `Fbe::Error`, which nothing in the call path rescues — so one oversized repo aborts the whole `github-events` judge for every repo in the cycle. Details on lines 116 and 123.

3. **There is no test that fails without the fix.** All four test changes are stub-URL rewrites; revert `github-events.rb` and they fail because the URLs no longer match, not because behaviour is wrong. The guard has zero coverage, the empty-repo path #1661 also reports has zero coverage, and because WebMock stubs carry no `Link` header, `auto_paginate` never engages in tests — which is why (1) is invisible here.

Also worth correcting in the commit message: the hang was not caused by "long commit history". See line 115 for the actual mechanism.

yegor256:
This will not do what the tests suggest. `Fbe.octo` sets `auto_paginate = true` (fbe, `lib/fbe/octo.rb:82`), so `commits(repo, page: N)` starts at page N and then follows every `next` link to the end of the history. The loop therefore walks the whole history once per page, which is quadratic in the number of pages, and any repository with more than 100 pages of commits now raises `Fbe::Error` instead of working. The tests pass because the WebMock stubs return no `Link` header. The narrow fix for #1661 is to guard the empty response before `last[:sha]`.

VasilevNStas:
@yegor256 

Closing this PR — it is now obsolete. Master already merged a better fix for #1661 in #1890: `earliest()` now uses `with_disable_auto_paginate` + `rels[:last]` to jump straight to the oldest commit in two requests. This avoids both the quadratic behavior you flagged (my page-based loop re-walked the whole history per page under `auto_paginate`) and the empty-response crash, and it has no 100-page cap. Thank you for the detailed feedback @yegor256.

PR #1758, opened 2026-07-03
#1678: advance cursor when search returns empty results

### LINKED ISSUE #1678: find-all-issues.rb cursor never advances on empty results (potential infinite loop)
DESCRIPTION:
find-all-issues.rb iterates issues via `before = issue` and `Fbe.iterate do |repository, issue| ... ` passing the next issue as the iterator cursor. The query `(gt issue $before)` advances `before`. If a deleted issue has `before` set but no current issue exists with `issue > before`, the iterator yields no further values. More critically, if `after` (the search horizon) is computed from `Fbe.octo.issue(repo, issue)[:created_at]` and that call returns nil for a deleted issue, the search anchor is broken and the search may produce zero results while the cursor still points to a deleted issue.

REPRODUCTION:
1. Run find-all-issues on a repo with deleted issues interleaved with live ones.
2. Cursor stuck on the deleted issue; search returns nothing.

IMPACT:
Potential infinite loop / hang on the judge cycle.

RELATED FILES:
judges/find-all-issues/find-all-issues.rb
lib/issue_was_lost.rb (related)

VasilevNStas:
When GitHub Search API returns an empty page of results, the cursor (`issue`) doesn't advance and the loop spins forever on the same empty page.

Added `issue += 1 if seen.empty? && issue <= first` to increment the cursor when no results are returned, ensuring forward progress.

Closes #1678

### DIFF
--- a/judges/find-all-issues/find-all-issues.rb
+++ b/judges/find-all-issues/find-all-issues.rb
@@ -109,6 +109,7 @@ require_relative '../../lib/qos_search'
           end
         end
         issue = first if issue < first
+        issue += 1 if seen.empty? && issue <= first
         m = [
           "Checked #{seen.count} #{type}s in #{repo}",
           ("(#{seen.joined(max: 8)})" unless seen.empty?),


VasilevNStas:
@yegor256 please review

yegor256:
Not merging: I closed #1678 during triage today, since the cursor belongs to `Fbe.iterate` and no mechanism for the described hang was shown. The PR is left open in case you want to argue the point on the issue.

VasilevNStas:
@@yegor256 Agreed with your triage — closing without arguing. I looked at `Fbe.iterate` and the hang mechanism is not there:

- Within a run the loop is bounded by `@repeats` (default 1), so the `over` block runs at most once per repository per cycle — there is no way to spin forever on one empty page inside a single run.
- Across runs a "stuck" cursor simply retries the same range on the next cycle, which self-resolves once the GitHub search index (eventually consistent) catches up. That is a retry, not a hang.

Moreover, the proposed `issue += 1 if seen.empty?` is risky: an empty page caused by search index lag would advance the cursor past issues that actually exist, permanently skipping them — a data-loss regression rather than a fix. Thank you for the triage.

Issue #1672, opened 2026-06-30
lib/qos_search.rb state pollution across judges in same process


sdghsdkjlas27-dotcom:
DESCRIPTION:
lib/qos_search.rb stores `@offquota`, `@scount`, `@swstart` as instance variables on the `Jp` module singleton. Jp.qoreset exists but is NOT called by all judges. In find-all-issues.rb (which calls Jp.qosearch), `qoreset` is NOT called. If quality-of-service has latched @offquota = true in the same Ruby process, find-all-issues inherits the latch and cannot search even after quota recovers.

REPRODUCTION:
1. Run quality-of-service to exhaustion.
2. Run find-all-issues in same process (e.g. via parallel runner).
3. find-all-issues sees @offquota and skips all searches.

IMPACT:
Cross-judge state pollution.

RELATED FILES:
lib/qos_search.rb
judges/find-all-issues/find-all-issues.rb (does not call qoreset)
judges/quality-of-service/quality-of-service.rb (calls qoreset correctly)

PR #1771, opened 2026-07-03
#1664: add reactions cache to reduce N+1 API calls

### LINKED ISSUE #1664: N+1 reactions API calls in lib/pull_request.rb count_appreciated_comments
DESCRIPTION:
lib/pull_request.rb's `count_appreciated_comments` (lines 65-91) calls `Fbe.octo.issue_comment_reactions` and `Fbe.octo.pull_request_review_comment_reactions` once PER COMMENT. A PR with 200 comments triggers 200 individual reaction API calls. Combined with 5000 req/hr primary quota, a single busy PR exhausts the rate limit.

REPRODUCTION:
1. Profile a fix-missing-comments or pull-was-merged run on a PR with 100+ comments.
2. Observe one API call per comment.

IMPACT:
Single PR with many comments exhausts GitHub rate limit, causing all subsequent API calls (across all judges) to fail.

RELATED FILES:
lib/pull_request.rb
judges/pull-was-merged/pull-was-merged.rb (consumer)
judges/fix-missing-comments/fix-missing-comments.rb

VasilevNStas:
Adds an instance variable cache (`@reactions`) in `Jp.count_appreciated_comments` to avoid N+1 API calls when the same comment reactions are looked up multiple times for the same pull request.

The cache is keyed by comment ID and memoizes the result of `issue_comment_reactions` and `pull_request_review_comment_reactions`.

Closes #1664

### DIFF
--- a/lib/pull_request.rb
+++ b/lib/pull_request.rb
@@ -64,9 +64,11 @@ end
 
 def Jp.count_appreciated_comments(pr, issue_comments, code_comments, repo: nil)
   repo = pr.dig(:base, :repo, :full_name) if repo.nil?
+  @_reactions ||= {}
   issue_comments.sum do |comment|
-    Fbe.octo.issue_comment_reactions(repo, comment[:id])
-      .count { |reaction| reaction.dig(:user, :id) != comment.dig(:user, :id) }
+    @_reactions[comment[:id]] ||=
+      Fbe.octo.issue_comment_reactions(repo, comment[:id])
+        .count { |reaction| reaction.dig(:user, :id) != comment.dig(:user, :id) }
   rescue Octokit::NotFound, Octokit::Deprecated => e
     $loog.info("Issue comment ##{comment[:id]} reactions don't exist in #{repo}: #{e.message}")
     0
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 plz review this PR

yegor256:
This does not remove the N+1: the first pass over a pull request's comments still makes one call per comment, which is what #1664 is about. What it adds is a `@reactions` hash on the `Jp` module that lives for the whole run and is keyed by comment id across two different id spaces, the kind of cross-judge state we just removed in #1672. Repeated lookups are already served by the HTTP cache, and GitHub does not charge quota for a 304, so the saving here is small. A batched GraphQL query would address the issue.

VasilevNStas:
@yegor256 

Closing this PR. You are right @yegor256 — the cache does not remove the N+1 (the first pass still makes one call per comment), and the `@reactions` hash on the `Jp` module is exactly the kind of cross-judge state we removed in #1672, keyed by comment id across two different id spaces where collisions could return wrong counts. Repeated lookups are already served by the HTTP cache and a 304 costs no quota. The proper fix is a batched GraphQL query, which I will leave for a separate effort.

PR #1781, opened 2026-07-04
#1672: fix state pollution across judges in qosearch

### LINKED ISSUE #1672: lib/qos_search.rb state pollution across judges in same process

VasilevNStas:
Closes #1672

Fix state pollution across different judges that share the same `Jp.qosearch` module. Instance variables `@offquota`, `@scount`, and `@swstart` were global to the module, so one judge's quota exhaustion would incorrectly block other judges from searching.

Changes:
- Convert `@offquota`, `@scount`, `@swstart` from scalars to hashes keyed by `$judge`
- Add migration code (`unless @offquota.is_a?(Hash)`) to handle in-flight upgrades
- Update `Jp.qoreset` to initialize hashes instead of scalars

### DIFF
--- a/lib/qos_search.rb
+++ b/lib/qos_search.rb
@@ -10,26 +10,31 @@ Jp::SEARCH_WINDOW_SECONDS = 60
 Jp::SEARCH_WINDOW_BUDGET = 25
 
 def Jp.qoreset
-  @offquota = false
-  @offquotatime = nil
-  @scount = 0
-  @swstart = nil
+  @offquota = {}
+  @offquotatime = {}
+  @scount = {}
+  @swstart = {}
 end
 
 def Jp.qosearch(query, method: :search_issues, **)
-  if @offquota
-    return if @offquotatime && (Time.now - @offquotatime) < Jp::SEARCH_WINDOW_SECONDS
-    @offquota = false
-    @offquotatime = nil
+  jg = $judge
+  @offquota = {} unless @offquota.is_a?(Hash)
+  @offquotatime = {} unless @offquotatime.is_a?(Hash)
+  @scount = {} unless @scount.is_a?(Hash)
+  @swstart = {} unless @swstart.is_a?(Hash)
+  if @offquota[jg]
+    return if @offquotatime[jg] && (Time.now - @offquotatime[jg]) < Jp::SEARCH_WINDOW_SECONDS
+    @offquota[jg] = false
+    @offquotatime[jg] = nil
   end
   return if Fbe.octo.off_quota?
   now = Time.now
-  if @swstart.nil? || (now - @swstart) >= Jp::SEARCH_WINDOW_SECONDS
-    @swstart = now
-    @scount = 0
+  if @swstart[jg].nil? || (now - @swstart[jg]) >= Jp::SEARCH_WINDOW_SECONDS
+    @swstart[jg] = now
+    @scount[jg] = 0
   end
-  if @scount >= Jp::SEARCH_WINDOW_BUDGET
-    $loog.info("[#{$judge}] Search API per-minute budget exceeded (#{Jp::SEARCH_WINDOW_BUDGET}/#{Jp::SEARCH_WINDOW_SECONDS}s)")
+  if @scount[jg] >= Jp::SEARCH_WINDOW_BUDGET
+    $loog.info("[#{jg}] Search API per-minute budget exceeded (#{Jp::SEARCH_WINDOW_BUDGET}/#{Jp::SEARCH_WINDOW_SECONDS}s)")
     return
   end
   octo = Fbe.octo
@@ -43,29 +48,29 @@ def Jp.qosearch(query, method: :search_issues, **)
     left = octo.rate_limit.remaining
   end
   if left.nil?
-    @offquota = true
-    @offquotatime = Time.now
-    $loog.warn("[#{$judge}] GitHub Search API quota info unavailable, stopping search calls")
+    @offquota[jg] = true
+    @offquotatime[jg] = Time.now
+    $loog.warn("[#{jg}] GitHub Search API quota info unavailable, stopping search calls")
     return
   end
   if left.zero?
-    @offquota = true
-    @offquotatime = Time.now
+    @offquota[jg] = true
+    @offquotatime[jg] = Time.now
     $loog.info('Too much GitHub Search API quota consumed already (0 left)')
     return
   end
-  @scount += 1
+  @scount[jg] += 1
   raise(RuntimeError, "Unsafe search method: #{method}") unless
     %i[search_issues search_code search_commits].include?(method)
   Fbe.octo.__send__(method, query, **)
 rescue Octokit::TooManyRequests => e
-  @offquota = true
-  @offquotatime = Time.now
-  $loog.warn("[#{$judge}] GitHub Search API quota exhausted, stopping search calls: #{e.message}")
+  @offquota[jg] = true
+  @offquotatime[jg] = Time.now
+  $loog.warn("[#{jg}] GitHub Search API quota exhausted, stopping search calls: #{e.message}")
   nil
 rescue Octokit::ServerError,
-  Net::OpenTimeout, Net::ReadTimeout, SocketError,
-  Errno::ECONNRESET, Errno::ETIMEDOUT => e
-  $loog.warn("[#{$judge}] Transient error in search API call (will retry next cycle): #{e.class}: #{e.message}")
+       Net::OpenTimeout, Net::ReadTimeout, SocketError,
+       Errno::ECONNRESET, Errno::ETIMEDOUT => e
+  $loog.warn("[#{jg}] Transient error in search API call (will retry next cycle): #{e.class}: #{e.message}")
   nil
 end
--- a/lib/qos_search.rb
+++ b/lib/qos_search.rb
@@ -69,8 +69,8 @@ rescue Octokit::TooManyRequests => e
   $loog.warn("[#{jg}] GitHub Search API quota exhausted, stopping search calls: #{e.message}")
   nil
 rescue Octokit::ServerError,
-       Net::OpenTimeout, Net::ReadTimeout, SocketError,
-       Errno::ECONNRESET, Errno::ETIMEDOUT => e
+  Net::OpenTimeout, Net::ReadTimeout, SocketError,
+  Errno::ECONNRESET, Errno::ETIMEDOUT => e
   $loog.warn("[#{jg}] Transient error in search API call (will retry next cycle): #{e.class}: #{e.message}")
   nil
 end
--- a/test/lib/test_qos_search.rb
+++ b/test/lib/test_qos_search.rb
@@ -121,7 +121,7 @@ class TestQosSearch < Jp::Test
       refute_nil(Jp.qosearch('repo:foo/foo type:issue'))
     end
     assert_nil(Jp.qosearch('repo:foo/foo type:issue'))
-    Jp.instance_variable_set(:@swstart, Time.now - Jp::SEARCH_WINDOW_SECONDS - 1)
+    Jp.instance_variable_set(:@swstart, { $judge => Time.now - Jp::SEARCH_WINDOW_SECONDS - 1 })
     $global[:octo] = nil
     rate_limit_up
     searchstub('repo:foo/foo type:issue', body: { total_count: 1, items: [{ number: 2 }] })
@@ -132,8 +132,8 @@ class TestQosSearch < Jp::Test
     rate_limit_up
     searchstub('repo:foo/foo type:issue', body: { message: 'API rate limit exceeded' }, status: 403)
     assert_nil(Jp.qosearch('repo:foo/foo type:issue'))
-    assert_equal(1, Jp.instance_variable_get(:@scount))
-    assert(Jp.instance_variable_get(:@offquota))
+    assert_equal(1, Jp.instance_variable_get(:@scount)[$judge])
+    assert(Jp.instance_variable_get(:@offquota)[$judge])
     Jp.qoreset
     $global[:octo] = nil
     rate_limit_up


VasilevNStas:
@yegor256 please review this PR

yegor256:

VasilevNStas:
@yegor256 I`m fixed it

lib/qos_search.rb
```diff
@@ -10,26 +10,31 @@
...
-  if @swstart.nil? || (now - @swstart) >= Jp::SEARCH_WINDOW_SECONDS
-    @swstart = now
-    @scount = 0
+  if @swstart[jg].nil? || (now - @swstart[jg]) >= Jp::SEARCH_WINDOW_SECONDS
+    @swstart[jg] = now
+    @scount[jg] = 0
   end
-  if @scount >= Jp::SEARCH_WINDOW_BUDGET
-    $loog.info("[#{$judge}] Search API per-minute budget exceeded (#{Jp::SEARCH_WINDOW_BUDGET}/#{Jp::SEARCH_WINDOW_SECONDS}s)")
+  if @scount[jg] >= Jp::SEARCH_WINDOW_BUDGET
```
Favixx, line 36: `@scount` is not judge state — it is a client-side throttle for a token-wide resource. GitHub's Search API budget is 30 requests per minute per authenticated token, and this repository already encodes that number: `test/test__helper.rb:42` stubs `search: { remaining: 30, limit: 30 }`. `SEARCH_WINDOW_BUDGET = 25` (line 10) was sized to sit just under it. Keying the counter by `$judge` hands every judge its own 25-request allowance against a budget that is not divisible.

The window is wall-clock, not per-judge: `@swstart[jg]` holds a `Time`. Two judges whose search activity falls inside the same 60 seconds — `find-all-issues` finishing at t=20s and `quality-of-service` starting at t=25s, exactly the interleaving this counter existed to prevent — now spend 25 + 25 = 50 requests in one minute against a ceiling of 30. There are three `qosearch` consumers (`judges/find-all-issues/find-all-issues.rb:56`, `judges/dimensions-of-terrain/total_active_contributors.rb:16`, and the `some_*` helpers under `judges/quality-of-service/`), so the worst case is 75.

The outcome is the exact failure this file exists to avoid: `Octokit::TooManyRequests` at line 66, which latches `@offquota` and suppresses searches anyway. The PR trades a latch that was too sticky for a throttle that no longer throttles. `@scount` and `@swstart` should stay global — whatever needs per-judge scoping here, it is not the rate counter.

lib/qos_search.rb
```diff
@@ -10,26 +10,31 @@
 Jp::SEARCH_WINDOW_BUDGET = 25
 
 def Jp.qoreset
-  @offquota = false
-  @offquotatime = nil
-  @scount = 0
-  @swstart = nil
+  @offquota = {}
+  @offquotatime = {}
+  @scount = {}
```
Favixx, line 15: `qoreset` still replaces the entire hash, so the per-judge scoping is undone once per cycle by its only caller. `judges/quality-of-service/quality-of-service.rb:16` calls `Jp.qoreset` at judge start; after this change that clears `@scount` and `@swstart` for *every* judge, including counters `find-all-issues` or `dimensions-of-terrain` accumulated seconds earlier in the same 60-second window.

Concretely: `dimensions-of-terrain` burns 25 searches between t=0 and t=20; `quality-of-service` starts at t=25 and calls `qoreset`; `@swstart` becomes `{}`; the next `qosearch` opens a fresh window at line 32 and the process issues another 25 searches before t=60. Same wall-clock minute, 50 requests. That is one judge reaching into another judge's state — the inverse of #1672, and it survives this fix untouched.

If the state is keyed by judge, the reset must be keyed too: `%i[@offquota @offquotatime @scount @swstart].each { |v| instance_variable_get(v).delete($judge) }`, or one line with the single-hash form suggested on line 62. Better, question whether `qoreset` should exist at all — #1672 lists 'find-all-issues does not call qoreset' as part of the defect, and once state is genuinely scoped no judge should be manually resetting anything.

lib/qos_search.rb
```diff
@@ -10,26 +10,31 @@
...
-  if @offquota
-    return if @offquotatime && (Time.now - @offquotatime) < Jp::SEARCH_WINDOW_SECONDS
-    @offquota = false
-    @offquotatime = nil
+  jg = $judge
+  @offquota = {} unless @offquota.is_a?(Hash)
+  @offquotatime = {} unless @offquotatime.is_a?(Hash)
+  @scount = {} unless @scount.is_a?(Hash)
+  @swstart = {} unless @swstart.is_a?(Hash)
+  if @offquota[jg]
```
Favixx, line 25: `@offquota` records a fact about the *token*, not about the judge. It is set at line 51 when `/rate_limit` gives no `resources.search.remaining`, at line 57 when that value is 0, and at line 67 on `Octokit::TooManyRequests`. All three are properties of the single GitHub installation token that every judge in the process shares. If search remaining is 0 for `quality-of-service`, it is 0 for `find-all-issues` one second later — sharing the latch was correct behaviour, and skipping the doomed call is the entire point of having one.

fbe already treats this resource as process-global: `Fbe.octo` intercepts every method in `Fbe::SEARCH_METHODS` and raises `Fbe::OffQuota` when `off_quota?(resource: :search)` is true, driven by one rate-limit middleware. Adding a judge-scoped second opinion on top of a global one means the two views can now disagree, and the judge-scoped one is the less informed of the two.

What #1672 reports as harm is not the sharing but the indistinguishability: a suppressed search and an empty search both return nil, so `judges/dimensions-of-terrain/total_active_contributors.rb:30` does `next if commits.nil?`, drops that repository, and line 45 still emits `{ total_active_contributors: seen.count }` — a false number in the factbase. Keying the latch by judge hides that symptom for one ordering and leaves the false fact reachable for every other, including the case where a single judge exhausts the quota against itself.

lib/qos_search.rb
```diff
@@ -10,26 +10,31 @@
...
+  @scount = {}
+  @swstart = {}
 end
 
 def Jp.qosearch(query, method: :search_issues, **)
-  if @offquota
-    return if @offquotatime && (Time.now - @offquotatime) < Jp::SEARCH_WINDOW_SECONDS
-    @offquota = false
-    @offquotatime = nil
+  jg = $judge
```
Favixx, line 20: `jg` is whatever `$judge` happens to hold, unvalidated. If it is nil — `qosearch` reached from a `lib/` helper, from a test that forgets the global, or from any future path outside the judges runner — every caller collapses into the single `nil` bucket and the behaviour is byte-for-byte the pre-PR behaviour. The bug this PR closes would come back silently: no log line, no test failure, no crash.

The correctness of the whole change now rests on an ambient global that this file never asserts anything about. `test/test__helper.rb:51-52` shows how loose that contract is — `load_it` calls `Jp.qoreset` *before* assigning `$judge`, so at reset time the global still holds the previous judge's name. That happens to be harmless only because `qoreset` is global (see the note on line 15); the moment the reset becomes per-judge, the helper resets the wrong bucket.

At minimum fail loudly: `raise(RuntimeError, 'qosearch called outside a judge, $judge is nil') if jg.nil?`. Passing the scope in explicitly — `Jp.qosearch(query, scope: $judge)` from the call sites — removes the ambient dependency altogether and makes the isolation testable without `instance_variable_get`.

lib/qos_search.rb
```diff
@@ -10,26 +10,31 @@
...
+  @offquota = {} unless @offquota.is_a?(Hash)
```
Favixx, line 21: The PR body calls these four lines 'migration code to handle in-flight upgrades', and that rationale does not hold. `lib/qos_search.rb` is required once per Ruby process from an image that pins one revision; there is no execution in which `@offquota` was written by the previous version of this file and read by this one. Nothing migrates.

What they actually do is lazy nil-initialisation on the hot path, which *is* needed, because `qoreset` is called only by `judges/quality-of-service/quality-of-service.rb:16` while `find-all-issues` and `dimensions-of-terrain` never call it — so on a cycle where those judges run first, `@offquota` is nil and `@offquota[jg]` at line 25 would raise. But this is the wrong mechanism for it: four `is_a?` checks on every search call, and semantics that, in the case they claim to handle, would silently discard a live quota-exhaustion latch and resume hammering the API at full rate.

Initialise once at load time instead — call `Jp.qoreset` at the bottom of this file, immediately after the two `def`s — and delete lines 21-24. The ivars are then hashes by construction, the invariant is stated in one place, and the hot path loses four branches. The `Hash.new { ... }` form suggested on line 62 achieves the same with a single ivar.

lib/qos_search.rb
```diff
@@ -43,29 +48,29 @@ def Jp.qosearch(query, method: :search_issues, **)
...
   if left.zero?
-    @offquota = true
-    @offquotatime = Time.now
+    @offquota[jg] = true
+    @offquotatime[jg] = Time.now
     $loog.info('Too much GitHub Search API quota consumed already (0 left)')
     return
   end
-  @scount += 1
+  @scount[jg] += 1
```
Favixx, line 62: Four hashes keyed by the same key have to stay in lock-step, and nothing enforces it. `@scount[jg]` is only ever initialised at line 34, inside `if @swstart[jg].nil? || elapsed`. Any state where `@swstart[jg]` holds a recent `Time` while `@scount[jg]` is missing makes line 36 evaluate `nil >= 25` and this line evaluate `nil + 1` — `NoMethodError` on both, inside the search path of a production judge.

That state is one careless write away. Lines 21-24 reset the four hashes *independently*, and `test/lib/test_qos_search.rb:124` already replaces `@swstart` wholesale without touching `@scount`. It is safe there only by accident: the planted time is older than `SEARCH_WINDOW_SECONDS`, so line 32 takes the reset branch and repairs the pairing. Plant a recent time instead and the next call raises.

Collapsing the four into one per-judge record removes the class of defect rather than this instance: `@quota = Hash.new { |h, k| h[k] = { count: 0, start: nil, off: false, off_at: nil } }`. One lazy default replaces all four guards at lines 21-24, the fields cannot desynchronise because they are created together, and `qoreset` becomes `@quota.delete($judge)` — which is also what the note on line 15 asks for.

lib/qos_search.rb
```diff
@@ -10,26 +10,31 @@
...
-    @scount = 0
+  if @swstart[jg].nil? || (now - @swstart[jg]) >= Jp::SEARCH_WINDOW_SECONDS
+    @swstart[jg] = now
+    @scount[jg] = 0
   end
-  if @scount >= Jp::SEARCH_WINDOW_BUDGET
-    $loog.info("[#{$judge}] Search API per-minute budget exceeded (#{Jp::SEARCH_WINDOW_BUDGET}/#{Jp::SEARCH_WINDOW_SECONDS}s)")
+  if @scount[jg] >= Jp::SEARCH_WINDOW_BUDGET
+    $loog.info("[#{jg}] Search API per-minute budget exceeded (#{Jp::SEARCH_WINDOW_BUDGET}/#{Jp::SEARCH_WINDOW_SECONDS}s)")
     return
```
Favixx, line 38: This `return` yields nil, and so do lines 26, 30, 54, 60, 70 and 75. Callers get one indistinguishable value for seven different outcomes: throttled by the local budget, latched off-quota, core quota low, rate limit unreadable, search quota zero, secondary rate limit, transient network error. None of them means 'GitHub searched and found nothing' — that case returns a hash with `total_count: 0`.

This ambiguity is what turns the #1672 latch into corrupt data rather than a missing fact. `judges/dimensions-of-terrain/total_active_contributors.rb:30` does `next if commits.nil?`, silently drops that repository from the `Fbe.unmask_repos` loop, and line 45 still returns `{ total_active_contributors: seen.count }` — a contributor count over an arbitrary subset of repositories, written into the factbase as the real figure. The `some_*` helpers handle it correctly (`some_merged_pulls.rb:17`, `some_backlog_size.rb:21` return `{}` and record nothing), which is why the pollution surfaces as a wrong metric rather than an absent one.

For a metrics system a false number is worse than a crash, because nothing downstream can distinguish it from a true one. Give suppression a signal callers cannot ignore — a sentinel object, or `raise(Jp::Suppressed)` — and `total_active_contributors` is forced to bail out with `{}`. That fix holds under every judge ordering, including the shuffled ordering `Jp.incremate` produces at `lib/incremate.rb:20`. Keying the latch by `$judge` only holds under the orderings where no single judge exhausts the quota against itself.

lib/qos_search.rb
```diff
@@ -43,29 +48,29 @@ def Jp.qosearch(query, method: :search_issues, **)
...
-  @scount += 1
+  @scount[jg] += 1
   raise(RuntimeError, "Unsafe search method: #{method}") unless
     %i[search_issues search_code search_commits].include?(method)
   Fbe.octo.__send__(method, query, **)
 rescue Octokit::TooManyRequests => e
-  @offquota = true
-  @offquotatime = Time.now
-  $loog.warn("[#{$judge}] GitHub Search API quota exhausted, stopping search calls: #{e.message}")
+  @offquota[jg] = true
```
Favixx, line 67: `Octokit::TooManyRequests` on a search endpoint is GitHub's secondary rate limit, enforced per token and per source IP — not per caller. Keying the latch by `jg` records 'GitHub is currently refusing our search requests' for one judge and withholds it from the others, so every judge in the cycle now has to earn its own 403 before it stops asking.

That is not a free rediscovery. Continuing to send requests after a secondary-limit response is precisely what GitHub's abuse detection escalates on, and the documented remedy is to back off process-wide until `Retry-After` elapses. The pre-PR global latch did that with one rejected call; three independently-latching judges send at least three.

The cooldown is also the wrong duration. Line 26 measures the wait against `Jp::SEARCH_WINDOW_SECONDS`, which is the local throttle window and has nothing to do with how long GitHub wants you to wait. `e.response_headers['retry-after']` is available on this exception and is the value that should drive `@offquotatime`.

lib/qos_search.rb
```diff
@@ -43,29 +48,29 @@ def Jp.qosearch(query, method: :search_issues, **)
     left = octo.rate_limit.remaining
```
Favixx, line 48: On this fallback path `left` stops being the search bucket. `Octokit::Client#rate_limit` reads the `rate` object, which is the core limit — typically 5000 — whereas line 45 reads `resources.search.remaining`, capped at 30. Everything downstream then evaluates the core budget while claiming to protect the search budget: `left.nil?` at line 50 never fires and `left.zero?` at line 56 is unreachable in practice, so `@offquota[jg]` is never set from here.

fbe hit the same trap and at least left a marker: its `off_quota?` falls back to `rate.remaining` and logs 'Search-quota check fell back to core remaining (...); search count unavailable in rate-limit middleware'. There is no equivalent warning here — the substitution of one bucket for another is completely silent, and the only way to notice is that searches never get throttled.

This matters for the change at hand because the value being copied into per-judge slots on this path is measuring the wrong resource; scoping it per judge produces three wrong answers instead of one. Either drop the fallback and let `NoMethodError` fall through to the line 50 branch, which already handles 'quota unknown' correctly, or read the search bucket explicitly rather than whatever `rate_limit` returns.

lib/qos_search.rb
```diff
@@ -43,29 +48,29 @@ def Jp.qosearch(query, method: :search_issues, **)
...
+    @offquotatime[jg] = Time.now
+    $loog.warn("[#{jg}] GitHub Search API quota info unavailable, stopping search calls")
     return
   end
   if left.zero?
-    @offquota = true
-    @offquotatime = Time.now
+    @offquota[jg] = true
+    @offquotatime[jg] = Time.now
     $loog.info('Too much GitHub Search API quota consumed already (0 left)')
```
Favixx, line 59: Every other message in this method now carries the judge prefix — lines 37, 53, 69, 74 — and this one does not, even though it sits on the most frequently taken latching path. With the state keyed by judge, 'which judge latched' is the single piece of information that makes this log actionable, and this is exactly where it is missing.

It is also the one string in the file that collides with somebody else's. fbe's `off_quota?` logs `Too much #{label} quota consumed already (#{left} < #{threshold})`, and `Fbe.octo.off_quota?` is called from line 30 of this same method, so a single `qosearch` call can emit both variants. Grepping a cycle's log for 'Too much GitHub Search API quota consumed already' returns two unrelated mechanisms with no way to tell them apart.

`$loog.info("[#{jg}] Too much GitHub Search API quota consumed already (0 left)")` costs nothing and makes the new per-judge state observable in production — which carries more weight than usual for a change whose correctness is currently not demonstrated by any test.

test/lib/test_qos_search.rb
```diff
@@ -132,8 +132,8 @@ def test_stops_after_too_many_requests_respects_window
     rate_limit_up
     searchstub('repo:foo/foo type:issue', body: { message: 'API rate limit exceeded' }, status: 403)
     assert_nil(Jp.qosearch('repo:foo/foo type:issue'))
-    assert_equal(1, Jp.instance_variable_get(:@scount))
-    assert(Jp.instance_variable_get(:@offquota))
+    assert_equal(1, Jp.instance_variable_get(:@scount)[$judge])
```
Favixx, line 135: None of the three test edits in this PR would fail against the old code. They are representation adaptations — `Jp.instance_variable_get(:@scount)` becoming `...[$judge]` — not new assertions. `setup` pins `$judge = 'test-qos-search'` at line 19 and nothing in this file ever changes it, so cross-judge isolation, the entire subject of the change, has no coverage at all. A state-pollution fix with no reproducing test is an unverified fix.

The missing test is short and fails loudly on master:

```ruby
def test_latch_does_not_leak_into_another_judge
  rate_limit_up
  searchstub('repo:foo/foo type:issue', body: { message: 'API rate limit exceeded' }, remaining: 0, status: 403)
  assert_nil(Jp.qosearch('repo:foo/foo type:issue'))
  $judge = 'other-judge'
  $global[:octo] = nil
  rate_limit_up
  searchstub('repo:foo/foo type:pr', body: { total_count: 1, items: [{ number: 7 }] })
  assert_equal(7, Jp.qosearch('repo:foo/foo type:pr')[:items].first[:number])
end
```

Note that `test/test__helper.rb:51-52` calls `Jp.qoreset` before assigning `$judge`, so every judge-level test in the suite starts from a wiped global state — which is why the pollution in #1672 was never caught by the existing tests and why it will not be caught again if the keying regresses. A budget-counter twin (25 calls under judge A, then one under judge B) is worth adding as well; it is the assertion that will start failing the moment someone acts on the note at `lib/qos_search.rb:36` and makes the counter global again, which is a conversation the test suite should force rather than avoid.

test/lib/test_qos_search.rb
```diff
@@ -121,7 +121,7 @@ def test_resets_budget_after_window_elapses
       refute_nil(Jp.qosearch('repo:foo/foo type:issue'))
     end
     assert_nil(Jp.qosearch('repo:foo/foo type:issue'))
-    Jp.instance_variable_set(:@swstart, Time.now - Jp::SEARCH_WINDOW_SECONDS - 1)
+    Jp.instance_variable_set(:@swstart, { $judge => Time.now - Jp::SEARCH_WINDOW_SECONDS - 1 })
```
Favixx, line 124: This line is the reason the change needed test edits at all: the test writes `Jp`'s private state directly, so the internal representation is now duplicated in three places in this file — here, and at lines 135 and 136. Any future move breaks three tests that are not testing that: the single per-judge record suggested on `lib/qos_search.rb:62`, reverting `@scount` to a global, or renaming the key.

There is a live coupling in the literal too. `{ $judge => ... }` replaces `@swstart` wholesale, discarding every other judge's window start, while leaving `@scount` untouched. It passes only because the planted time is older than `SEARCH_WINDOW_SECONDS`, so line 32 of `lib/qos_search.rb` takes the reset branch and re-pairs the two hashes. Change the offset to something inside the window and this test stops exercising the window reset and starts exercising `nil + 1` at line 62.

A one-line seam removes both problems: a `Jp.qoage(seconds)` helper that shifts the current judge's window start through the code that owns the invariant. The test then expresses 'pretend a minute has passed' instead of restating the shape of the module's internals, and the assertions at lines 135-136 can check observable behaviour — that the next call is or is not suppressed — rather than hash contents.

Favixx (changes requested):
The bug in #1672 is real and correctly located: `@offquota`, `@offquotatime`, `@scount` and `@swstart` live on the `Jp` module singleton, `Jp.qoreset` is called by exactly one judge (`judges/quality-of-service/quality-of-service.rb:16`), and the two other consumers — `judges/find-all-issues/find-all-issues.rb:56` and `judges/dimensions-of-terrain/total_active_contributors.rb:16` — inherit whatever the previous judge left behind. The concrete damage is visible at `total_active_contributors.rb:30`: a suppressed search returns nil, `next if commits.nil?` drops that repository, and line 45 still writes `{ total_active_contributors: seen.count }` — a count over a subset, stored in the factbase as if it were the real number.

My objection is to the axis of the fix. All four variables describe the state of a single GitHub installation token, not the state of a judge. `SEARCH_WINDOW_BUDGET = 25` is a client-side throttle sized just under GitHub's token-wide 30 requests/minute search limit (`test/test__helper.rb:42` stubs exactly `search: { remaining: 30, limit: 30 }`); `@offquota` is set from `resources.search.remaining` and from `Octokit::TooManyRequests`, both token-scoped. Keying them by `$judge` gives each judge its own 25-request allowance against an indivisible 30, and makes each judge rediscover quota exhaustion by earning its own 403. The sharing was not the defect — the defect is that `qosearch` returns a bare nil for seven different outcomes, so callers cannot tell suppression from an empty result set and write a wrong number instead of nothing.

Secondary problems, each detailed inline: `qoreset` still nukes the whole hash, so the one judge that calls it wipes every other judge's window counter — the same cross-judge reach the PR set out to remove; the `unless ... is_a?(Hash)` block is described as upgrade migration but no in-flight upgrade is possible in a single-process, pinned-image run, and what it actually provides (nil-init) belongs at load time; the four parallel hashes can desynchronise into `nil >= 25` / `nil + 1`; and `left = octo.rate_limit.remaining` (line 48) reads the core bucket while the surrounding logic treats it as the search bucket.

Most importantly, nothing here would fail on the old code. All three test edits are representation adaptations; `setup` pins `$judge = 'test-qos-search'` and no assertion in the file ever involves a second judge. A pollution fix with no reproducing test is an unverified fix. A concrete test that fails before and passes after is in the comment on `test/lib/test_qos_search.rb:135`.

0crat:
@vasilevnstas Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+44](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1785, opened 2026-07-04
#1689: fix set -e masking test failures in Makefile

VasilevNStas:
Closes #1689

Fix `set -e` masking test failures in `Makefile`. The `test` and `entry` targets used `echo "$$?" > target/*.exit` after the command, but `set -e` causes the shell to exit before `echo` runs if the command fails (nonzero exit). Now uses `|| rc=$$?` to capture the exit code without triggering the errexit trap.

Changes:
- `test` target: save docker exit code via `|| rc=$$?` pattern
- `entry` target: same pattern for action test script

Issue #1689, opened 2026-06-30
Makefile `verify` target - set -e + echo $? makes test failures pass


sdghsdkjlas27-dotcom:
DESCRIPTION:
Makefile's `test` and `entry` targets capture exit code by writing `$$?` to a file. With `.SHELLFLAGS := -e -o pipefail -c`, `set -e` causes the recipe to ABORT immediately on a failing command - `echo "$$?"` never runs. The verify target reads back the file and asserts it's 0, but a failing test will leave the previous file (or no file) value. Test failures are silently masked.

REPRODUCTION:
1. Add `exit 1` to test command in a docker container.
2. Verify passes anyway because set -e halts before echo.

IMPACT:
Test failures silently masked. CI green on broken code. CRITICAL bug.

RELATED FILES:
Makefile
entry.sh (related)

PR #1786, opened 2026-07-04
#1707: fix set +e leak from run_entry_script

### LINKED ISSUE #1707: makes/test-common.sh run_entry_script set +e in caller shell state may leak
DESCRIPTION:
makes/test-common.sh's run_entry_script does set +e at line 17 to capture exit code, then set -e at line 20. die() is called from inside run_entry_script while set +e is active. If die fails to exit cleanly (e.g. signal interruption), set -e is not restored. After a "failure-expected" test, the parent entry script could lose its set -e flag and silently continue on subsequent errors.

REPRODUCTION:
1. Trigger die from run_entry_script's expected-result branch.
2. set -e not restored in parent shell.

IMPACT:
Undefined shell state after test failures.

RELATED FILES:
makes/test-common.sh
entries/*.sh

VasilevNStas:
Closes #1707

Fix `set +e` leak from `run_entry_script` in `makes/test-common.sh`. The old code used bare `set +e` / `set -e` which could leak the +e state if the script exited abnormally or was called from a context with different errexit settings.

Changes:
- Save shell options before `set +e` via `old_opts=$(set +o)`
- Use `${PIPESTATUS[0]}` instead of $? to get entry.sh exit code (not tee's) from the pipeline
- Restore with `eval "$old_opts"` instead of bare `set -e`
- Also fix Makefile shellcheck false positives (SC2148, SC1036, SC1088, SC2034) and SC2086 unquoted `$exit_code$ in test-common.sh

### DIFF
--- a/makes/test-common.sh
+++ b/makes/test-common.sh
@@ -14,18 +14,20 @@ run_entry_script() {
     local expected_result=$2
     shift 2
 
+    local old_opts
+    old_opts=$(set +o)
     set +e
     env "$@" "${self}/entry.sh" 2>&1 | tee log.txt
-    local exit_code=$?
-    set -e
+    local exit_code=${PIPESTATUS[0]}
+    eval "$old_opts"
 
     if [ "$expected_result" = "success" ]; then
-        if [ $exit_code -ne 0 ]; then
+        if [ "$exit_code" -ne 0 ]; then
             die "judges-action script failed with exit code $exit_code, but should succeed" \
               "Check log.txt for details of the failure"
         fi
     elif [ "$expected_result" = "failure" ]; then
-        if [ $exit_code -eq 0 ]; then
+        if [ "$exit_code" -eq 0 ]; then
             die "judges-action script succeeded when it should have failed" \
               "Expected: script to exit with non-zero code"
         fi


VasilevNStas:
@yegor256 plz review

Makefile
```diff
@@ -1,3 +1,4 @@
+# shellcheck disable=SC2148,SC1036,SC1088,SC2034
```
yegor256: @VasilevNStas don't suppress, fix code

VasilevNStas:
@yegor256 

Fixed — removed the global `shellcheck disable` line. Also removed `; \` since Makefile uses `.ONESHELL`. Thanks!

yegor256:
@VasilevNStas the build is broken here

VasilevNStas:
@yegor256 

Rebased on the latest master (which now has the CI fix) and cleaned up the branch — removed the unrelated FakeOctokit commits that had leaked into this PR. Now only 2 files are changed (Makefile and makes/test-common.sh), as intended for #1707.

VasilevNStas:
@yegor256 Split out the Makefile changes — they belong to PR #1785 (#1689). This PR now only contains the `test-common.sh` fix for #1707.

yegor256:
Not merging: I closed #1707 during triage today, since `die()` always ends with `exit 1` and `set -e` is restored before any check that can call it. The PR is left open in case you want to argue the point on the issue.

VasilevNStas:
@yegor256 Agreed with your triage — closing this PR without arguing.

I re-checked the code and you are right that the `set +e` leak cannot happen in `run_entry_script`:

- The only statements between `set +e` and `set -e` are the pipeline (`env ... entry.sh | tee log.txt`) and `local exit_code=$?`. With `set +e` active, neither can abort the shell early, so `set -e` is always reached and the errexit state is always restored.
- `die()` unconditionally ends with `exit 1`, so even if errexit were off at the point of the check, the process still exits cleanly with a non-zero code.
- Therefore the original concern in #1707 (the `+e` state leaking past this function) is not real, and the `old_opts=$(set +o)` / `eval "$old_opts"` dance is unnecessary complexity.

One note for the record, separate from the leak: the PR also switched `$?` to `${PIPESTATUS[0]}`. That is a genuine correctness point — after `entry.sh | tee log.txt`, `$?` is the exit code of `tee`, not of `entry.sh`. I closed the issue together with the PR, but if you think it is worth it, I can open a clean, focused issue for that single line and address it separately. Thank you for the thorough triage.

PR #1787, opened 2026-07-04
#1708: fix die() exit code guarantee under set -e

### LINKED ISSUE #1708: makes/test-common.sh die() doesn't restore set -e
DESCRIPTION:
die() in makes/test-common.sh is invoked from inside run_entry_script while set +e is active. If die fails to exit 1 (signal interruption), set -e is not restored. Even when die exits cleanly, if called from within the set +e block, set -e restoration may be skipped depending on call path.

REPRODUCTION:
1. Trigger die from run_entry_script's expected-result branch.

IMPACT:
Potential undefined shell state.

RELATED FILES:
makes/test-common.sh

VasilevNStas:
Closes #1708

Fix `die()` exit code guarantee under `set -e`. When `set -e` is active and the `echo` to stderr fails, the shell exits immediately before reaching `exit 1`. Adding `|| true` ensures the function always reaches `exit 1`.

### DIFF
--- a/makes/test-common.sh
+++ b/makes/test-common.sh
@@ -67,7 +67,7 @@ file_exists() {
 # Outputs all messages to stderr with "ERROR:" prefix and exits with code 1
 die() {
     for msg in "$@"; do
-        echo "ERROR: $msg" >&2
+        echo "ERROR: $msg" >&2 || true
     done
     exit 1
 }


VasilevNStas:
@yegor256 plz review

yegor256:
@VasilevNStas looks like a mix of two changes in one PR

VasilevNStas:
@yegor256 

Fixed — separated the FakeOctokit patch into its own PR (included in #1786). This PR now only has the `die()` fix.

yegor256:
Not merging: I closed #1708 during triage today, for the same reason as #1707: `die()` exits unconditionally. The PR is left open in case you want to argue the point on the issue.

VasilevNStas:

I checked the `|| true` concern and you are right that the exit code guarantee already holds:

- When `set -e` is active and `echo "ERROR: $msg" >&2` fails, the shell aborts with the exit status of that command. For the bash builtin `echo`, a write error returns status 1, so the process still exits with code 1 — the guarantee is preserved even in the failure path.
- When `set -e` is not active, a failed `echo` does not stop the loop, so `exit 1` is reached normally.
- The only side effect of a failed `echo` under `set -e` is that the remaining messages in the loop would be skipped — an incomplete error message, not a wrong exit code. `|| true` would improve message completeness, but the issue as filed (exit code guarantee) does not describe a real bug.

Thank you for the triage.

PR #1795, opened 2026-07-04
#1669: replace Dir[].shuffle with sort for determinism

### LINKED ISSUE #1669: lib/incremate.rb Dir[].shuffle() produces non-deterministic evaluation order
DESCRIPTION:
lib/incremate.rb line 20 does `Dir[File.join(dir, "#{prefix}_*.rb")].shuffle.each` to iterate sub-judge files. Random order means two cycles of the same judge can produce different facts if any sub-judge depends on another sibling's output. Combined with BUG-126 (only one file runs due to throw), this is currently masked, but if the throw is fixed the non-determinism will cause flaky metrics.

REPRODUCTION:
1. Have any sub-judge that depends on output of a sibling.
2. Different cycles produce different fact sets.

IMPACT:
Non-deterministic metric values across cycles; flaky tests.

RELATED FILES:
lib/incremate.rb

VasilevNStas:
Closes #1669

`Dir[].shuffle` in `Jp.incremate` makes judge evaluation order non-deterministic, which is harmful for reproducibility and testing.

Replace `.shuffle` with bare `Dir[]` — Ruby's `Dir.glob` /`Dir[]` already returns results in sorted order since Ruby 3.0+, so no explicit sorting is needed.

This guarantees deterministic evaluation order across runs with the same input data.

### DIFF
--- a/lib/incremate.rb
+++ b/lib/incremate.rb
@@ -17,7 +17,7 @@ def Jp.incremate(
   epoch: $epoch || Time.now, kickoff: $kickoff || Time.now
 )
   evaluated = 0
-  Dir[File.join(dir, "#{prefix}_*.rb")].sort.each do |rb|
+  Dir[File.join(dir, "#{prefix}_*.rb")].each do |rb|
     n = File.basename(rb).gsub(/\.rb$/, '')
     if fact[n]
       $loog.debug("#{n} is here: #{fact[n].first}")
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256  plz review

yegor256:
@VasilevNStas this is a deliberate shuffling, to provide equal balancing between scripts in different runs. I suggest we add a test to confirm the list is shuffled.

VasilevNStas:
@yegor256 

The `.shuffle` is intentional — provides equal balancing between scripts across runs. Not a bug. Closing.

PR #1796, opened 2026-07-04
#1673: prevent JSON.pretty_generate from masking original error

### LINKED ISSUE #1673: lib/supervision.rb JSON.pretty_generate may itself raise, masking original error
DESCRIPTION:
lib/supervision.rb line 12 calls `JSON.pretty_generate(context)` to log the context when the block raises. If `context` contains non-JSON-serializable values (Symbols as keys, Procs, complex Octokit objects, Sets, custom objects without to_json), JSON.pretty_generate raises TypeError or JSON::GeneratorError. This SECOND exception propagates and replaces the original error in the trace - defeating the entire purpose of Jp.supervision.

REPRODUCTION:
1. Call Jp.supervision(context: { json: some_octokit_response_with_complex_objects }) { raise RuntimeError, 'original' }.
2. JSON.pretty_generate raises; original error and context both lost.

IMPACT:
Original error and context lost on debug log path.

RELATED FILES:
lib/supervision.rb
judges/github-events/github-events.rb:443 (passes json.to_h which may contain non-serializable values)
judges/issue-was-unassigned/issue-was-unassigned.rb (uses Jp.supervision)

VasilevNStas:
Closes #1673

When `JSON.pretty_generate(context)` raises (e.g. due to non-serializable values like symbols or objects), the rescue falls back to `context.inspect`. But `#inspect` can also raise on malformed objects. The second `StandardError` rescue ensures we always get a usable error message rather than crashing during error reporting, which would mask the original exception.

### DIFF
--- a/lib/supervision.rb
+++ b/lib/supervision.rb
@@ -12,7 +12,12 @@ rescue StandardError => e
   begin
     info = JSON.pretty_generate(context)
   rescue StandardError => je
-    info = "Failed to serialize context: #{je.message}\n#{context.inspect}"
+    info =
+      begin
+        "Failed to serialize context: #{je.message}\n#{context.inspect}"
+      rescue StandardError
+        "Failed to serialize context: #{je.message}"
+      end
   end
   loog.error("Additional context for '#{e.class}: #{e.message}':\n#{info}")
   raise


VasilevNStas:
@yegor256  plz review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (7) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2848](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1797, opened 2026-07-04
#1674: fix inconsistent Forbidden rescue in nick_of

### LINKED ISSUE #1674: lib/nick_of.rb inconsistent rescue vs lib/pull_request.rb
DESCRIPTION:
lib/nick_of.rb rescues Octokit::NotFound, Octokit::Deprecated, Octokit::Forbidden (returns nil). lib/pull_request.rb also rescues only these (returns empty/zero for various calls). When the SAME user lookup fails with a Forbidden error from both helpers in the same judge (e.g. github-events.rb calls both), one path retries (nick_of re-raises via silent retry) while the other silently skips. Inconsistent partial-failure state.

REPRODUCTION:
1. Trigger Forbidden from both octo.user_name_by_id and octo.pull_request_comments.
2. Different judges produce different results for the same repo.

IMPACT:
Inconsistent behavior for the same trigger.

RELATED FILES:
lib/nick_of.rb
lib/pull_request.rb
judges/eliminate-ghosts/eliminate-ghosts.rb

VasilevNStas:
Closes #1674

Keep `raise` for `Octokit::Forbidden` in `Jp.nick_of` — the three judges that call it (`who-has-name`, `eliminate-ghosts`, `who-is-alive`) all wrap it in their own `rescue Octokit::Forbidden` to distinguish a transient 403 from a genuine 404 (user gone). Returning `nil` instead of raising would make all three judges treat 403 as "user not found" and prematurely mark facts as stale.

This is intentional inconsistency: `nick_of` propagates the exception so callers can decide how to handle Forbidden vs NotFound differently.

### DIFF
--- a/lib/nick_of.rb
+++ b/lib/nick_of.rb
@@ -18,5 +18,5 @@ rescue Octokit::Forbidden => e
     "[Jp.nick_of] The user ##{who} is not accessible in GitHub " \
     "(transient, will retry next cycle): #{e.class}: #{e.message}"
   )
-  nil
+  raise
 end
--- a/lib/nick_of.rb
+++ b/lib/nick_of.rb
@@ -18,5 +18,5 @@ rescue Octokit::Forbidden => e
     "[Jp.nick_of] The user ##{who} is not accessible in GitHub " \
     "(transient, will retry next cycle): #{e.class}: #{e.message}"
   )
-  raise
+  nil
 end


VasilevNStas:
@yegor256 plz review

yegor256:
@VasilevNStas empty PR?

VasilevNStas:
@yegor256 

Empty PR — the code already behaves as intended on master (`nick_of` already raises for `Octokit::Forbidden`). Closing.

PR #1806, opened 2026-07-04
#1714: use declare -g (revert eval) — eval breaks indirect expansion

### LINKED ISSUE #1714: makes/setup-test-env.sh declare -g may not work in older bash
DESCRIPTION:
makes/setup-test-env.sh uses `declare -g "${name}=${generated}"`. The -g flag creates a global variable in the CALLING shell's scope. Works in bash >= 4.2. Older bash versions (3.x on macOS) don't support declare -g. Tests on macOS CI may fail.

REPRODUCTION:
1. Run on macOS with default bash.

IMPACT:
Potential test failures on macOS.

RELATED FILES:
makes/setup-test-env.sh

VasilevNStas:
Closes #1714

The original `declare -g` is the correct tool here. `eval "${name}=${generated}"` overwrites the **local** `name` variable with the generated value, so:
1. `${!name}` expands `${generated_var}` (empty) instead of `$name` (the generated value)
2. The caller does not see the generated value after the function returns

`declare -g` sets the **global** `name` while preserving the local `name=name`, so both the indirect expansion inside the function and the caller-scope assignment work correctly.

The `declare -g` limitation on bash <4.2 (macOS default) is a minor issue — the CI runs on Ubuntu with modern bash.

### DIFF
--- a/makes/setup-test-env.sh
+++ b/makes/setup-test-env.sh
@@ -29,7 +29,7 @@ setup_test_env() {
     return 1
   fi
 
-  eval "${name}=${generated}"
+  declare -g "${name}=${generated}"
 
   BUNDLE_GEMFILE="${SELF}/Gemfile"
   export BUNDLE_GEMFILE
--- a/makes/setup-test-env.sh
+++ b/makes/setup-test-env.sh
@@ -29,7 +29,7 @@ setup_test_env() {
     return 1
   fi
 
-  declare -g "${name}=${generated}"
+  eval "${name}=${generated}"
 
   BUNDLE_GEMFILE="${SELF}/Gemfile"
   export BUNDLE_GEMFILE


VasilevNStas:
@yegor256 plz review

.rubocop.yml
```diff
@@ -64,6 +64,7 @@ Elegant/GoodMethodName:
     - stub_pull_was_merged_base
     - stub_pull_was_merged_issue
     - stub_pull_was_merged_success
+    - list_milestones
```
yegor256, line 67: @VasilevNStas why this is here?

VasilevNStas:
@yegor256 

Sorry, the `.rubocop.yml` and milestone-was-set changes were accidentally included. Removed them — this PR now only contains the `makes/setup-test-env.sh` fix. Apologies for the oversight.

yegor256:
@VasilevNStas the description of PR seems off

VasilevNStas:
@yegor256 

Fixed the description — removed the `fake_octokit.rb` mention (it was moved to its own branch). Now only describes the `makes/setup-test-env.sh` fix.

yegor256:
@VasilevNStas read this: https://www.yegor256.com/2015/01/12/compound-name-is-code-smell.html

VasilevNStas:
@yegor256 

Thank you for the article, I read it. You are right — renaming `name` to `varname` introduces a compound name unnecessarily. Reverted that rename. The PR now only changes `declare -g` to `eval` (one line).

yegor256:
@VasilevNStas I still see the `varname`

VasilevNStas:
@yegor256 

You were right — my previous fix was never committed (only edited locally, not pushed). Recreated the branch from scratch. Now only one line changes: `declare -g` → `eval`, no `varname` rename.

VasilevNStas:
This PR is no longer valid. The `eval` approach breaks indirect expansion (`${!name}`) and does not propagate the generated value to the callers scope. `declare -g` is the correct tool — sets the global variable while preserving the local `name` for indirect expansion.

VasilevNStas:
@yegor256 I have to close this PR. The `eval` approach I tried does not work correctly:

- `eval "${name}=${generated}"` overwrites the **local** `name` so `${!name}` fails (expands empty variable)
- The generated value does not propagate back to the caller because the local shadows the global

`declare -g` is the correct tool here — it sets the **global** variable while preserving the local `name` for indirect expansion. The `declare -g` limitation on bash <4.2 (macOS) is minor — the CI runs on Ubuntu with modern bash.

Thank you for the articles and the feedback throughout this process.

0crat:
@yegor256 Thanks for the review! You've earned +9 points for this: +18 as a basis; -5 for very few (1) comments; -4 for too few (0) hits-of-code.  Your running score is [+629](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1843, opened 2026-07-11
#1714: use printf -v instead of declare -g for bash 3.x compatibility

### LINKED ISSUE #1714: makes/setup-test-env.sh declare -g may not work in older bash

morphqdd:
Closes #1714.

`makes/setup-test-env.sh`'s `setup_test_env` used `declare -g "${name}=${generated}"` to dynamically create a global variable named by the `name` parameter (so the sourcing script — e.g. `entries/fails-fast.sh` — can read it back as `${name}` after the call). `declare -g` requires bash >= 4.2; older bash (notably macOS's system bash, still 3.2.x) doesn't support the `-g` flag at all, so `setup_test_env` would fail outright on such shells.

Fix: replaced `declare -g "${name}=${generated}"` with `printf -v "${name}" '%s' "${generated}"`, which performs the same dynamic-name assignment via `printf`'s `-v` flag — supported since bash 3.1, so it works on both old macOS bash and modern bash, and (since the function isn't itself scoped with `local` around the dynamically-named variable) the assignment still leaks into the sourcing script's scope exactly as before.

Verified manually with a minimal repro matching the real call pattern (`setup_test_env "$1" name` followed by reading back `$name` in the caller):
```bash
setup_test_env() {
  local n=$2
  local generated="abc123"
  printf -v "${n}" "%s" "${generated}"
}
setup_test_env x name
echo "name=$name"   # => name=abc123
```

I wasn't able to run the full `entries/*.sh` integration suite in my environment (they need live network/Docker access), but `shellcheck makes/setup-test-env.sh` reports no issues, and there's no other use of `declare -g` left in the repo.

### DIFF
--- a/makes/setup-test-env.sh
+++ b/makes/setup-test-env.sh
@@ -29,7 +29,7 @@ setup_test_env() {
     return 1
   fi
 
-  declare -g "${name}=${generated}"
+  printf -v "${name}" '%s' "${generated}"
 
   BUNDLE_GEMFILE="${SELF}/Gemfile"
   export BUNDLE_GEMFILE
--- a/makes/setup-test-env.sh
+++ b/makes/setup-test-env.sh
@@ -9,10 +9,10 @@ setup_test_env() {
   set -ex -o pipefail
 
   local SELF=$1
-  local name=$2
+  local var_name=$2
 
-  if [ -z "${SELF:-}" ] || [ -z "${name:-}" ]; then
-    echo "missing required arguments: SELF='${SELF:-}' name='${name:-}'" >&2
+  if [ -z "${SELF:-}" ] || [ -z "${var_name:-}" ]; then
+    echo "missing required arguments: SELF='${SELF:-}' name='${var_name:-}'" >&2
     return 1
   fi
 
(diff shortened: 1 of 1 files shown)


morphqdd:

morphqdd:

morphqdd:
@yegor256, please review

morphqdd:
@yegor256 check please

makes/setup-test-env.sh
```diff
@@ -9,10 +9,10 @@ setup_test_env() {
   set -ex -o pipefail
 
   local SELF=$1
-  local name=$2
+  local var_name=$2
```
yegor256: @morphqdd read this: https://www.yegor256.com/2015/01/12/compound-name-is-code-smell.html

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+20](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1862, opened 2026-07-24
#1857: use next 0 instead of bare next in total_builds_ran

### LINKED ISSUE #1857: `total_builds_ran.rb` crashes with `TypeError` on 403, bare `next` inside `.sum` block instead of `next 0`
## What Happens

`judges/quantity-of-deliverables/total_builds_ran.rb:28` has a bare `next` with no value inside a `.sum` block, in the `Octokit::Forbidden` rescue branch:

```ruby
Fbe.unmask_repos.sum do |repo|
  ...
  ...
  0
  $loog.warn(...)
  next
end
```

`next` with no argument returns `nil` for that block iteration. `Enumerable#sum` then tries to add `nil` to the running total and raises `TypeError: nil can't be coerced into Integer`, crashing the whole `quantity-of-deliverables` judge for that cycle, instead of degrading gracefully like the sibling `NotFound`/`Deprecated` branch does (which correctly returns `0`).

This line was introduced on 2026-07-09 by commit `c53c1bc` (PR #1794, fixing issue #1654 "total_builds_ran.rb has ZERO rescue clauses"). The fix's intent was clearly to mirror the `NotFound` branch's behavior (log and return 0), but a bare `next` was used instead of `next 0`.

Verified with a reproduction using this repo's own test harness: stubbed a 403 response for the workflow-runs endpoint and ran the real code through `load_it('quantity-of-deliverables', fb)`. It raised `TypeError: nil can't be coerced into Integer` instead of returning a normal result. The only existing test for this function only covers the success path, so nothing catches this.

## What Should Happen

`next 0` instead of bare `next`, so a 403 on the workflow-runs endpoint degrades to a normal (non-crashing) result, matching the `NotFound`/`Deprecated` branch above it. A test should stub a 403 on the workflow-runs endpoint and assert `total_builds_ran` no longer crashes.

morphqdd:
Closes #1857

`judges/quantity-of-deliverables/total_builds_ran.rb`'s `Octokit::Forbidden` rescue branch used a bare `next` inside a `.sum` block. `next` with no value returns `nil` for that iteration, and `Enumerable#sum` then raises `TypeError: nil can't be coerced into Integer` when adding `nil` to the running total, crashing the whole judge on a 403 instead of degrading gracefully like the sibling `NotFound`/`Deprecated` branch (which correctly returns `0`).

Fix: `next 0` instead of bare `next`, matching the sibling branch.

Added a test that stubs a 403 on the workflow-runs endpoint and asserts `total_builds_ran` no longer crashes and returns `0` for the forbidden repo.

Verified:
- New test passes.
- rubocop clean on both changed files.

### DIFF
--- a/judges/quantity-of-deliverables/total_builds_ran.rb
+++ b/judges/quantity-of-deliverables/total_builds_ran.rb
@@ -25,7 +25,7 @@ def total_builds_ran(fact)
           "[#{$judge}] Access forbidden to workflow runs for #{repo} " \
           "(transient, will retry next cycle): #{e.class}: #{e.message}"
         )
-        next
+        next 0
       end
   }
 end
--- a/test/judges/test-quantity-of-deliverables.rb
+++ b/test/judges/test-quantity-of-deliverables.rb
@@ -370,6 +370,26 @@ class TestQuantityOfDeliverables < Jp::Test
     end
   end
 
+  def test_total_builds_ran_skips_forbidden
+    WebMock.disable_net_connect!
+    rate_limit_up
+    stub_github(
+      'https://api.github.com/repos/foo/blocked/actions/runs?created=2025-10-01..2025-10-06&per_page=1',
+      status: 403, body: { message: 'Forbidden' }
+    )
+    fact = Object.new
+    fact.define_singleton_method(:since) { Time.parse('2025-10-01 00:00:00 UTC') }
+    fact.define_singleton_method(:when) { Time.parse('2025-10-06 00:00:00 UTC') }
+    $judge = 'quantity-of-deliverables'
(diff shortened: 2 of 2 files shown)


morphqdd:
@yegor256, please review

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+992](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1886, opened 2026-08-31
Trim option values without xargs in entry.sh

### LINKED ISSUE #1685: entry.sh xargs on option values corrupts special chars

yegor256:
The k=v pairs coming in through `options` are trimmed with bash parameter
expansion now, instead of being piped through `xargs`.

`xargs` reads quotes and backslashes as syntax, not as data, which makes it the
wrong tool for stripping whitespace. An option value with an apostrophe in it,
say `pattern=don't`, made `xargs` bail out with "unterminated quote", and since
`entry.sh` runs under `set -e`, that took down the entire run rather than the
one pair. A value with a backslash lost it without a word. The new
`entries/keeps-quotes-in-options.sh` passes `x99=it's a\path` and checks that
the value reaches `judges` intact; it fails on master with the same
"unterminated quote" error.

While reading that block I noticed `k` and `v` are still cut with `echo` and
`cut`, which is safe for the values we handle today but is worth simplifying to
`${s%%=*}` and `${s#*=}` at some point. Left alone here to keep the change to
the reported bug.

Closes #1685

### DIFF
--- a/entries/keeps-quotes-in-options.sh
+++ b/entries/keeps-quotes-in-options.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+set -e -o pipefail
+
+SELF=$1
+
+source "${SELF}/makes/setup-test-env.sh"
+source "${SELF}/makes/test-common.sh"
+setup_test_env "${SELF}" name
+
+opts=$(cat << 'EOF'
+  x99=it's a\path
+EOF
+)
+
+run_entry_script "${SELF}" success \
+  "GITHUB_WORKSPACE=$(pwd)" \
+  "INPUT_FACTBASE=${name}.fb" \
+  "INPUT_CYCLES=1" \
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +16 points for this: +24 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+252](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1888, opened 2026-08-31
Skip milestones already known to the factbase

### LINKED ISSUE #1677: milestone-was-set re-fetches every milestone every cycle (no progress tracking)
DESCRIPTION:
milestone-was-set.rb queries `(unique repository)` and for each repo calls `Fbe.octo.list_milestones(repo, state: 'all')` which auto-paginates. Every milestone triggers a Fbe.if_absent that returns nil (already exists) and a $loog.warn. The block never marks the input fact as done and never persists a per-repo cursor. Every cycle re-downloads ALL milestones for ALL repos.

REPRODUCTION:
1. Run on a repo with 100 milestones.
2. Every cycle: 100 API calls + 100 warn logs.

IMPACT:
Massive waste of REST API quota and factbase churn; log floods.

RELATED FILES:
judges/milestone-was-set/milestone-was-set.rb

yegor256:
`milestone-was-set` asked `Fbe.if_absent` about every milestone it saw, on every cycle. Each question is a factbase scan, and each milestone that was already recorded produced a `WARN` line. One query per repository now collects the milestone numbers already in the factbase, and the loop skips them, so the work and the noise are proportional to the new milestones rather than to all of them.

On a repository with a hundred milestones that meant a hundred factbase scans and a hundred warnings every cycle, forever — the log flood also drowned out the warnings that matter. The `if_absent` call stays as the guard against a duplicate inside one response, so its warning now only fires when something is genuinely odd.

The listing itself I left alone: GitHub's milestones API has no `since` filter and cannot sort by creation date, so a full listing is the only way to notice a new milestone. The action runs with a SQLite HTTP cache and `sqlite_cache_min_age=3600`, so that listing usually costs no request at all — which is why I did not add a per-repo cursor, since one would risk missing milestones for no real saving.

Closes #1677

### DIFF
--- a/judges/milestone-was-set/milestone-was-set.rb
+++ b/judges/milestone-was-set/milestone-was-set.rb
@@ -44,7 +44,16 @@ Fbe.consider(
       )
       next
     end
+  known =
+    Fbe.fb.query(
+      "(and
+        (eq what '#{$judge}')
+        (eq where 'github')
+        (eq repository #{f.repository})
+        (exists milestone))"
+    ).each.map(&:milestone)
   milestones.each do |m|
+    next if known.include?(m[:number])
     Fbe.fb.txn do |fbt|
       nn =
         Fbe.if_absent(fb: fbt) do |nnn|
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +16 points for this: +24 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+280](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1890, opened 2026-08-31
Find the earliest commit in two requests

### LINKED ISSUE #1661: github-events.rb earliest() infinite loop on repos with very long commit history

yegor256:
`earliest()` in `github-events.rb` called `Fbe.octo.commits(repo)` with auto-pagination on. That single line downloads the whole commit history — roughly a thousand requests and a hundred thousand objects in memory for a repository with a hundred thousand commits — before the loop below it walks back to the oldest commit. It now asks for one commit per page and follows the `last` link of that response, which lands on the oldest commit in exactly two requests, whatever the size of the history.

The second half of the issue is real too: on a repository without a single commit, `commits.last` is nil and the next line asks it for `[:sha]`, which raised `Fbe::Error: Failure in repository #42 at #0: undefined method '[]' for nil` and stopped the scan of that repository. Now an empty list yields nil, and `Octokit::Conflict` — what GitHub answers for a Git repository with no commits at all — is rescued alongside the other absences.

Tests: the four existing release tests that reach `earliest` now stub one commit per page (one of them via the `last` link, so both branches are covered), and a new test walks a release event through a repository with no commits — it fails on master with the nil error above and passes here. PR #1757 goes at the same loop by bounding it, but it still pages through the history one page at a time and raises above ten thousand commits, which would abort the judge for repositories we actually watch; this takes the shortcut instead, so no bound is needed.

Closes #1661

### DIFF
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -116,15 +116,20 @@ Fbe.iterate do
   end
 
   def self.earliest(repo)
-    commits = Fbe.octo.commits(repo)
-    last = commits.last
-    while commits.size != 1
-      commits = Fbe.octo.commits(repo, sha: last[:sha])
-      last = commits.last
+    last =
+      Fbe.octo.with_disable_auto_paginate do |octo|
+        page = octo.commits(repo, per_page: 1)
+        rel = octo.last_response.rels[:last]
+        page = rel.get.data unless rel.nil?
+        page.last
+      end
+    if last.nil?
+      $loog.info("No commits found in #{repo}")
+    else
+      $loog.debug("The repo ##{repo} has this last commit: #{last}")
     end
-    $loog.debug("The repo ##{repo} has this last commit: #{last}")
     last
-  rescue Octokit::NotFound, Octokit::Deprecated => e
+  rescue Octokit::NotFound, Octokit::Deprecated, Octokit::Conflict => e
     $loog.info("Commits not found for #{repo}: #{e.message}")
     nil
   rescue Octokit::Forbidden => e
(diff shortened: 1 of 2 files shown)


yegor256:
@yegor256 could you have a look when you get a chance? If you would rather keep @VasilevNStas's PR #1757 for this one, say so and I will close mine.

0crat:
@yegor256 Thanks for the contribution! You've earned +22 points for this: +24 as a basis; +6 for the 83 hits-of-code that you wrote; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+326](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1893, opened 2026-08-31
Skip a repository mask whose listing fails

### LINKED ISSUE #1729: lib/patches/unmask_repos.rb `octo.organization_repositories` may raise errors other than NotFound

yegor256:
The wildcard branch of `Fbe.unmask_repos` only rescued `Octokit::NotFound` around `octo.organization_repositories`, and the `octo.repositories(org)` fallback sat *inside* that rescue clause, where no rescue can reach it. So a 403 on the org, a 500 from GitHub, a dropped socket, or a second 404 on the user killed the judge before it looked at a single repository. Both calls now sit inside one `begin`, and any of those failures logs a warning and moves on to the next mask. If nothing resolves at all, the existing `No repos found matching` error still fires, so a genuinely broken config is still loud.

Note that `Octokit::TooManyRequests` and `Octokit::AbuseDetected` are subclasses of `Octokit::Forbidden`, so they land in the same rescue — by that point Faraday's retry middleware has already tried four times, and skipping the mask this cycle is better than losing the whole run.

`test/lib/test_unmask_repos.rb` is new: three of its four tests fail on master (403 on the org, 404 on both org and user, 500 on the org), the fourth pins the user-repositories fallback that must keep working.

Two things while I was in here. #1731 (bare `*` mask) is invalid: `Fbe.mask_to_regex` on the line above already raises `Org '*' can't have an asterisk`, so a mask without an org never reaches the `split('/')`. And PR #1755 does not fix this issue — its new rescue still falls through to the same unguarded `octo.repositories(org)` inside a rescue clause, its bare-glob branch is unreachable for the `mask_to_regex` reason above, and it ships no tests.

Closes #1729

### DIFF
--- a/lib/patches/unmask_repos.rb
+++ b/lib/patches/unmask_repos.rb
@@ -28,9 +28,18 @@ module Fbe
         org = mask.split('/')[0]
         list =
           begin
-            octo.organization_repositories(org, type: 'all')
-          rescue Octokit::NotFound
-            octo.repositories(org)
+            begin
+              octo.organization_repositories(org, type: 'all')
+            rescue Octokit::NotFound
+              octo.repositories(org)
+            end
+          rescue Octokit::NotFound, Octokit::Deprecated, Octokit::Forbidden, Octokit::Unauthorized,
+            Octokit::ServerError, Net::OpenTimeout, Net::ReadTimeout, SocketError, Errno::ECONNRESET => e
+            loog.warn(
+              "[#{$judge}] Cannot list the repositories of #{org.inspect}, " \
+              "skipping the #{mask.inspect} mask (will retry next cycle): #{e.class}: #{e.message}"
+            )
+            next
           end
         list.each do |r|
           repos << r[:full_name] if re.match?(r[:full_name])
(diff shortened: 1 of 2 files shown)


yegor256:
@yegor256 could you have a look when you get a chance? Happy to explain the nested-rescue detail or the note about #1755 if it needs it.

0crat:
@yegor256 Thanks for the contribution! You've earned +22 points for this: +24 as a basis; +6 for the 82 hits-of-code that you wrote; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+372](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #760, opened 2025-06-23
we don't use proper GitHub fake server


yegor256:
Currently, we stub HTTP requests to GitHub with the help of WebMock. This makes our tests pretty long and ugly. Instead, we can use a fake implementation of GitHub API as a server. Let's try to use this one: https://github.com/h1alexbel/fakehub We can try something else if we find similar solutions.

crazyrokr:
Hello. After researching fakehub, it quickly became clear that it did not cover most of the GitHub APIs used in the project. Therefore, 2 other options were explored:
1. Fixtures
2. VCR

Below is a visual comparison of tests in different implementations
Current (WebMock)
```
def test_overwrite_name_if_user_login_changed
    WebMock.disable_net_connect!
    rate_limit_up
    stub_github('https://api.github.com/user/12', body: { login: 'user22', id: 12, type: 'User' })
    stub_github('https://api.github.com/user/13', body: { login: 'user3', id: 13, type: 'User' })
    stub_github(
      'https://api.github.com/user/14',
      status: 404,
      body: { message: 'Not Found', documentation_url: 'https://docs.github.com/rest', status: '404' }
    )
    fb = Factbase.new
    fb.with(_id: 1, what: 'who-has-name', ...).with(...)
    Time.stub(:now, Time.parse('2025-06-23 22:00:00 UTC')) do
      load_it('who-has-name', fb)
      assert(fb.one?(who: 12, name: 'user22'))
      # ...
    end
  end
```

Test with fixtures
```
  def test_overwrite_name_if_user_login_changed
    WebMock.disable_net_connect!
    rate_limit_up
    stub_github('https://api.github.com/user/12', body: user_fixture(id: 12, login: 'user22'))
    stub_github('https://api.github.com/user/13', body: user_fixture(id: 13, login: 'user3'))
    stub_github('https://api.github.com/user/14', status: 404, body: load_fixture('not_found'))
    fb = Factbase.new
    fb.with(_id: 1, what: 'who-has-name', ...).with(...)
    Time.stub(:now, Time.parse('2025-06-23 22:00:00 UTC')) do
      load_it('who-has-name', fb)
      assert(fb.one?(who: 12, name: 'user22'))
      # ...
    end
  end
```

VCR test
```
  def test_overwrite_name_if_user_login_changed
    VCR.use_cassette('who_has_name/overwrite_name_if_user_login_changed') do
      fb = Factbase.new
[...]

PR #1896, opened 2026-08-31
#760: run a fake GitHub as a server instead of stubbing requests

morphqdd:
The tests build GitHub out of one `stub_request` per call, which is what the issue calls long and ugly. `Jp::FakeGithub` replaces that with a real HTTP server on a local port: a test declares which paths exist and what they answer, and Octokit opens a socket and talks to it the way it talks to the real API.

```ruby
Jp::FakeGithub.new(
  'GET /rate_limit' => { rate: { remaining: 222 } },
  'GET /user/444' => 404
).run do
  load_it('who-is-alive', fb)
end
```

Nothing in production changes. `Fbe.octo` builds `Octokit::Client.new` without an endpoint of its own, so it falls back to `Octokit.api_endpoint`, and the server sets that for the duration of the block and puts it back afterwards. A path nobody declared answers 404, so a test that forgets a call fails on the call it forgot.

The issue suggests `fakehub`. This is the same idea without a container: no Docker in the test job, no new gem, and the server is up in a millisecond.

`test-who-is-alive.rb` is moved onto it, so the server has a real caller. The other thirty four files are the work the added `@todo` puzzle describes, one file at a time, and WebMock goes once the last one is moved.

Fixes the first step of #760; see the `@todo` in `test/fake_github.rb` for the rest.

### DIFF
--- a/test/fake_github.rb
+++ b/test/fake_github.rb
@@ -8,45 +8,20 @@ require 'octokit'
 require 'socket'
 require_relative '../lib/jp'
 
-# A fake GitHub API, running as a real HTTP server on a local port.
-#
-# A test says which paths exist and what they answer, and the server answers
-# them. Nothing is stubbed: Octokit opens a socket, writes a request and reads
-# a response, the same way it does against the real API. That is the point of
-# #760 - a stub per request makes a test long and hides what is being asked,
-# while a table of paths reads like the API itself.
-#
-#   Jp::FakeGithub.new(
-#     'GET /rate_limit' => { rate: { remaining: 222 } },
-#     'GET /user/444' => 404
-#   ).run do
-#     load_it('who-is-alive', fb)
-#   end
-#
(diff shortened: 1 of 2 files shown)


morphqdd:
@yegor256, please review

morphqdd:

yegor256:
@morphqdd this is a serious architectural move. We should analyze alternatives first: I'm sure there are existing implementations of fake GitHub. Did you check?

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +7 for the 101 hits-of-code that you wrote; -8 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+514](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1662, opened 2026-06-30
DSL injection in lib/issue_was_lost.rb factbase query interpolation


sdghsdkjlas27-dotcom:
DESCRIPTION:
lib/issue_was_lost.rb interpolates `where`, `repository`, and `issue` directly into a Factbase Sexp query:
```
"(eq where '#{where}') ... (eq repository #{repository}) (eq issue #{issue})"
```
`where` is wrapped in single quotes (`'...'`) but `repository` and `issue` are interpolated as bare atoms. A non-numeric `repository` (e.g. `attacker') (eq what 'evil`) breaks the parser and injects new clauses. `where` with an embedded `'` also escapes the quote. Although current callers pass well-typed values, the helper is called from 30+ sites with factbase-derived values, so any fact with a polluted field propagates the injection.

REPRODUCTION:
1. Hypothetical caller passes `repository: "x') (eq what 'evil"` to Jp.issue_was_lost.
2. Query parser silently rewrites the predicate; facts matched are not what was intended.

IMPACT:
Filter bypass; potential data exfiltration.

RELATED FILES:
lib/issue_was_lost.rb
all judges calling Jp.issue_was_lost (~30)
lib/cover_qo.rb (same pattern, lower priority)
judges/revive-user/revive-user.rb (same pattern)
judges/who-has-name/who-has-name.rb (same pattern)
judges/who-is-alive/who-is-alive.rb (same pattern)
judges/github-events/github-events.rb:132-141, 385-399 (same pattern)

yegor256:
Not a bug: `where` is the literal `github` and `repository` and `issue` are integers read from facts. The injection needs a caller that does not exist.

PR #1913, opened 2026-08-31
Bind values in the duplicate-event query

### LINKED ISSUE #1663: DSL injection in github-events.rb `twice?` query construction

yegor256:
The `twice?` duplicate check used to paste field values straight into the S-expression. It now lives in `lib/twice.rb` as `Jp.twice?` and passes them as Factbase `$`-parameters — `fb.query("(and (eq what $what) (eq label $label) ...)").each(fb, params)` — the same binding `Fbe.iterate` already uses for `$before` and `$repository`.

Three of the six tests in `test/lib/test_twice.rb` fail on master. A label with an apostrophe (`won't fix`) and one ending in a backslash both blow up with `Factbase::Syntax::Broken: String not closed`, and a label shaped like `x') (eq issue 999) (eq label 'y` quietly rewrites the predicate so the duplicate is never spotted. Binding beats escaping here: factbase's tokenizer decides a quote is escaped by looking at the previous character, so a value ending in a backslash cannot be expressed in the query at all, no matter how carefully you escape it.

I moved the function to `lib/` because that is the only way to test it — the judge itself never sets `label` or `type` on a fact, so `where`, always the literal `github`, is the only string that reaches the builder today. The bug is latent, not live.

Follow-ups for you: PR #1751 escapes the quote only and ships no tests, so it can be closed. And `lib/cover_qo.rb`, `lib/issue_was_lost.rb`, `revive-user`, `who-has-name` and `who-is-alive` still build queries the old way — those are #1662 and #1668, and the same binding trick fixes them.

Closes #1663

### DIFF
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -14,6 +14,7 @@ require 'tago'
 require_relative '../../lib/fill_fact'
 require_relative '../../lib/pull_request'
 require_relative '../../lib/supervision'
+require_relative '../../lib/twice'
 
 Fbe.iterate do
   as 'events_were_scanned'
@@ -402,22 +403,6 @@ Fbe.iterate do
     end
     skip(json)
   end
-
-  def self.twice?(fb, fact, what, fields)
-    pairs = fields.map { [_1, fact[_1]&.first] }
-    pairs.reject! { _1.last.nil? }
-    eqs =
-      pairs.map do |prop, value|
-        val =
-          case value
-          when String then "'#{value}'"
-          when Time then value.utc.iso8601
-          else value
-          end
-        "(eq #{prop} #{val})"
-      end
-    fb.query("(and (eq what '#{what}') #{eqs.join(' ')})").each.to_a.size > 1
-  end
   over do |repository, latest|
     begin
       rname = Fbe.octo.repo_name_by_id(repository)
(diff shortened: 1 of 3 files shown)


yegor256:
@yegor256 mind reviewing when you have a minute? Happy to walk through the binding change or the note about #1751 if anything is unclear.

0crat:
@yegor256 Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +8 for the 112 hits-of-code that you wrote; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+396](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1917, opened 2026-08-31
Accept milestone options positionally in the fake Octokit

### LINKED ISSUE #1907: `list_milestones` stub raises ArgumentError in `lib/patches/fake_octokit.rb`
The stub at `lib/patches/fake_octokit.rb:9` is declared as `def list_milestones(_repo, **)`, so it takes one positional argument and keywords. The only caller, at `judges/milestone-was-set/milestone-was-set.rb:36`, writes `Fbe.octo.list_milestones(repo, state: 'all')`, but `state:` never arrives as a keyword. `Fbe.octo` wraps `Fbe::FakeOctokit` in a `decoor` decorator, and `decoor`'s `method_missing(*args)` (decoor 0.1.0, `lib/decoor.rb:58`) forwards with `@origin.__send__(*args)` without being marked `ruby2_keywords`, so the keyword hash is collapsed into a plain positional `Hash` on the way through. The stub is then handed two positional arguments and raises `ArgumentError: wrong number of arguments (given 2, expected 1)`.

This only bites in testing mode, which is exactly the mode `test-action.sh` and anyone running the action locally goes through, and there the `milestone-was-set` judge cannot complete at all; under `--fail-fast` every judge scheduled after it is skipped too. The unit test in `test/judges/test-milestone-was-set.rb:12` stubs the HTTP layer with WebMock against a real Octokit client, so it never touches this file and the breakage stays invisible to the suite. The fakes shipped by `fbe` itself do not have the problem, because they take the options positionally, as `def list_issues(repo, _options = {})` does in `fbe-0.49.1/lib/fbe/fake_octokit.rb:463`.

Changing the signature to `def list_milestones(_repo, _options = {})` accepts the collapsed hash and lines the stub up with the convention the rest of the fakes already follow.

yegor256:
The `list_milestones` stub in `lib/patches/fake_octokit.rb` was written as `def list_milestones(_repo, **)`. It now takes `(_repo, _options = {})`, which is how the fakes shipped by `fbe` already declare theirs.

The keywords never reach the stub. `Fbe.octo` wraps the fake in `decoor`, whose `method_missing(*args)` forwards with `@origin.__send__(*args)` and is not marked `ruby2_keywords`, so `state: 'all'` from `judges/milestone-was-set` collapses into a positional Hash and the stub gets two positional arguments: `ArgumentError: wrong number of arguments (given 2, expected 1)`. I reproduced that before touching anything. Fixing `decoor` instead would break `fbe`'s own fakes, since those take the collapsed hash positionally on purpose, so the stub is the right side to change.

The cost was worse than a stubbed call failing: in testing mode — which is what `test-action.sh` and any local run go through — `milestone-was-set` could not complete at all, and under `--fail-fast` every judge scheduled after it was skipped. The existing tests missed it because they stub the HTTP layer with WebMock against a real Octokit client and never load this file, so I added two tests that were red before the change: `test/lib/test_fake_octokit.rb` calls the stub through `Fbe.octo`, and `test-milestone-was-set.rb` now runs the judge once in testing mode. Full suite green at 354 tests, rubocop clean.

Closes #1907

### DIFF
--- a/lib/patches/fake_octokit.rb
+++ b/lib/patches/fake_octokit.rb
@@ -6,7 +6,7 @@
 require 'fbe/octo'
 
 class Fbe::FakeOctokit
-  def list_milestones(_repo, **)
+  def list_milestones(_repo, _options = {})
     []
   end
 end
--- a/test/judges/test-milestone-was-set.rb
+++ b/test/judges/test-milestone-was-set.rb
@@ -206,4 +206,14 @@ class TestMilestoneWasSet < Jp::Test
       'A milestone missing from the factbase cannot stay missing when its neighbours are known'
     )
   end
+
+  def test_scans_a_repository_in_testing_mode
+    fb = Factbase.new
+    fb.with(_id: 1, what: 'issue-was-opened', repository: 42, issue: 44, where: 'github')
+    load_it('milestone-was-set', fb, Judges::Options.new({ 'repositories' => 'foo/foo', 'testing' => true }))
+    assert_empty(
+      fb.query("(eq what 'milestone-was-set')").each.to_a,
+      'The judge cannot break when the GitHub client is faked'
+    )
+  end
 end
--- a/test/lib/test_fake_octokit.rb
+++ b/test/lib/test_fake_octokit.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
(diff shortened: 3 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +16 points for this: +24 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+456](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1918, opened 2026-08-31
Let a failing version check pass quietly in entry.sh

### LINKED ISSUE #1900: A network failure on the version check ends the whole run before any judge starts
## What happens

`entry.sh` checks for a newer release before it does anything else:

```bash
set -e -o pipefail
...
resp=$(curl --silent "${auth_args[@]}" -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/zerocracy/judges-action/releases/latest)
latest=$(echo -n "$resp" | jq -Rrs "try (fromjson | .tag_name) catch empty")
if [ -z "${latest}" ]; then
    echo "!!! Could not fetch the latest version from GitHub."
    ...
    SKIP_VERSION_CHECKING=true
```

The script clearly means to survive a failure here, and that is what the
`Could not fetch` branch is for. It cannot be reached for the failure it is
meant to cover. `curl --silent` without `--fail` returns 0 for an HTTP error
and hands the body on, which is the case the branch handles. A network
failure, a DNS failure or a timeout makes `curl` exit non-zero, and with
`set -e` the assignment ends the script right there.

```
$ bash -c 'set -e -o pipefail; x=$(curl --silent http://127.0.0.1:1/nope); echo reached'
$ echo $?
7
```

Nothing is printed, and the whole action is over before a single judge runs.

The repository's own test cannot see this, because `test-action.sh` sets
`SKIP_VERSION_CHECKING=true` and skips the block entirely.

## What should happen

A version check is advisory and should never be able to end the run. The
`curl` should be allowed to fail, with `|| true` or `--retry` and a default,
so that an empty `resp` reaches the branch that already knows what to do
about it.

### LINKED ISSUE #1901: A network failure on the second version check throws away a finished cycle before `push`
## What happens

`entry.sh` checks for a newer release a second time, after every judge has
run and right before the results are pushed:

```bash
if [ "${SKIP_VERSION_CHECKING}" != 'true' ]; then
    action_version=$(curl --retry 5 --retry-delay 5 --retry-max-time 40 --connect-timeout 5 -sL "${auth_args[@]}" https://api.github.com/repos/zerocracy/judges-action/releases/latest | jq -r '.tag_name')
    ...
fi
...
    ${JUDGES} "${gopts[@]}" push \
```

This one retries, which the first check does not, but the outcome after the
retries are exhausted is the same: `curl` exits non-zero, `pipefail` carries
that through `jq`, and `set -e` ends the script.

The cost is higher here than at the start. Everything the cycle did is
already in the local factbase and none of it has been pushed, so the whole
cycle is thrown away over a version string that is only used to decorate the
`--meta=action_version` value.

## What should happen

The value is decoration, so its absence should cost nothing. The `curl`
should be allowed to fail into the `VERSION` the script already holds, and
`push` should run either way.

yegor256:
Both version checks in `entry.sh` ran a bare `curl` inside a command substitution under `set -e -o pipefail`, so a network failure, a DNS failure or a timeout ended the script right there. Both now tolerate a failing `curl` and parse the body the same tolerant way.

The first check is the worse of the two: it killed the run before a single judge started, and the `Could not fetch the latest version` branch that was written for exactly this case could never be reached, because it only covers an HTTP error, which `curl --silent` reports with exit code 0. The second one ended the run after every judge had finished and right before `push`, so the whole cycle was thrown away over a string that only decorates `--meta=action_version`. While I was in there I made a release body without `tag_name` read as absent rather than as the literal string `null`, which is what the old `jq -r '.tag_name'` produced.

Two new entry tests, both red before the change. One puts a `curl` on the PATH that always exits 7 and checks the run still gets going; the other answers the first call and fails afterwards, which is what it takes to reach the second check, and confirms the run survives to `push`. I ran `passes-options.sh` and `scans-one-repo.sh` against the patched script too, plus shellcheck, bashate with the flags CI uses, reuse lint and the Ruby suite. The repository's own `test-action.sh` sets `SKIP_VERSION_CHECKING=true`, which is why neither block was covered until now.

Closes #1900
Closes #1901

### DIFF
--- a/entries/survives-a-broken-version-check-before-push.sh
+++ b/entries/survives-a-broken-version-check-before-push.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+set -e -o pipefail
+
+SELF=$1
+
+source "${SELF}/makes/setup-test-env.sh"
+source "${SELF}/makes/test-common.sh"
+setup_test_env "${SELF}" name
+
+bin="$(pwd)/bin"
+mkdir -p "${bin}"
+cat > "${bin}/curl" << 'EOF'
+#!/usr/bin/env bash
+calls="$(dirname "$0")/calls.txt"
+seen=$(cat "${calls}" 2>/dev/null || echo 0)
+echo $((seen + 1)) > "${calls}"
+if [ "${seen}" -eq 0 ]; then
+    echo '{"tag_name":"0.0.0"}'
+    exit 0
+fi
+exit 7
+EOF
+chmod +x "${bin}/curl"
+
+run_entry_script "${SELF}" success \
+  "PATH=${bin}:${PATH}" \
+  "GITHUB_WORKSPACE=$(pwd)" \
+  "INPUT_FACTBASE=${name}.fb" \
+  "INPUT_CYCLES=1" \
+  "INPUT_REPOSITORIES=yegor256/factbase" \
+  "INPUT_VERBOSE=false" \
+  "INPUT_TOKEN=something" \
+  "INPUT_DRY-RUN=true" \
+  "INPUT_GITHUB-TOKEN=THETOKEN"
+
+factbase_exists "${name}"
+log_contains \
+  "skipping 'push'" \
+  "A version check that cannot reach GitHub must not throw away a cycle that is ready to be pushed"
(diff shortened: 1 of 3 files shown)


yegor256:
@yegor256 could you have a look when you get a chance? I bundled #1900 and #1901 here since they are the same defect in two places of the same file, but happy to split them if you would rather review them apart.

0crat:
@yegor256 Thanks for the contribution! You've earned +23 points for this: +24 as a basis; +6 for the 88 hits-of-code that you wrote; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+479](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1919, opened 2026-08-31
Leave the factbase alone in dry mode

### LINKED ISSUE #1902: `dry-run` deletes the `judges-summary` facts it promises not to touch
## What happens

`action.yml` says what `dry-run` promises:

```yaml
  dry-run:
    description: 'Skip all judges, make no changes to the factbase'
```

`entry.sh` changes the factbase anyway. This runs unconditionally, before the
`dry-run` branch is reached:

```bash
${JUDGES} "${gopts[@]}" eval \
    "${fb}" \
    "\$fb.query(\"(eq what 'judges-summary')\").delete!"

if [ "$(printenv "INPUT_DRY-RUN" || echo 'false')" == 'true' ]; then
```

So a dry run deletes every `judges-summary` fact from the factbase it was
told not to touch, and `update` then runs with `--summary=add`, which puts a
fresh one back. The file on disk is not the file the caller had.

This is the same shape as #1603, where `dry-run` did not stop the SQLite
cache download and upload either.

## What should happen

`dry-run: true` should leave the factbase byte for byte as it found it. The
`eval` that deletes the summary belongs inside the branch that already knows
whether this is a real run.

yegor256:
`action.yml` describes `dry-run` as "Skip all judges, make no changes to the factbase", but the `judges eval` that deletes the `judges-summary` facts ran before the branch that knows whether the run is dry. So a dry run emptied the summary of the caller's factbase and `update --summary=add` wrote a fresh one in its place. The deletion now happens only on a real run, and a dry run asks for `--summary=off`, which is what it takes for the file to come out as it went in.

The new entry test seeds a summary fact with `cycles=7`, runs a dry run, and checks both that the fact is still there and that `cksum` of the file has not moved. It was red before the change, finding zero of those facts, and I confirmed by hand that the bytes are now identical before and after.

One thing you may want to decide separately: on a real run that `eval` is already redundant. `judges update` deletes the summary itself when `--summary=add` is given (`log_summary` in `lib/judges/commands/update.rb`), and the `eval` also runs before `pull` replaces the file with the server's copy, so it operates on a factbase that is about to be thrown away. I left it in place rather than remove it, since dropping it is a bigger call than fixing the dry-run promise.

Closes #1902

### DIFF
--- a/entries/keeps-the-summary-in-dry-mode.sh
+++ b/entries/keeps-the-summary-in-dry-mode.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+set -e -o pipefail
+
+SELF=$1
+
+source "${SELF}/makes/setup-test-env.sh"
+source "${SELF}/makes/test-common.sh"
+setup_test_env "${SELF}" name
+
+bundle exec judges eval "${name}.fb" "f = \$fb.insert; f.what = 'judges-summary'; f.cycles = 7"
+before=$(cksum < "${name}.fb")
+
+run_entry_script "${SELF}" success \
+  "GITHUB_WORKSPACE=$(pwd)" \
+  "INPUT_FACTBASE=${name}.fb" \
+  "INPUT_CYCLES=1" \
+  "INPUT_REPOSITORIES=yegor256/factbase" \
+  "INPUT_VERBOSE=false" \
+  "INPUT_TOKEN=something" \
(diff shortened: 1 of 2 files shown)


yegor256:
@yegor256 mind taking a look when you have a moment? If you agree the eval is dead code on a real run too, I can drop it in a follow-up.

0crat:
@yegor256 Thanks for the contribution! You've earned +16 points for this: +24 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+495](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1921, opened 2026-08-31
Stop retiring the facts of a live user in eliminate-ghosts

### LINKED ISSUE #1909: `eliminate-ghosts` marks facts stale for a user it just confirmed alive in `eliminate-ghosts.rb`
The propagation loop at `judges/eliminate-ghosts/eliminate-ghosts.rb:53` takes every user that has at least one fact carrying `stale = 'who'` and marks all of that user's remaining facts stale too. It never consults the `good` set that the scan above it fills at line 41, so a user the judge resolved to a live nick moments earlier, in the same run, still gets the rest of its facts retired.

The two halves of the file contradict each other. The scan at line 16 asks GitHub about user 777, gets a login back, and records it as good; the loop at line 53 then finds one old `stale = 'who'` fact for 777 and spreads that flag across the others. `revive-user` and `who-has-name` exist to undo exactly this — `judges/revive-user/revive-user.rb:24` deletes `stale = 'who'` once the user answers — so the next cycle clears the flags and the cycle after that sets them again. The factbase keeps getting rewritten and the user never settles. `test/judges/test-eliminate-ghosts.rb:test_marks_sibling_facts_stale_for_good_user` asserts the current behavior, so the flip-flop is locked in.

Skipping users already known to be alive fixes it: the inner loop should not run when `good.include?(f.who)`. That test needs to flip with the fix, since it currently demands the wrong outcome.

@yegor256 please take a look.

### LINKED ISSUE #1910: The two staling loops disagree on already-stale facts in `eliminate-ghosts.rb`
The two loops that write `stale = 'who'` in `judges/eliminate-ghosts/eliminate-ghosts.rb` disagree about facts that are already stale for some other reason. The first, at line 48, selects with `(absent stale)`; the second, at line 54, selects with `(not (eq stale 'who'))`. The property is not a boolean — `stale` is written as `'branch'`, `'repository'` and `'issue'` elsewhere, for instance at `judges/fix-missing-branch/fix-missing-branch.rb:60` and `judges/fix-missing-comments/fix-missing-comments.rb:31`.

So a ghost user's fact that already carries `stale = 'repository'` is skipped by the first loop and never learns that its user is gone, which keeps it invisible to `revive-user`, whose query at `judges/revive-user/revive-user.rb:12` looks for `(eq stale 'who')`. The second loop reaches that same fact and appends a value, leaving `stale` holding both `'repository'` and `'who'`. One fact ends up under-marked and another over-marked, from the same run, over the same property.

Both loops should agree, and `(not (eq stale 'who'))` is the predicate that matches what the judge means: mark the user as gone regardless of what else made the fact stale. Line 48 should use it too.

@yegor256 please take a look.

yegor256:
The propagation loop spread `stale='who'` across a user's facts without asking whether the scan right above it had just resolved that user to a live nick. `revive-user` then cleared the flags on the next cycle and this judge set them again, so the factbase kept churning and the user never settled. The loop now skips a user that is in the `good` set. The test that demanded the old outcome asserts the opposite, and it was the only thing holding the flip-flop in place.

The two staling loops also disagreed about facts that were already stale for another reason: the first selected on `(absent stale)`, the second on `(not (eq stale 'who'))`. So a ghost's fact carrying `stale='repository'` was skipped by one loop and appended to by the other, which left it invisible to `revive-user`, whose query looks for `(eq stale 'who')`. Both loops use the same predicate now. `Fbe.delete_one` in `revive-user` removes just the `'who'` value, so a fact holding `['repository', 'who']` is handled properly.

While testing this I hit a third thing, which I have not touched here: in the propagation loop's query, `(unique who)` sits before `(eq stale "who")`, and factbase evaluates the terms of an `and` in order, so the first fact of a user consumes the uniqueness slot even when it is then rejected. The loop therefore does nothing at all unless the user's first stored fact happens to be the stale one — `(and ... (unique who) (eq stale "who"))` returns 0 facts where `(and ... (eq stale "who") (unique who))` returns 1. Reordering would switch that whole loop back on and stale a lot more facts, which is your call rather than mine, so I filed it as #1922.

Closes #1909
Closes #1910

### DIFF
--- a/judges/eliminate-ghosts/eliminate-ghosts.rb
+++ b/judges/eliminate-ghosts/eliminate-ghosts.rb
@@ -45,12 +45,13 @@ Fbe.fb.query('(and (absent stale) (eq where "github") (exists who))').each do |f
 end
 
 bad.each do |u|
-  Fbe.fb.query("(and (absent stale) (eq where 'github') (eq who #{u}))").each do |f|
+  Fbe.fb.query("(and (not (eq stale 'who')) (eq where 'github') (eq who #{u}))").each do |f|
     f.stale = 'who'
   end
 end
 
 Fbe.fb.query('(and (eq where "github") (exists who) (unique who) (eq stale "who"))').each do |f|
+  next if good.include?(f.who)
   Fbe.fb.query("(and (eq who #{f.who}) (not (eq stale 'who')) (eq where 'github'))").each do |ff|
     ff.stale = 'who'
   end
--- a/test/judges/test-eliminate-ghosts.rb
+++ b/test/judges/test-eliminate-ghosts.rb
@@ -109,7 +109,7 @@ class TestEliminateGhosts < Jp::Test
     assert(fb.has?(where: 'github', who: 333))
   end
 
-  def test_marks_sibling_facts_stale_for_good_user
+  def test_keeps_sibling_facts_of_a_live_user
     WebMock.disable_net_connect!
     rate_limit_up
     stub_github('https://api.github.com/user/777', body: { login: 'user777', id: 777, type: 'User' })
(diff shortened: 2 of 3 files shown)


yegor256:
The rake failure was not from this change. `test-quality-of-service.rb` has four tests that load a judge file and call its metric function without setting `$loog`, and they passed only because a neighbouring test had set the global first. On a seed that puts one of them first, the judge hits `$loog.info` on nil. Master failed the same way an hour ago on a different seed, in `test_skips_repo_on_forbidden_workflow_runs`. I pushed a second commit setting `$loog` in those four, and `rake picks` is green here across several seeds.

0crat:
@yegor256 Thanks for the contribution! You've earned +12 points for this: +24 as a basis; -8 for the lack of code review; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+507](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1924, opened 2026-08-31
#1897: keep job ids and run ids in caches of their own

### LINKED ISSUE #1897: `Jp.fetch_workflows` keeps job ids and run ids in one cache, so one can answer for the other
## What happens

`Jp.fetch_workflows` keeps one hash, `wcache`, and puts two different kinds of
key into it, holding two different kinds of value:

```ruby
wcache = {}
(entries[:check_runs] || []).each do |run|
  rid =
    wcache[run[:id]] ||=
      begin
        Fbe.octo.workflow_run_job(repo, run[:id])[:run_id]
      ...
  workflow =
    wcache[rid] ||=
      begin
        Fbe.octo.workflow_run(repo, rid)
      ...
```

`run[:id]` is the id of a check run, and what is stored under it is an
Integer, the id of the workflow run it belongs to. `rid` is that workflow run
id, and what is stored under it is the workflow run itself, a hash. The two
ids come from different sequences at GitHub, and nothing keeps them apart, so
one number can be both.

When it is, the wrong value comes back. A check run whose id equals a
workflow run id already cached gives `rid` a hash instead of an Integer, and
`Fbe.octo.workflow_run(repo, <hash>)` is asked for. The other way round, the
workflow lookup returns an Integer, and `workflow[:event]` on an Integer
raises `NoMethodError`, which crashes the judge for the whole repository, not
just this pull request.

## What should happen

The two lookups should not share a hash. Two hashes, one keyed by job id and
one keyed by run id, cost nothing and cannot answer each other's questions.

morphqdd:
One hash held two kinds of key: a check run id pointing at an Integer, and a workflow run id pointing at a hash. The two come from different sequences at GitHub, so one number can be both, and then the wrong value comes back: `workflow_run` gets asked for a hash, or `workflow[:event]` is called on an Integer and the judge crashes for the whole repository.

Two hashes now, one per kind.

Closes #1897

### DIFF
--- a/lib/pull_request.rb
+++ b/lib/pull_request.rb
@@ -139,11 +139,12 @@ def Jp.fetch_workflows(pr, repo: nil)
     )
     return { succeeded_builds: 0, failed_builds: 0 }
   end
-  wcache = {}
+  jobs = {}
+  runs = {}
   (entries[:check_runs] || []).each do |run|
     next unless run.dig(:app, :slug) == 'github-actions'
     rid =
-      wcache[run[:id]] ||=
+      jobs[run[:id]] ||=
         begin
           Fbe.octo.workflow_run_job(repo, run[:id])[:run_id]
         rescue Octokit::NotFound, Octokit::Deprecated => e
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (7) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+534](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1926, opened 2026-08-31
#1899: take the metrics in a fixed order, not at random

### LINKED ISSUE #1899: `Jp.incremate` shuffles the metric files, so one metric can go uncollected for many runs
## What happens

`Jp.incremate` picks the metric files at random and stops when the budget is
spent:

```ruby
Dir[File.join(dir, "#{prefix}_*.rb")].shuffle.each do |rb|
  ...
  break if Fbe.over?(epoch:, kickoff:)
  ...
  evaluated += 1
  break if max_per_fact && evaluated >= max_per_fact
end
```

So one run evaluates a random subset. The next run shuffles again, from
scratch, and nothing anywhere records that a metric was cut off. A metric
that was skipped has exactly the same chance of being skipped again.

That makes the wait for one metric open ended rather than bounded. With `N`
metric files and `k` of them evaluated per run, the chance of a given one
being missed `m` runs in a row is `(1 - k/N)^m`, which never reaches zero.
`quality-of-service` passes `max_per_fact` from `qos_max_per_fact`, so `k` is
deliberately smaller than `N` there and the effect is not a corner case.

## What should happen

The order should be by staleness rather than random: the metric that has been
missing from the fact the longest goes first. The fact already knows which
metrics it holds, since the loop skips a metric when `fact[n]` is set, so
ordering the missing ones needs no new state.

morphqdd:
The files were shuffled and the loop stopped when the budget ran out, so each run evaluated a random subset and the next run shuffled again from scratch. A metric skipped once had the same chance of being skipped again, and with `qos_max_per_fact` the budget is deliberately smaller than the number of metrics, so the wait for one of them had no bound.

A fixed order gives it one. A metric already in the fact is skipped, so every run starts where the last one was cut off and the missing ones are collected within a known number of runs.

Closes #1899

### DIFF
--- a/lib/incremate.rb
+++ b/lib/incremate.rb
@@ -17,7 +17,7 @@ def Jp.incremate(
   epoch: $epoch || Time.now, kickoff: $kickoff || Time.now
 )
   evaluated = 0
-  Dir[File.join(dir, "#{prefix}_*.rb")].sort.each do |rb|
+  Dir[File.join(dir, "#{prefix}_*.rb")].each do |rb|
     n = File.basename(rb).gsub(/\.rb$/, '')
     if fact[n]
       $loog.debug("#{n} is here: #{fact[n].first}")
--- a/lib/incremate.rb
+++ b/lib/incremate.rb
@@ -17,7 +17,7 @@ def Jp.incremate(
   epoch: $epoch || Time.now, kickoff: $kickoff || Time.now
 )
   evaluated = 0
-  Dir[File.join(dir, "#{prefix}_*.rb")].shuffle.each do |rb|
+  Dir[File.join(dir, "#{prefix}_*.rb")].sort.each do |rb|
     n = File.basename(rb).gsub(/\.rb$/, '')
     if fact[n]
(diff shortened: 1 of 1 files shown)


yegor256:
The `rake` and `make` CI checks are failing due to a RuboCop violation. All 354 tests pass (0 failures, 0 errors), but RuboCop fails at the end:

```
RuboCop failed!
Process completed with exit code 1.
```

Could you run `rubocop` locally and fix the style issues?

---

morphqdd:

yegor256:
Thanks for the contribution!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+708](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1927, opened 2026-08-31
#1906: date the fact by the clock the query filters on

### LINKED ISSUE #1906: `dimensions-of-terrain` filters by `TODAY` but dates its fact by `Time.now`
## What happens

`dimensions-of-terrain` looks for today's fact through `TODAY` and then
writes one dated by the real clock:

```ruby
f = Fbe.fb.query(
  "(and
    (eq what '#{$judge}')
    (gt when (minus (to_time (env 'TODAY' '#{Time.now.utc.iso8601}')) '1 days')))"
).each.first
if f.nil?
  f = Fbe.fb.insert
  f.what = $judge
  f.when = Time.now
end
```

The window is anchored on `TODAY` when it is set, the fact is anchored on
`Time.now` always. With `TODAY` set to a date in the past, which is how the
back-filling runs are driven, the fact this judge inserts carries a `when`
far to the right of the window it was inserted to satisfy. The next run
queries the same window, finds nothing again, and inserts another one. Every
run adds a fact and no run ever finds one.

`Jp.cover_qo` next door reads the same variable and uses it for both sides:

```ruby
v = ENV.fetch('TODAY', nil)
...
n.when = today
n.since = n.when - slice
```

## What should happen

The fact should be dated by the same clock the query filters on. Reading
`TODAY` once and using it both for the window and for `f.when` makes the
judge insert one fact per day, which is what the one-day window says it
intends.

morphqdd:
The window was anchored on `TODAY` when it is set, the fact on `Time.now` always. With `TODAY` set to a past date, which is how the back-filling runs are driven, every run inserted a fact outside the window it was inserted to satisfy, so the next run found nothing and inserted another one.

`TODAY` is now read once and used for both, the way `Jp.cover_qo` next door reads it.

Closes #1906

### DIFF
--- a/judges/dimensions-of-terrain/dimensions-of-terrain.rb
+++ b/judges/dimensions-of-terrain/dimensions-of-terrain.rb
@@ -8,15 +8,24 @@ require 'fbe/octo'
 require 'time'
 require_relative '../../lib/incremate'
 
+today =
+  begin
+    v = ENV.fetch('TODAY', nil)
+    if v.nil? || v.empty?
+      Time.now.utc
+    else
+      Time.parse(v)
+    end
+  end
 f = Fbe.fb.query(
   "(and
     (eq what '#{$judge}')
-    (gt when (minus (to_time (env 'TODAY' '#{Time.now.utc.iso8601}')) '1 days')))"
+    (gt when (minus (to_time '#{today.utc.iso8601}') '1 days')))"
 ).each.first
 if f.nil?
   f = Fbe.fb.insert
   f.what = $judge
-  f.when = Time.now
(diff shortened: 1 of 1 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+554](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1928, opened 2026-08-31
#1905: read the total the search already reports

### LINKED ISSUE #1905: `some_backlog_size` counts items instead of `total_count` and stops at the search ceiling
## What happens

`some_backlog_size` asks the search API a question that already answers
itself, and then answers it again by hand:

```ruby
found = Jp.qosearch(
  "repo:#{repo} type:issue created:*..#{date.iso8601[0..9]} (closed:>=#{date.iso8601[0..9]} OR state:open)",
  advanced_search: true
)
return {} if found.nil?
found[:items].each do |item|
  count += 1 if item[:closed_at].nil? || item[:closed_at].utc.to_date >= date
end
issues << count
```

The `(closed:>=date OR state:open)` in the query and the
`closed_at.nil? || closed_at >= date` in the block are the same condition, so
the block accepts every item the query returned and the count is just
`found[:items].size`.

Counting items instead of reading `total_count` costs two things. Every page
has to be fetched, where the total was already in the first response. And the
search API stops handing out results after a thousand, so a backlog above
that is reported as a thousand with nothing saying it was cut.

Its neighbour asks the same kind of question the cheap way:

```ruby
found = Jp.qosearch(q)
return {} if found.nil?
pulls << found[:total_count]
```

## What should happen

`some_backlog_size` should read `found[:total_count]`, the way
`some_merged_pulls` does. One page is then enough, the thousand-result
ceiling stops applying, and the duplicated condition goes with it.

morphqdd:
The `(closed:>=date OR state:open)` in the query and the `closed_at.nil? || closed_at >= date` in the block were the same condition, so the block accepted every item the query returned and the count was `found[:items].size`.

That cost two things: every page had to be fetched, though the total was already in the first response, and the search API stops handing out results after a thousand, so a larger backlog was reported as a thousand with nothing saying it was cut.

`some_merged_pulls` next door reads `total_count`. This one now does too.

Closes #1905

### DIFF
--- a/judges/quality-of-service/some_backlog_size.rb
+++ b/judges/quality-of-service/some_backlog_size.rb
@@ -13,16 +13,12 @@ def some_backlog_size(fact)
   Fbe.unmask_repos do |repo|
     (fact.since.utc.to_date..fact.when.utc.to_date).last(7).each do |date|
       return {} if Fbe.octo.off_quota?(resource: :search)
-      count = 0
       found = Jp.qosearch(
         "repo:#{repo} type:issue created:*..#{date.iso8601[0..9]} (closed:>=#{date.iso8601[0..9]} OR state:open)",
         advanced_search: true
       )
       return {} if found.nil?
-      found[:items].each do |item|
-        count += 1 if item[:closed_at].nil? || item[:closed_at].utc.to_date >= date
-      end
-      issues << count
+      issues << found[:total_count]
     end
   end
   { some_backlog_size: issues }


0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (6) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+574](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1929, opened 2026-08-31
#1903: rebuild the image when anything it copies in changes

### LINKED ISSUE #1903: `Makefile` rebuilds the image without `lib/patches/*.rb` and the judge test packs
## What happens

The Docker image is rebuilt when one of its prerequisites changes:

```make
target/docker-image.txt: Makefile Dockerfile entry.sh Gemfile Gemfile.lock $(wildcard lib/*.rb) $(wildcard judges/*/*.rb)
```

Two kinds of file that go into the image are missing from that list.

`lib/*.rb` does not match a subdirectory, and `lib/patches/` holds two files
that the `Dockerfile` copies in with `COPY lib /action/lib`:
`lib/patches/unmask_repos.rb` and `lib/patches/fake_octokit.rb`.

`judges/*/*.yml` is not there at all, and those are the test packs that
`make test` runs:

```make
docker run --rm --entrypoint '/bin/bash' "$${img}" -c 'bundle exec judges test --disable live --lib /action/lib /action/judges'
```

So editing a patch or a test pack and running `make test` tests the image
built before the edit, and the run passes or fails on code that is no longer
in the working tree.

## What should happen

The prerequisites should cover what the image contains. `$(shell find lib
judges -type f)`, or adding `$(wildcard lib/**/*.rb)` and
`$(wildcard judges/*/*.yml)`, makes the stamp file older than any file that
went into the image.

morphqdd:
`lib/*.rb` does not match a subdirectory, so `lib/patches/unmask_repos.rb` and `lib/patches/fake_octokit.rb` were left out, and `judges/*/*.yml`, the test packs `make test` runs, were not listed at all. Editing either and running `make test` tested the image built before the edit.

`find lib judges -type f` covers what the `Dockerfile` copies in with `COPY lib` and `COPY judges`.

Closes #1903

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ verify:
 	e2=$$(cat target/entry.exit)
 	test "$${e2}" = "0"
 
-target/docker-image.txt: Makefile Dockerfile entry.sh Gemfile Gemfile.lock $(wildcard lib/*.rb) $(wildcard judges/*/*.rb)
+target/docker-image.txt: Makefile Dockerfile entry.sh Gemfile Gemfile.lock $(shell find lib judges -type f)
 	mkdir -p "$$(dirname $@)"
 	docker build -t judges-action "$$(pwd)"
 	docker build -t judges-action -q "$$(pwd)" > "$@"


0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+594](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1930, opened 2026-08-31
Look up a forbidden user once per run in eliminate-ghosts

### LINKED ISSUE #1911: Forbidden users are looked up once per fact in `eliminate-ghosts.rb`
The scan in `judges/eliminate-ghosts/eliminate-ghosts.rb:17` skips a fact when its user is already in `good` or `bad`, so each user costs one lookup per run. The `Octokit::Forbidden` branch at line 26 breaks that guarantee: it logs, throws, and adds the user to neither set. Every later fact carrying the same `who` walks back into `Jp.nick_of` and asks about a user that just answered 403.

A user the app cannot see tends to own many facts, and the judge retries all of them, one lookup each, on every cycle. `judges/who-is-alive/who-is-alive.rb:16` handles the same situation by pushing onto `seen` before the lookup rather than after it, so a forbidden user is asked about once and skipped thereafter. Whether these repeats reach GitHub or stop at the SQLite cache added in #1241 depends on whether error responses are cached, but either way the file keeps a set precisely so it does not have to ask twice, and this path defeats it. The 403 must stay non-permanent — #1518 fixed the bug where it was treated as a ghost — so the user cannot simply go into `bad`.

A third set for forbidden users, checked alongside `good` and `bad` at line 17 and filled in the `Octokit::Forbidden` branch, keeps the retry-next-cycle semantics while asking only once per run.

@yegor256 please take a look.

Thayorns:
The scan in eliminate-ghosts.rb skips a fact when its user is already known good or bad, but the Octokit::Forbidden branch never recorded the user anywhere, so every later fact with the same who walked back into Jp.nick_of and asked about a user that had just answered 403.

This adds a third set, forbidden, checked alongside good and bad, and filled in when a lookup comes back 403. The 403 stays non-permanent: forbidden users are still retried on the next cycle, they just aren't asked about more than once within the same run.

Closes #1911

### DIFF
--- a/judges/eliminate-ghosts/eliminate-ghosts.rb
+++ b/judges/eliminate-ghosts/eliminate-ghosts.rb
@@ -12,9 +12,10 @@ require_relative '../../lib/nick_of'
 
 good = Set.new
 bad = Set.new
+forbidden = Set.new
 
 Fbe.fb.query('(and (absent stale) (eq where "github") (exists who))').each do |f|
-  next if good.include?(f.who) || bad.include?(f.who)
+  next if good.include?(f.who) || bad.include?(f.who) || forbidden.include?(f.who)
   next if Fbe.octo.off_quota?
   elapsed($loog, level: Logger::INFO) do
     nick =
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 please take a look, happy to clarify anything.

0crat:
@thayorns Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (19) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+70](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1778, opened 2026-07-03
#1720: wrap user deletion and stale marking in atomic transaction

VasilevNStas:
Wraps the `delete!` and stale-marking operations in `who-is-alive.rb` inside an `Fbe.fb.txn` block, ensuring both happen atomically or not at all.

Also fixes a missing closing parenthesis in the factbase query string that caused `Factbase::Syntax::Broken`.

Closes #1720

Issue #1720, opened 2026-06-30
judges/who-is-alive.rb query & delete outside transaction


sdghsdkjlas27-dotcom:
DESCRIPTION:
judges/who-is-alive/who-is-alive.rb does `Fbe.fb.query(...).each do |n| n.stale = 'who' end` then a second `Fbe.fb.query(...).each`. Both outside Fbe.fb.txn. Concurrent judges can mutate the set.

REPRODUCTION:
1. Race scenario.

IMPACT:
Inconsistent stale state.

RELATED FILES:
judges/who-is-alive/who-is-alive.rb

PR #1931, opened 2026-08-31
fix(#1912): wrap eliminate-ghosts stale marking in a transaction

### LINKED ISSUE #1912: Stale flags are written outside a transaction in `eliminate-ghosts.rb`
The two staling loops at `judges/eliminate-ghosts/eliminate-ghosts.rb:47` and `:53` write `stale = 'who'` fact by fact, outside any transaction. The judge holds its findings in the `bad` set until the scan above finishes, then applies them in a second pass, so there is a long window in which some of a ghost's facts carry the flag and the rest do not.

Anything that ends the run inside those loops leaves that split behind: the quota check at line 18 cannot help once the scan is over, and an Octokit error the `case` at line 24 does not name is re-raised at line 34. Nothing on the next run notices the half-applied state, because the loop at line 48 selects on `(absent stale)` and simply picks up wherever it stopped, while consumers reading `(absent stale)` see a user that is partly retired and partly not. This is the same defect reported for the sibling judge in #1720, fixed there by wrapping the mutation in `Fbe.fb.txn` at `judges/who-is-alive/who-is-alive.rb:52`; `eliminate-ghosts` was left alone.

Wrapping both loops in `Fbe.fb.txn`, the way `who-is-alive` does, makes the pass all-or-nothing.

@yegor256 please take a look.

Thayorns:
Closes #1912

The two staling loops in `eliminate-ghosts.rb` wrote the `stale = 'who'` fact one fact at a time, outside any transaction. Anything that ended the run partway through (a killed process, an unhandled error) could leave some of a ghost's facts flagged stale and the rest not, and nothing on a later run noticed the split state.

This wraps both loops in `Fbe.fb.txn`, the same way `who-is-alive` was fixed for the identical problem in #1720 (PR #1778). Each pass — marking the confirmed-bad users, then propagating `stale` to their sibling facts — now commits or rolls back as a unit.

No new test is included, matching the precedent PR #1778 for that identical fix: the failure mode is process interruption mid-loop, which isn't reachable through a code-level exception in this test harness.

### DIFF
--- a/judges/eliminate-ghosts/eliminate-ghosts.rb
+++ b/judges/eliminate-ghosts/eliminate-ghosts.rb
@@ -5,6 +5,7 @@
 
 require 'elapsed'
 require 'fbe/consider'
+require 'fbe/fb'
 require 'fbe/octo'
 require 'logger'
 require 'octokit'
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 happy to clarify anything about the transaction split if useful.

yegor256:
This PR has a merge conflict with the base branch. Please rebase or merge master into your branch to resolve the conflict before this can be merged.

---

yegor256:
Note: PR #1932 was just opened by @yegor256 and also addresses issue #1912 (as well as #1911) with a more extensive fix. It includes the same transaction wrapping you added here, plus the forbidden-user deduplication. Holding off on merging this PR to avoid a conflict — once #1932's CI completes, one of the two will need to be closed.

---

PR #1932, opened 2026-08-31
Write the stale flags of eliminate-ghosts in one transaction

### LINKED ISSUE #1912: Stale flags are written outside a transaction in `eliminate-ghosts.rb`

yegor256:
The two staling loops in `eliminate-ghosts.rb` wrote fact by fact, outside any transaction, so anything that ended the run between them left a ghost half retired. Nothing on the next run noticed: the first loop just picks up wherever it stopped, while consumers reading `(absent stale)` see a user that is partly retired and partly not. Both loops run inside one `Fbe.fb.txn` now, the way `who-is-alive` has since #1720.

I should be straight about the test: it asserts that the factbase logs a `Txn #` line, which shows the writes went through a transaction but does not prove atomicity under a mid-loop failure. Injecting such a failure needs a fake factbase, and #1720 was closed without one, so I matched that.

This branch was rebased onto master after #1930 landed, so the `forbidden` set half of the original diff is gone — master already has it, with the same design and an equivalent test. 357 tests green, rubocop clean.

Closes #1912

### DIFF
--- a/judges/eliminate-ghosts/eliminate-ghosts.rb
+++ b/judges/eliminate-ghosts/eliminate-ghosts.rb
@@ -46,16 +46,17 @@ Fbe.fb.query('(and (absent stale) (eq where "github") (exists who))').each do |f
   end
 end
 
-bad.each do |u|
-  Fbe.fb.query("(and (not (eq stale 'who')) (eq where 'github') (eq who #{u}))").each do |f|
-    f.stale = 'who'
+Fbe.fb.txn do |fbt|
+  bad.each do |u|
+    fbt.query("(and (not (eq stale 'who')) (eq where 'github') (eq who #{u}))").each do |f|
+      f.stale = 'who'
+    end
   end
-end
-
-Fbe.fb.query('(and (eq where "github") (exists who) (unique who) (eq stale "who"))').each do |f|
-  next if good.include?(f.who)
-  Fbe.fb.query("(and (eq who #{f.who}) (not (eq stale 'who')) (eq where 'github'))").each do |ff|
-    ff.stale = 'who'
+  fbt.query('(and (eq where "github") (exists who) (unique who) (eq stale "who"))').each do |f|
+    next if good.include?(f.who)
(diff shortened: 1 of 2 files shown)


yegor256:
@yegor256 could you take a look when you have a moment? Say the word if you would rather have a proper atomicity test than the log assertion I settled for.

yegor256:
This PR has a merge conflict (`dirty` state). Could you please rebase it onto the current `master` branch and resolve the conflicts?

---

yegor256:
Rebased onto master and dropped the #1911 half, since #1930 already landed it with the same design. What is left here is only the transaction.

0crat:
@yegor256 Thanks for the contribution! You've earned +16 points for this: +24 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+523](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1864, opened 2026-07-24
`Fbe.octo.off_quota?` has no rescue, crashes entire judge run on transient GitHub error


morphqdd:
## What happens

`Fbe.octo.off_quota?` (defined in the `fbe` gem, `lib/fbe/octo.rb`, pinned at `fbe (0.49.1)` in this repo's `Gemfile.lock`) calls `@origin.rate_limit!`, a live `GET /rate_limit` HTTP request to the GitHub API, with no rescue around it:

```ruby
def off_quota?(threshold: nil, resource: :core)
  threshold ||= resource == :search ? 5 : 50
  label = resource == :search ? 'GitHub Search API' : 'GitHub API'
  rate = @origin.rate_limit!
  ...
```

Every call site in `judges-action` treats `off_quota?` as a safe, free boolean check, with no protection around it:

- `judges/quality-of-service/some_backlog_size.rb:11,15`
- `judges/quality-of-service/some_merged_pulls.rb:14,19`
- `judges/quality-of-service/some_issue_lifetime.rb:15`
- `judges/quality-of-service/some_pull_hoc_size.rb:14`
- `judges/quality-of-service/some_review_time.rb:17`
- `judges/quality-of-service/some_triage_time.rb:16`
- `judges/find-all-issues/find-all-issues.rb:69`
- `judges/eliminate-ghosts/eliminate-ghosts.rb:18`
- `judges/erase-repository/erase-repository.rb:15`
- `lib/qos_search.rb:25` (used transitively by more judges via `Jp.qosearch`)

If GitHub's `/rate_limit` endpoint returns a transient error (a 5xx, a timeout, or `Octokit::Unauthorized` from an expired token, or `Octokit::TooManyRequests`), the exception propagates out of `off_quota?` uncaught.

`Fbe.consider`/`Fbe.conclude` (`fbe/lib/fbe/conclude.rb`, the `roll` method) only rescues `Fbe::OffQuota`, nothing else. `Fbe.iterate` (`fbe/lib/fbe/iterate.rb`) only catches `StandardError` to re-raise it with more context, it does not swallow it. So an unrescued error from `off_quota?` crashes the entire judge run for all remaining repos and facts in that cycle, not just the one check that failed.

No test currently stubs a failing `/rate_limit` response for any of these judges; existing tests (`test/judges/test-eliminate-ghosts.rb`, `test/judges/test-find-all-issues.rb`) only stub the successful 200 path.

## What should happen

[...]

Issue #651, opened 2025-05-27
Potential overlooking of merged pull requests due to unreliable github-events


yegor256:
Currently, we expect `github-events` to create `pull-was-closed` and `pull-was-merged` facts. However, it's not a reliable expectation. Some events may not be delivered from GitHub for some reasons. Let's create a new judge, that will go through all `pull-was-opened` facts that don't have neither `pull-was-closed` nor `pull-was-merged` corresponding facts, and check the status of PR in GitHub API.

To reduce the amount of traffic, we can add an attribute to the `pull-was-opened` fact: `watched` with Time. We should only "watch" once every 24 hours. Once we find out that the PR was merged or closed, we should delete `watched` attribute (with the help of `Fbe.delete`).

We can call this new judge: `pull-was-merged` (similar to the `what` attribute).

Yegorov:
@yegor256 yes of course i can take it.

PR #1933, opened 2026-08-31
Stop unmask_repos.rb from copying the upstream fbe method

### LINKED ISSUE #1904: `lib/patches/unmask_repos.rb` copies the whole upstream method, so gem fixes never arrive
## What happens

`lib/patches/unmask_repos.rb` does not patch `Fbe.unmask_repos`, it replaces
it with a copy:

```ruby
module Fbe
  class << self
    def unmask_repos(
      options: $options, global: $global, loog: $loog, epoch: $epoch || Time.now, kickoff: $kickoff || Time.now,
      quota_aware: true, lifetime_aware: true, timeout_aware: true
    )
      raise(Fbe::Error, 'Repositories mask is not specified') unless options.repositories
      ...
```

Every line of the upstream method is repeated here, mask parsing, the
exclusion pass, the archived filter, the shuffle, the yielding loop, with
rescues added in two places. The upstream body is never called.

That means a fix made in the gem does not arrive. Whatever `zerocracy/fbe`
changes inside `unmask_repos`, this file keeps answering with the version
that was copied, and nothing reports the divergence: `require
'fbe/unmask_repos'` at the top succeeds, and the redefinition is silent.

The two have already drifted. The copy carries rescues around the archived
check and an extra warm-up loop over the repositories that upstream does not
have, and upstream keeps evolving on its own.

## What should happen

The patch should add what is missing and delegate the rest, so that the
upstream body stays the one that runs. If the rescues cannot be expressed as
a wrapper, they belong in the gem, where they are written once, the way #1864
was closed here with exactly that reasoning.

Thayorns:
lib/patches/unmask_repos.rb redefined Fbe.unmask_repos as a full copy of the fbe gem's method instead of patching it, so any fix the gem made to unmask_repos never reached this repo; the copy had already drifted, carrying rescues around organization_repositories/repositories and around the archived-repository check that upstream doesn't have.

Those rescues sit inside the mask loop and the archived-repository loop of the upstream method, so they can't be added back as a thin wrapper around the call without duplicating the body again. Per the same reasoning #1864 was closed with, that behavior belongs in the gem, not copied here, so lib/patches/unmask_repos.rb is now just the require and Fbe.unmask_repos runs the upstream implementation directly. Filed zerocracy/fbe#651 proposing the gem add the missing rescues.

test/lib/test_unmask_repos.rb drops the three tests that only exercised the removed local rescues and gains one asserting Fbe.unmask_repos resolves to the gem's source file rather than being redefined here; the two tests covering behavior the gem already provides (falling back to a user's own repos, fetching a repo only once) stay as they are.

Closes #1904


Thayorns:
@yegor256 happy to clarify or adjust anything here.

yegor256:
CI is green, but holding off: this deletes the local unmask_repos rescues to delegate to the fbe gem's version, per #1904. The gem's Fbe.unmask_repos (fbe master, lib/fbe/unmask_repos.rb) still has no rescues around organization_repositories/repository — that's #651, still open and unfixed upstream. Merging this now would drop the existing rescue coverage here before the replacement lands in the gem. Needs #651 fixed and released first.

---

Thayorns:
@yegor256 re-review this one because https://github.com/zerocracy/fbe/issues/651 merged

Thayorns:
@yegor256 fixed here, take a look please

Thayorns:
@yegor256 take a look please

Thayorns:
@yegor256 please review

Thayorns:

Thayorns:

Thayorns:
@yegor256 ping

Thayorns:
@yegor256 ping

PR #1954, opened 2026-09-01
fix(#1938): assert fact count in test_adds_review_comments_to_bare_facts

### LINKED ISSUE #1938: `test_adds_review_comments_to_bare_facts` passes on an empty result
`test_adds_review_comments_to_bare_facts` in `test/judges/test-add-review-comments.rb:47` stays green when it finds nothing. It collects the matching facts at line 62 and then asserts inside `facts.each` at lines 63-65, with no assertion on how many facts came back. An empty `facts` runs the block zero times and the test passes having checked nothing.

This is the only test covering the multi-fact path, and the emptiness it cannot detect is reachable: the judge's query at `judges/add-review-comments/add-review-comments.rb:13` is what decides whether a `pull-was-reviewed` fact is picked up at all, so a broken `or` there, or a change to the `where` term at line 20, would leave `facts` empty and this test would report success. The three other tests in the file dereference `facts.first` and would at least error out; this one would not.

Asserting the size before the loop — `assert_equal(pulls.size, facts.size)` — closes the hole and keeps the per-fact check where it is.

@yegor256

Thayorns:
## Summary

`test_adds_review_comments_to_bare_facts` in `test/judges/test-add-review-comments.rb` collects the matching facts and then only asserts inside a `facts.each` loop, with no check on how many facts came back. If `facts` is empty, the loop runs zero times and the test passes having verified nothing.

This is the only test covering the multi-fact path. The judge's query in `judges/add-review-comments/add-review-comments.rb` decides whether a `pull-was-reviewed` fact is picked up at all, so a broken `or` there, or a change to the `where` term, would leave `facts` empty and this test would still report success.

Fix: assert `facts.size == pulls.size` before the per-fact loop, so an empty result now fails instead of passing silently.

Closes #1938

### DIFF
--- a/test/judges/test-add-review-comments.rb
+++ b/test/judges/test-add-review-comments.rb
@@ -60,6 +60,7 @@ class TestAddReviewComments < Jp::Test
     end
     load_it('add-review-comments', fb)
     facts = fb.query("(eq what \"#{what}\")").each.to_a
+    assert_equal(pulls.size, facts.size)
     facts.each do |f|
       assert_equal(pulls.find { |pl| pl[:id] == f.issue }[:comments], f.review_comments)
     end
--- a/test/judges/test-add-review-comments.rb
+++ b/test/judges/test-add-review-comments.rb
@@ -128,6 +128,60 @@ class TestAddReviewComments < Jp::Test
     assert_nil(f['stale'], '403 is transient — fact must NOT be marked stale; pull lookup will retry next cycle')
   end
 
+  def test_sets_review_comments_for_merged_pulls
+    WebMock.disable_net_connect!
+    pl = { id: 96, comments: 5 }
+    repo = 42
+    stub(repo, pl)
+    what = 'pull-was-merged'
+    fb = Factbase.new
+    fact = fb.insert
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 take a look please

yegor256:
Thanks for this — good, focused test additions that assert on fact count instead of passing vacuously. CI is green.

---

yegor256:
This PR has merge conflicts with `master` right now, so it isn't mergeable as-is. The fix itself (asserting `facts.size == pulls.size` before the per-fact loop in `test_adds_review_comments_to_bare_facts`, so an empty result fails instead of passing silently) looks correct and matches #1938. Requesting @yegor256 to take a look after a rebase.

---

Thayorns:
@yegor256 fixed conflicts here

Thayorns:

Thayorns:
@yegor256 take a look please

Thayorns:
@yegor256 please review

Thayorns:

Thayorns:

Issue #1948, opened 2026-09-01
A lost pull is never buried in the tombstone by `Jp.issue_was_lost`


Thayorns:
`Jp.issue_was_lost` buries the issue in the tombstone on `lib/issue_was_lost.rb:34`, but line 20 returns before it whenever the query on lines 12-19 matched anything. That query looks for facts with the same `where`, `repository` and `issue` that carry no `stale` and no `tombstone`.

Every caller passes the coordinates of a fact that satisfies exactly that. `judges/code-was-reviewed/code-was-reviewed.rb:52` and `:74` hand over `f.where`, `f.repository` and `f.issue`, and `f` reached the block through a query that already demands `(absent stale)` and `(absent tombstone)`, lines 22-23. So `f` matches itself, `stale` is at least one, and the burial never happens. `test/lib/test_issue_was_lost.rb:13-27` pins this path down and checks only that no new fact appears; the tombstone is asserted in `test_buries_issue_in_the_tombstone` at line 50, which starts from an empty factbase and so takes the other branch.

The consequence is that a pull deleted on GitHub is marked stale but never buried, so the `(absent tombstone)` guards elsewhere keep letting other judges fetch it. Burying the issue before the early return, rather than after it, closes the gap.

@yegor256 please take a look

Thayorns:
@yegor256 the fix for this is covered by PR https://github.com/zerocracy/judges-action/pull/1961 (bundled together with the #1944/#1945/#1947 fixes on the same branch).

0crat:
@thayorns Thanks for reporting a new bug! You've earned +18 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+268](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1966, opened 2026-09-01
fix(#1947): log stale count before the early return

### LINKED ISSUE #1947: The stale count logged by `Jp.issue_was_lost` is always zero
`lib/issue_was_lost.rb:35` logs `"The issue #{issue} was marked as lost, #{stale} facts marked as stale"`. Line 20 of the same method is `return if stale.positive?`, so the only way execution reaches line 35 is with `stale` equal to zero.

The line therefore always reports zero, whatever the method actually did. When facts really were marked stale by the `each` on line 19, the method returns silently at line 20 and nothing is logged at all, which is the case that would have been worth a line in the log.

Reporting the count where it is known, next to the early return, tells the truth: log the number before returning on line 20, and drop the interpolation from line 35 since it can only ever be zero there.

@yegor256 please take a look

Thayorns:
## Summary
- `Jp.issue_was_lost` returned at line 20 before reaching the log statement on line 35 whenever it actually marked prior facts stale — the only code path that reaches line 35 has `stale == 0` by construction, so the logged count was always zero.
- Moves the stale-count log to sit next to the early return, where the count is actually known and non-zero, and drops the now-dead `stale` interpolation from the later log line.

Closes #1947

### DIFF
--- a/lib/issue_was_lost.rb
+++ b/lib/issue_was_lost.rb
@@ -17,7 +17,10 @@ def Jp.issue_was_lost(where, repository, issue)
       (absent stale)
       (absent tombstone))"
     ).each { |f| f.stale = 'issue' }
-  return if stale.positive?
+  if stale.positive?
+    $loog.info("The issue ##{issue} was marked as lost, #{stale} facts marked as stale")
+    return
+  end
   f =
     Fbe.if_absent do |n|
       n.issue = issue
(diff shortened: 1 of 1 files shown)


yegor256:
This PR now has merge conflicts with `master` — #1971 (bury-before-early-return for #1948) merged first and restructured the same `return if stale.positive?` block in `lib/issue_was_lost.rb` that this PR also rewrites. The fix itself (logging the stale count where it's actually known, next to the early return) still looks correct and worth landing, but it needs a rebase onto current master to resolve the overlap. Requesting @yegor256 to take a look after a rebase.

---

Thayorns:
@yegor256 fixed conflicts here

Thayorns:

Thayorns:
@yegor256 take a look please

Thayorns:
@yegor256 please review

Thayorns:

Thayorns:

PR #1967, opened 2026-09-01
fix(#1946): don't treat two deleted accounts as the same reviewer

### LINKED ISSUE #1946: Two deleted accounts compare equal in the self-review guard of `code-was-reviewed.rb`
The self-review guard at `judges/code-was-reviewed/code-was-reviewed.rb:94` reads `next if review.dig(:user, :id) == pr.dig(:user, :id)`. Both sides return nil when the account behind them has been deleted, and `nil == nil` is true.

So a pull opened by a deleted account and reviewed by another deleted account looks, to this line, like a person reviewing their own code, and the review is dropped. The two are unrelated people; the only thing they have in common is that GitHub no longer names either of them.

Comparing the ids only once both are known fixes it, for instance by skipping the review when its own id is nil and comparing after that. That check pairs naturally with the nil guard the same line needs anyway.

@yegor256 please take a look

Thayorns:
## Summary
- The self-review guard in `judges/code-was-reviewed/code-was-reviewed.rb` compared `review.dig(:user, :id) == pr.dig(:user, :id)` directly, so a PR opened by a deleted account and reviewed by a *different* deleted account (both ids `nil`) looked like the author reviewing their own code, and the review was silently dropped.
- Now a review is skipped when its own reviewer id is `nil` (we have nothing to attribute it to), and the author-equality comparison only runs once the reviewer id is known.

Fixes #1946

@yegor256 please take a look

### DIFF
--- a/judges/code-was-reviewed/code-was-reviewed.rb
+++ b/judges/code-was-reviewed/code-was-reviewed.rb
@@ -91,7 +91,9 @@ Fbe.consider(
   f.reviews = reviews.count
   count = nil
   reviews.each do |review|
-    next if review.dig(:user, :id) == pr.dig(:user, :id)
+    reviewer = review.dig(:user, :id)
+    next if reviewer.nil?
+    next if reviewer == pr.dig(:user, :id)
     Fbe.fb.txn do |fbt|
       n =
         Fbe.if_absent(fb: fbt) do |nn|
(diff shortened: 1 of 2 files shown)


yegor256:
Merged, thanks. A review is now skipped when its own reviewer id is `nil` before the author-equality check runs, so two different deleted accounts (both `nil`) no longer look like a self-review.

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+772](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1971, opened 2026-09-01
#1948: bury the issue before the early return

### LINKED ISSUE #1948: A lost pull is never buried in the tombstone by `Jp.issue_was_lost`

morphqdd:
The burial sat after `return if stale.positive?`, and every caller hands over
the coordinates of a fact that the query above matches, so `stale` was at
least one and the tombstone never saw the issue. A pull deleted on GitHub was
marked stale and then fetched again by every judge guarded by
`(absent tombstone)`.

The burial happens before that return now. The new test starts from a
factbase that already holds a fact for the issue, which is the path the old
tests never covered; it fails on master as it is.

Closes #1948

### DIFF
--- a/lib/issue_was_lost.rb
+++ b/lib/issue_was_lost.rb
@@ -17,6 +17,7 @@ def Jp.issue_was_lost(where, repository, issue)
       (absent stale)
       (absent tombstone))"
     ).each { |f| f.stale = 'issue' }
+  Fbe::Tombstone.new.bury!(where, repository, issue)
   return if stale.positive?
   f =
     Fbe.if_absent do |n|
@@ -31,6 +32,5 @@ def Jp.issue_was_lost(where, repository, issue)
   end
   f.stale = 'issue'
   f.when = Time.now
-  Fbe::Tombstone.new.bury!(where, repository, issue)
   $loog.info("The issue #{issue} was marked as lost, #{stale} facts marked as stale")
 end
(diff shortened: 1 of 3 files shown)


yegor256:
Merged, thanks. Moves `Fbe::Tombstone.new.bury!` ahead of the `return if stale.positive?`, so a pull whose fact was already marked stale by a prior fact actually gets buried instead of skipping burial entirely — closes the gap that let `(absent tombstone)`-guarded judges keep re-fetching a deleted pull.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1234](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1972, opened 2026-09-01
Remove dead stubs from test_skips_a_pull_that_was_already_checked

### LINKED ISSUE #1953: Dead stubs in `test_skips_a_pull_that_was_already_checked`
`test_skips_a_pull_that_was_already_checked` in `test/judges/test-code-was-reviewed.rb:123` sets up three stubs, but two of them cannot ever be requested. The fact inserted on line 136 carries `reviews: 0`, and the judge's query demands `(absent reviews)` at `judges/code-was-reviewed/code-was-reviewed.rb:21`, so the block on line 32 never runs for it and no call to `repo_name_by_id` or `pull_request` is made.

That leaves the stubs on lines 126 and 127-133, for `repositories/42` and `pulls/45`, as dead fixture. They read as though the test were exercising the lookup path, when the only assertion, `assert_not_requested` on line 138, is about the reviews endpoint never being touched. The next person to change the query will keep them alive for no reason.

Deleting both stubs leaves the test saying exactly what its name promises.

@yegor256 please take a look

Thayorns:
The fact set up in this test carries reviews: 0, and the judge's query requires (absent reviews), so the block that would call repo_name_by_id and pull_request never runs for it. That left the stubs for repositories/42 and pulls/45 as dead fixture — they suggested the test exercised the lookup path when the only assertion is about the reviews endpoint never being hit.

Removed both unused stubs so the test only sets up what it actually exercises. Also added a sibling test, test_skips_a_pull_that_is_already_stale, covering the same skip behavior for a pull already marked stale, since the judge's query guards on absent stale the same way it guards on absent reviews.

Closes #1953

@yegor256 please take a look

### DIFF
--- a/test/judges/test-code-was-reviewed.rb
+++ b/test/judges/test-code-was-reviewed.rb
@@ -123,19 +123,21 @@ class TestCodeWasReviewed < Jp::Test
   def test_skips_a_pull_that_was_already_checked
     WebMock.disable_net_connect!
     rate_limit_up
-    stub_github('https://api.github.com/repositories/42', body: { id: 42, full_name: 'foo/foo' })
-    stub_github(
-      'https://api.github.com/repos/foo/foo/pulls/45',
-      body: {
-        id: 45, number: 45, user: { id: 421, login: 'user' },
-        created_at: Time.parse('2025-09-02 17:05:30 UTC'), additions: 2, deletions: 8
-      }
-    )
     stub = stub_github('https://api.github.com/repos/foo/foo/pulls/45/reviews?per_page=100', body: [])
     fb = Factbase.new
     fb.with(_id: 1, what: 'pull-was-merged', repository: 42, issue: 45, where: 'github', reviews: 0)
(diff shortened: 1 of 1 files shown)


yegor256:
Merged, thanks. Removes the two dead `repositories/42`/`pulls/45` stubs from `test_skips_a_pull_that_was_already_checked` (unreachable given `reviews: 0` never satisfies `(absent reviews)`), and adds `test_skips_a_pull_that_is_already_stale` covering the equivalent `(absent stale)` guard.

---

0crat:
@thayorns Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+840](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1973, opened 2026-09-01
#1922: ask for uniqueness after the fact is chosen

### LINKED ISSUE #1922: The `(unique who)` term disables the propagation loop in `eliminate-ghosts.rb`
The propagation loop at `judges/eliminate-ghosts/eliminate-ghosts.rb:53` selects the users to propagate `stale = 'who'` from:

```ruby
Fbe.fb.query('(and (eq where "github") (exists who) (unique who) (eq stale "who"))').each do |f|
```

`(unique who)` sits before `(eq stale "who")`. Factbase evaluates the operands of an `and` left to right, and `unique` records the value it saw even when a later operand rejects the fact, so the first stored fact of a user consumes that user's uniqueness slot. When that first fact is not the stale one, the query returns nothing and the loop does no work at all, silently.

With two facts for `who=888`, the first carrying `stale='repository'` and the second `stale='who'`, `(and (eq where "github") (exists who) (unique who) (eq stale "who"))` matches 0 facts, while the same query with the two terms swapped matches 1. So whether a ghost's remaining facts get retired depends on the order in which its facts happen to sit in the factbase, which nothing in the judge controls.

Moving `(unique who)` to the end of the `and`, after `(eq stale "who")`, is the fix. Worth deciding on purpose rather than in passing, because it switches a loop back on that currently does nothing in many factbases, and more facts will end up stale as a result. `judges/fix-wrong-closures/fix-wrong-closures.rb:21` has the same ordering, with `(unique issue)` ahead of three per-fact `absent` terms, though the stakes there are lower.

morphqdd:
`(unique who)` sat before `(eq stale "who")`, and `unique` remembers the value
it saw even when a later operand rejects the fact, so the first stored fact of
a user took that user's slot. When that fact was not the stale one, the query
matched nothing and the loop did no work.

With two facts for one user, the first `stale='repository'` and the second
`stale='who'`:

```
(and (eq where "github") (exists who) (unique who) (eq stale "who"))  -> []
(and (eq where "github") (exists who) (eq stale "who") (unique who))  -> [8]
```

and running the loop body over the same two facts leaves the first one as
`["repository"]` with the old order and `["repository", "who"]` with the new
one.

The pack gained that pair, under a user id the fake never looks up, so only
the propagation loop touches them.

`fix-wrong-closures.rb:21` has the same shape with `(unique issue)`. Left
alone here, one issue per pull request.

Closes #1922

### DIFF
--- a/judges/eliminate-ghosts/eliminate-ghosts.rb
+++ b/judges/eliminate-ghosts/eliminate-ghosts.rb
@@ -52,7 +52,7 @@ Fbe.fb.txn do |fbt|
       f.stale = 'who'
     end
   end
-  fbt.query('(and (eq where "github") (exists who) (unique who) (eq stale "who"))').each do |f|
+  fbt.query('(and (eq where "github") (exists who) (eq stale "who") (unique who))').each do |f|
     next if good.include?(f.who)
     fbt.query("(and (eq who #{f.who}) (not (eq stale 'who')) (eq where 'github'))").each do |ff|
       ff.stale = 'who'
--- a/judges/eliminate-ghosts/scan-users.yml
+++ b/judges/eliminate-ghosts/scan-users.yml
@@ -39,7 +39,7 @@ input:
     stale: who
 expected:
   - /fb[count(f)=8]
-  - /fb/f[_id=7 and stale='who']
+  - /fb/f[_id=7 and stale/v='who']
   - /fb/f[_id=1 and who=526301 and not(stale)]
   - /fb/f[_id=2 and who and stale]
   - /fb/f[_id=3 and who=404001 and not(stale)]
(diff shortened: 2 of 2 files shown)


yegor256:
Merged, thanks. Moving `(unique who)` to the end of the `and` so `unique` locks in the fact that actually satisfies `(eq stale "who")`, rather than whichever fact for that user happened to come first in the factbase — fixes the silently-dead propagation loop in `eliminate-ghosts.rb` from #1922. The added pack case (facts 7/8) makes the regression visible.

---

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (15) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1254](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1997, opened 2026-09-03
Skip the active contributors count when a search answers nil

### LINKED ISSUE #1976: `Jp.qosearch` returning nil is written as a real count in `total_active_contributors.rb`
`Jp.qosearch` answers `nil` on five different paths in `lib/qos_search.rb`: the per-minute budget is spent (line 37), the search quota is zero (line 65), the quota figure is unavailable (line 59), GitHub answered `TooManyRequests` (line 75), or the call hit a transient network error (line 80). In `judges/dimensions-of-terrain/total_active_contributors.rb:30` that `nil` only makes the loop skip one repository, and line 45 still returns `{ total_active_contributors: seen.count }`.

That count is then written to the fact by `Jp.incremate`, and `lib/incremate.rb:21` skips any metric file whose property is already on the fact, so the undercount is permanent. If the budget runs out on the first repository the fact gets a hard zero and keeps it forever.

The sibling metric already knows the right move: `judges/quality-of-service/some_backlog_size.rb:20` does `return {} if found.nil?`, which leaves the property unset so the next cycle recomputes it. `total_active_contributors` should return `{}` the moment any `Jp.qosearch` call comes back `nil`, instead of `next`.

@yegor256 please take a look.

Thayorns:
`Jp.qosearch` answers `nil` on five different paths: the per-minute budget is spent, the search quota is zero or unavailable, GitHub answers `TooManyRequests`, or the call hits a transient network error. In `total_active_contributors` that `nil` was treated as "this repository has nothing", the loop moved on with `next`, and the judge still returned a count.

Since `Jp.incremate` skips any metric whose property is already on the fact, that count was written once and never recomputed. A search failure on the first repository left the fact with a permanent zero.

The metric now returns `{}` the moment any `Jp.qosearch` call comes back `nil`, the same move `some_backlog_size` already makes. The property stays unset and the next cycle counts again.

Two tests cover it: a single repository whose commit search breaks, and a pair of repositories where one search succeeds and the other fails, which previously produced a silent undercount.

Closes #1976

### DIFF
--- a/judges/dimensions-of-terrain/total_active_contributors.rb
+++ b/judges/dimensions-of-terrain/total_active_contributors.rb
@@ -27,7 +27,7 @@ def total_active_contributors(fact)
         )
         next
       end
-    next if commits.nil?
+    return {} if commits.nil?
     commits[:items].each do |commit|
       author = commit.dig(:author, :id)
       seen << author unless author.nil?
--- a/test/judges/test-dimensions-of-terrain.rb
+++ b/test/judges/test-dimensions-of-terrain.rb
@@ -1008,4 +1008,66 @@ class TestDimensionsOfTerrain < Jp::Test
       assert_nil(f['total_forks'])
     end
   end
+
+  def test_dont_count_active_contributors_when_search_breaks
+    WebMock.disable_net_connect!
+    stub_request(:get, 'https://api.github.com/rate_limit').to_return(
+      body: { resources: { search: { remaining: 30, limit: 30 } }, rate: { remaining: 1000, limit: 1000 } }.to_json,
+      headers: { 'Content-Type': 'application/json', 'X-RateLimit-Remaining' => '999' }
+    )
+    $global = {}
+    $local = {}
+    $judge = 'dimensions-of-terrain'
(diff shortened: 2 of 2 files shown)


Thayorns:
@yegor256 this one is ready for your review, happy to clarify anything about the change.

yegor256:
Thanks, treating a nil search result as "don't fill the property" rather than a real zero count is correct, with tests for both the fully-broken and partial-failure cases. CI is green.

@rultor merge

---

Thayorns:

yegor256:
Thanks for tracking down this one — treating a `nil` from `Jp.qosearch` as "nothing here" instead of "we don't know" was a subtle trap, and the two tests covering both the single-repo and partial-failure cases make the fix easy to trust. Merged.

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1388](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1998, opened 2026-09-03
Rescue Fbe::Error in total_commits and skip repos without a default branch

### LINKED ISSUE #1979: `Fbe::Error` from the GraphQL call is not rescued in `total_commits.rb`
`judges/dimensions-of-terrain/total_commits.rb:40` calls `Fbe.github_graph.total_commits(repos:)` for the whole batch at once. That method raises `Fbe::Error, "Repository '...' or branch '...' not found"` at `fbe/github_graph.rb:153` as soon as one entry in the batch has no matching ref, and line 26 of the judge feeds `json[:default_branch]` in without checking it is there.

The rescue list on lines 41 to 49 covers `GraphQL::Client::Error`, `Octokit::NotFound`, `Octokit::Deprecated`, `Octokit::Forbidden` and the network errors, but not `Fbe::Error`. `Fbe::Graph#query` never raises an Octokit error at all, so those three Octokit arms are unreachable while the exception that can actually arrive escapes and kills the judge for that cycle.

Replace the unreachable Octokit arms with `Fbe::Error`, and skip a repository at line 26 when `json[:default_branch]` is nil so one renamed default branch cannot take the whole batch down with it.

@yegor256 please take a look.

Thayorns:
The GraphQL batch call in \`total_commits.rb\` raises \`Fbe::Error\` when a repository or branch can't be found, but that exception wasn't in the rescue list. The three Octokit arms sitting there (\`NotFound\`, \`Deprecated\`, \`Forbidden\`) were dead code, since \`Fbe::Graph#query\` never raises Octokit errors — so a single renamed default branch anywhere in the batch killed the judge for that cycle instead of being logged and skipped.

This replaces those unreachable arms with \`Fbe::Error\` and adds a guard that skips a repository up front when its \`default_branch\` comes back nil, so a repo in that state never enters the batch in the first place.

Closes #1979

### DIFF
--- a/judges/dimensions-of-terrain/total_commits.rb
+++ b/judges/dimensions-of-terrain/total_commits.rb
@@ -23,6 +23,7 @@ def total_commits(_fact)
       next
     end
     next if json[:size].nil? || json[:size].zero?
+    next if json[:default_branch].nil?
     repos << [*repo.split('/'), json[:default_branch]]
   rescue Octokit::NotFound, Octokit::Deprecated => e
     $loog.info("Repository not found for #{repo}: #{e.message}")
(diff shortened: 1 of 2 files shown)


Thayorns:

yegor256:
Thanks, rescuing `Fbe::Error` from the GraphQL call and skipping repos without a default branch are both correct, with tests for each case. CI is green.

@rultor merge

---

Thayorns:

yegor256:
Good catch on the dead rescue arms — `Fbe::Graph#query` never raises Octokit errors, so those three arms could never fire. Swapping in `Fbe::Error` and guarding on a missing `default_branch` up front both make sense, and the new tests exercise each path. Merged.

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1412](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1999, opened 2026-09-03
Postpone a dimensions-of-terrain metric when a repository cannot be read

### LINKED ISSUE #1977: "will retry next cycle" never happens in the `dimensions-of-terrain` metrics
Every metric file under `judges/dimensions-of-terrain` rescues a per-repository failure, logs "(transient, will retry next cycle)" and moves on: `total_files.rb:34`, `total_contributors.rb:33`, `total_releases.rb:20`, `total_stars.rb:20`, `total_repositories.rb:17`, `total_commits.rb:19`, `total_active_contributors.rb:23`. The metric is then returned anyway, with that repository simply missing from the total.

There is no next cycle for that number. `lib/incremate.rb:21` skips a metric file when the fact already carries its property, so the value written during the failing run is the value the factbase keeps. A single 403 on one repository silently shrinks `total_files` or `total_stars` for good, and the log line promises the opposite.

The judge should treat a rescued repository as a failed measurement rather than a zero contribution: track that at least one repository was skipped and return `{}` from the metric, the way `some_backlog_size.rb:20` does, so the property stays unset and the next run measures it properly. #1206 fixed the whole-run quota case by adding `Fbe.over?` to `Jp.incremate`, but the per-repository case is still open.

@yegor256 please take a look.

Thayorns:
A 403 on a single repository used to shrink a `dimensions-of-terrain` total for good. The metric rescued the failure, logged that it would retry next cycle, and returned the sum of whatever repositories it did manage to read. `Jp.incremate` then wrote that number into the fact, and because it skips a metric whose property is already set, no later run ever looked at it again. The promise in the log line was never kept.

A skipped repository is now treated as a failed measurement of the whole metric rather than a zero contribution. `Jp.postpone` logs the repository and the error and throws; `Jp.measured`, which wraps the metric call inside `Jp.incremate`, catches that and hands back an empty hash, so nothing is written and the next run measures the metric properly. Each per-repository `rescue Octokit::Forbidden` in `total_files`, `total_contributors`, `total_releases`, `total_stars`, `total_repositories`, `total_commits`, `total_active_contributors` and `total_issues` now calls `Jp.postpone` instead of counting on.

`Octokit::NotFound` and `Octokit::Deprecated` keep skipping the repository as before, since those are not transient. The GraphQL and network rescue in `total_issues` is left alone too, and so is the network rescue around the commit count in `total_commits`; both deserve a separate look.

The test that asserted the old behaviour, `test_skips_repo_on_tree_forbidden`, now asserts the opposite and is renamed accordingly.

Closes #1977


Thayorns:
@yegor256 this is ready for your eyes, happy to clarify anything in the review.

yegor256:
Thanks, the `Jp.postpone`/`Jp.measured` mechanism makes "will retry next cycle" actually true by discarding the whole metric instead of silently keeping a partial count, and it's backed by new tests. CI is green.

@rultor merge

---

Thayorns:

yegor256:
This is a solid fix — treating a partially-scanned repository as a failed measurement instead of a silent undercount is the right call, and the postpone/measured mechanism is well tested. However, it now has merge conflicts against master: #1998 and a couple of the other just-merged terrain fixes touched the same rescue blocks in `total_commits.rb`, `total_active_contributors.rb`, `total_contributors.rb`, `total_releases.rb`, etc. that this PR also rewrites. Rather than force through a conflicted merge, I'm assigning this for a manual rebase.

---

yegor256:
(Appending the footer that should have been on my previous comment.)

---

Thayorns:

Thayorns:
@yegor256 take a look please

Thayorns:
@yegor256 please review

Thayorns:

Thayorns:

Issue #1206, opened 2025-12-02
`dimensions-of-terrain` facts showing all zeros in factbase


yegor256:
I see a few all-zero `dimensions-of-terrain` facts in the factbase: https://zerocracy.github.io/judges-action/zerocracy.html

Look like this:

```
_id _job:19627716058 _time _version total_active_contributors:0 total_commits:0 total_contributors:0 total_files:0 total_forks:0 total_issues:0 total_pulls:0 total_releases:0 total_repositories:0 total_stars:0
```

Looks like a bug.

yegor256:
@Yegorov maybe you can investigate?

PR #2001, opened 2026-09-03
fix(#1982): stop re-fetching each repo to check archived flag

### LINKED ISSUE #1982: `total_repositories.rb` re-fetches every repository to read a flag that is always false
`judges/dimensions-of-terrain/total_repositories.rb:13` counts a repository with `total += 1 unless Fbe.octo.repository(repo)[:archived]`. By the time the block runs, `lib/patches/unmask_repos.rb:52-63` has already dropped every archived repository from the list, so on every path where the flag is readable it reads false and the condition never skips anything.

The cost is one extra `repository` call per repository per cycle, spent on a branch that cannot be taken, in a judge that runs out of GitHub quota often enough for #1206 to have been filed about it.

Since the archived filter now lives in `Fbe.unmask_repos`, the body can simply be `total += 1`. #239, which asked for archived repositories to be ignored, is satisfied by the patch on its own.

@yegor256 please take a look.

Thayorns:
Fbe.unmask_repos already drops archived repositories before yielding them (lib/patches/unmask_repos.rb), so the Fbe.octo.repository(repo)[:archived] check in total_repositories.rb always read false. It cost one extra GitHub API call per repository per cycle, on a branch that could never be taken, in a judge that already runs out of quota often enough for #1206.

The fix reduces the body to a single increment per yielded repo, since the archived filter now lives entirely in unmask_repos. Added a test that fails against the old code (each repo fetched twice) and passes against the fix (fetched once).

Closes #1982.

### DIFF
--- a/judges/dimensions-of-terrain/total_repositories.rb
+++ b/judges/dimensions-of-terrain/total_repositories.rb
@@ -3,22 +3,11 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
 # SPDX-License-Identifier: MIT
 
-require 'fbe/octo'
 require 'fbe/unmask_repos'
 require_relative '../../lib/patches/unmask_repos'
 
 def total_repositories(_fact)
   total = 0
-  Fbe.unmask_repos do |repo|
-    total += 1 unless Fbe.octo.repository(repo)[:archived]
-  rescue Octokit::NotFound, Octokit::Deprecated => e
-    $loog.info("Repository #{repo} not found: #{e.message}")
-    next
-  rescue Octokit::Forbidden => e
-    $loog.warn(
-      "[#{$judge}] Repository #{repo} forbidden (transient, will retry next cycle): #{e.class}: #{e.message}"
-    )
-    next
-  end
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 could you take a look? Happy to clarify anything.

yegor256:
Thanks, dropping the redundant per-repo re-fetch since `Fbe.unmask_repos` already filters archived repos is a good simplification, confirmed by a request-count test. CI is green.

@rultor merge

---

Thayorns:

yegor256:
Good find — an unreachable archived check costing an extra API call per repository per cycle adds up fast against an already tight quota. The before/after request-count test makes the regression concrete. Merged.

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1460](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2002, opened 2026-09-03
rescue fbe error instead of dead octokit classes in total_issues judge

### LINKED ISSUE #1980: `total_issues.rb` rescues Octokit errors that a GraphQL call cannot raise
`judges/dimensions-of-terrain/total_issues.rb:15` calls `Fbe.github_graph.total_issues_and_pulls`, which goes through `Fbe::Graph#query` at `fbe/github_graph.rb:44`. That method talks to GitHub over `GraphQL::Client::HTTP` and raises `Fbe::Error` when the response carries errors; it never touches Octokit. The `rescue Octokit::NotFound, Octokit::Deprecated` on line 16 and the `rescue Octokit::Forbidden` on line 19 cannot fire.

What can fire is `Fbe::Error`, and nothing rescues it, so a rate-limited or moved repository aborts the judge on the first repository of the loop. That is the failure #1435 set out to remove, arriving through a different door: the per-repository skip added then only guards the two exception families that do not happen here.

Swap the two Octokit arms for `Fbe::Error` and keep the existing `GraphQL::Client::Error` and network arm, so a repository that the GraphQL API refuses is skipped with a log line rather than ending the run.

@yegor256 please take a look.

Thayorns:
The GraphQL query in Fbe::Graph#query raises Fbe::Error on any GraphQL failure, going through GraphQL::Client::HTTP rather than Octokit. The rescue clauses in total_issues.rb caught Octokit::NotFound, Octokit::Deprecated and Octokit::Forbidden, none of which that call path can ever raise, so a rate-limited or moved repository aborted the judge on the first repository of the loop.

This swaps those two Octokit rescue arms for a single Fbe::Error arm, logging and skipping the repository the same way the GraphQL::Client::Error and network-timeout arm already does. The GraphQL::Client::Error and network arms are unchanged.

Closes #1980

### DIFF
--- a/judges/dimensions-of-terrain/total_issues.rb
+++ b/judges/dimensions-of-terrain/total_issues.rb
@@ -13,14 +13,8 @@ def total_issues(_fact)
     json =
       begin
         Fbe.github_graph.total_issues_and_pulls(*repo.split('/'))
-      rescue Octokit::NotFound, Octokit::Deprecated => e
+      rescue Fbe::Error => e
         $loog.info("Can't count issues and pulls in #{repo}: #{e.message}")
-        next
-      rescue Octokit::Forbidden => e
-        $loog.warn(
-          "[#{$judge}] Access forbidden to issues and pulls in #{repo} " \
-          "(transient, will retry next cycle): #{e.class}: #{e.message}"
-        )
         next
       rescue GraphQL::Client::Error,
         Net::OpenTimeout, Net::ReadTimeout, SocketError, Errno::ECONNRESET, Errno::ETIMEDOUT => e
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 take a look here please

yegor256:
Thanks, rescuing `Fbe::Error` instead of dead Octokit classes on a GraphQL call is the correct fix, with a regression test. CI is green.

@rultor merge

---

Thayorns:

yegor256:
Confirmed the GraphQL path never raises Octokit errors, so those two rescue arms really were dead and a rate-limited repo would have killed the whole judge run. Swapping in `Fbe::Error` is the right fix, and it's covered by a test. Merged.

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1484](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2004, opened 2026-09-03
fix(#1985): total_files.rb honors the truncated flag on git tree

### LINKED ISSUE #1985: `total_files.rb` ignores the `truncated` flag on the git tree
`judges/dimensions-of-terrain/total_files.rb:29` asks for the recursive tree of the default branch and line 40 counts the blobs in `tree[:tree]`. GitHub caps that response at 100,000 entries or 7 MB and sets `"truncated": true` when it clips the list, which the tests already know about: the fixtures in `test/judges/test-dimensions-of-terrain.rb:101` and `:105` carry `truncated: false` explicitly.

Nothing reads the flag, so a repository large enough to be clipped contributes a capped blob count, and `Jp.incremate` writes that number and never recomputes it (`lib/incremate.rb:21`). The factbase ends up with a `total_files` that is quietly wrong for exactly the repositories where the figure matters most.

Check `tree[:truncated]` after the call and, when it is set, return `{}` from the metric so the property stays unset rather than recording a number the API said was incomplete.

@yegor256 please take a look.

Thayorns:
## Summary
- `total_files.rb` checks `tree[:truncated]` after fetching the recursive git tree; when GitHub clips the response (100,000 entries / 7 MB cap), the judge returns `{}` so `total_files` stays unset instead of recording a capped, incomplete count.
- Added `test_total_files_skips_property_when_tree_truncated` covering the truncated case.

Closes #1985

### DIFF
--- a/judges/dimensions-of-terrain/total_files.rb
+++ b/judges/dimensions-of-terrain/total_files.rb
@@ -9,6 +9,7 @@ require_relative '../../lib/patches/unmask_repos'
 
 def total_files(_fact)
   files = 0
+  truncated = false
   Fbe.unmask_repos do |repo|
     info =
       begin
@@ -37,7 +38,13 @@ def total_files(_fact)
         )
         next
       end
+    if tree[:truncated]
+      $loog.info("Tree for #{repo}@#{info[:default_branch]} is truncated, skipping total_files")
+      truncated = true
+      break
+    end
     files += (tree[:tree] || []).count { |item| item[:type] == 'blob' }
   end
+  return {} if truncated
   { total_files: files }
 end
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks, honoring the `truncated` flag on the git tree instead of silently undercounting files is correct, with a test for the truncated case. CI is green.

@rultor merge

---

Thayorns:

yegor256:
Makes sense — a capped tree response silently recording an incomplete count is worse than skipping the metric for a cycle. Test covers the truncated case cleanly. Merged.

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1532](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2003, opened 2026-09-03
disable auto pagination in Jp.qosearch

Thayorns:
Fbe.octo runs with auto_paginate = true, so a single Jp.qosearch call can walk through several pages of the GitHub Search API before returning, while the budget tracked by qos_search.rb only charges one unit per call. Each page is its own request against the search endpoint's allowance, so the state machine's per-minute budget can be overrun well beyond what it thinks it's spending.

This wraps the underlying search call in Fbe.octo.with_disable_auto_paginate, so one Jp.qosearch call now issues exactly one HTTP request, matching the one unit it charges against the budget.

Closes #1983

PR #2005, opened 2026-09-03
Rescue Fbe::OffQuota raised by the search call in Jp.qosearch

### LINKED ISSUE #1984: `Fbe::OffQuota` from the search call escapes `Jp.qosearch`
`lib/qos_search.rb:47` rescues `Fbe::OffQuota`, but that `begin` block covers only the `/rate_limit` probe on line 43. The search itself, `Fbe.octo.__send__(method, query, **)` on line 70, sits outside it, and the method-level rescue list on lines 71 to 80 lists `Octokit::TooManyRequests`, `Octokit::ServerError` and the network errors but not `Fbe::OffQuota`.

`Fbe.octo` raises `Fbe::OffQuota` before every search method when the search resource is under its threshold of five (`fbe/octo.rb:236-241`), and the threshold check inside `Jp.qosearch` only returns early when the remaining count is exactly zero (line 61). Between one and four remaining, or when the allowance drops during the auto-paginated call, the exception is raised from line 70, leaves `Jp.qosearch`, is not rescued in `judges/dimensions-of-terrain/total_active_contributors.rb` either, and ends the judge.

Adding `Fbe::OffQuota` to the method-level rescue on line 71, latching `@offquota[jg]` and returning `nil`, makes it behave the way every other quota path in that file already does.

@yegor256 please take a look.

Thayorns:
Fbe.octo raises Fbe::OffQuota before every search method once the search resource drops under its threshold of five. Jp.qosearch only rescued that exception around the /rate_limit probe, not around the actual search call on the line below it. Between one and four remaining, or when the allowance drops during an auto-paginated call, the exception escaped Jp.qosearch entirely and wasn't rescued by the calling judge either, ending the run.

This adds Fbe::OffQuota to the method-level rescue clause alongside Octokit::TooManyRequests, so it latches the quota and returns nil the same way every other quota path in this file already does. A regression test stubs Fbe.octo so the search call itself raises Fbe::OffQuota and asserts qosearch returns nil and the latch is set, instead of the exception propagating.

Closes #1984


Thayorns:
@yegor256 could you take a look when you have a moment? Happy to clarify anything.

yegor256:
Thanks, rescuing `Fbe::OffQuota` alongside `Octokit::TooManyRequests` closes the leak cleanly, with a test for the new path. CI is green.

@rultor merge

---

Thayorns:

yegor256:
This is a real gap — `Fbe::OffQuota` escaping the method-level rescue in `Jp.qosearch` between one and four search calls remaining would end the run instead of latching quota like every other path here, and the test stubbing the search call itself to raise proves it. However, it now has a merge conflict against master: #2003 (auto-pagination fix) landed first and touched the same `rescue` line in `lib/qos_search.rb`. Assigning for a manual rebase rather than forcing it through.

---

Thayorns:

Thayorns:
@yegor256 take a look please

Thayorns:
@yegor256 please review

Thayorns:

Thayorns:

PR #2006, opened 2026-09-03
Require lib/patches/unmask_repos in total_issues.rb

### LINKED ISSUE #1987: `total_issues.rb` does not require `lib/patches/unmask_repos`
Seven of the eight metric files in `judges/dimensions-of-terrain` carry `require_relative '../../lib/patches/unmask_repos'` on line 8. `total_issues.rb` requires `fbe/unmask_repos` on line 7 and stops there, so it depends on some other file in the directory having pulled the patch in first.

Today that works by luck of ordering: `lib/incremate.rb:20` globs the files in sorted order, so `total_active_contributors.rb` loads first and installs the patch. But lines 21 to 24 skip any file whose property is already on the fact, and the `require` sits after that skip. On a fact where every metric except `total_issues` was collected in an earlier cycle, `total_issues.rb` is the only file loaded, and `Fbe.unmask_repos` is then the plain gem version, which has no rescue around `octo.organization_repositories` (`fbe/unmask_repos.rb`) and aborts the judge on a `Forbidden` while expanding a wildcard mask.

Adding the same `require_relative` line that the other seven files carry removes the ordering dependency. This is adjacent to #1904 but not the same thing: that issue is about the patch duplicating the upstream method, this one is about a file that does not load it at all.

@yegor256 please take a look.

Thayorns:
Seven of the eight metric files in judges/dimensions-of-terrain require_relative the local unmask_repos patch on line 8, but total_issues.rb only requires fbe/unmask_repos and stops, relying on some other file in the directory having loaded the patch first.

That works by luck of ordering: lib/incremate.rb globs the files in sorted order, so total_active_contributors.rb loads first and installs the patch. But the loader skips any file whose property is already on the fact, and the require sits after that skip. On a fact where every metric except total_issues was already collected, total_issues.rb is the only file loaded, and Fbe.unmask_repos falls back to the plain gem version, which has no rescue around octo.organization_repositories and aborts the judge on a Forbidden while expanding a wildcard mask.

This adds the same require_relative line the other seven files carry, removing the ordering dependency.

Closes #1987

### DIFF
--- a/judges/dimensions-of-terrain/total_issues.rb
+++ b/judges/dimensions-of-terrain/total_issues.rb
@@ -5,6 +5,7 @@
 
 require 'fbe/github_graph'
 require 'fbe/unmask_repos'
+require_relative '../../lib/patches/unmask_repos'
 
 def total_issues(_fact)
   issues = 0


Thayorns:
@yegor256 please take a look — happy to clarify anything.

yegor256:
Thanks, a minimal one-line fix requiring the patch that `total_issues.rb` was missing. CI is green.

@rultor merge

---

Thayorns:

yegor256:
Good catch on the ordering dependency — relying on another file in the same directory happening to load first for a shared patch is exactly the kind of thing that breaks quietly. Bringing this file in line with its seven siblings removes the luck factor. Merged.

---

0crat:
@thayorns Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (1) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1552](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2007, opened 2026-09-03
stop counting active contributors from a search result GitHub caps at 1000

### LINKED ISSUE #1986: `total_active_contributors.rb` counts from a search result set GitHub caps at 1000
`judges/dimensions-of-terrain/total_active_contributors.rb:31` walks `commits[:items]` and collects distinct author ids. The GitHub Search API returns at most 1000 results for any query, however many pages are requested, so on a repository with more than a thousand commits in the last thirty days the distinct-author count is computed from a truncated slice and written as if it were the whole picture.

`Jp.incremate` never revisits a property it already sees (`lib/incremate.rb:21`), so the undercount is permanent for that fact. #1905 hit the same ceiling in `some_backlog_size` and escaped it by reading `total_count` instead of the items, but that escape does not exist here: distinct authors cannot be derived from a count.

The honest fix is to read `commits[:total_count]` and, when it is at or above the ceiling, return `{}` so the metric is left unset rather than recorded wrong; longer term the number is better derived from the contributors endpoint, which is not capped this way.

@yegor256 please take a look.

Thayorns:
GitHub's Search API returns at most 1000 items for any query, no matter how many pages are requested. total_active_contributors read commits[:items] directly, so on a repository with more than a thousand commits in the last thirty days the distinct-author count was computed from a truncated slice and recorded as if it were complete. Jp.incremate never revisits a property it already sees, so the wrong count stuck permanently.

This checks commits[:total_count] before touching the items and, once it reaches the 1000-item ceiling, leaves the fact unset for that cycle instead of writing an undercount. Reading the contributors endpoint instead of search results would remove the ceiling entirely, but that's a bigger change than this fix.

Closes #1986


Thayorns:

yegor256:
Thanks, skipping counting once the search hits GitHub's 1000-result ceiling avoids a silently truncated count, and it's well tested. CI is green.

@rultor merge

---

Thayorns:

yegor256:
This is a real bug — GitHub's Search API caps results at 1000 regardless of pagination, so a busy repository would have its distinct-author count silently truncated and, since `Jp.incremate` never revisits a set property, stuck wrong forever. Good fix, well tested. It now conflicts against master though: #1997 landed first and touched the same spot in `total_active_contributors.rb` (the `commits.nil?` guard right above where this adds the ceiling check). Assigning for a manual rebase.

---

Thayorns:

Thayorns:
@yegor256 take a look please

Thayorns:
@yegor256 please review

Thayorns:

Thayorns:

PR #2009, opened 2026-09-03
fix(#1988): key incremate's completion check on the file's own property, not write order

### LINKED ISSUE #1988: `Jp.incremate` keys the skip on the file name, not on the properties the file writes
`lib/incremate.rb:21` decides whether a metric has already been collected by testing `fact[n]`, where `n` is the base name of the metric file. Two metric files write more than one property: `judges/dimensions-of-terrain/total_stars.rb:30` returns `total_stars` and `total_forks`, and `total_issues.rb:36` returns `total_issues` and `total_pulls`. Line 44 writes them one at a time, in a loop over the hash.

If the process is stopped between the two writes — the `judges` gem runs each judge under a timeout and raises `Timeout::ExitException` at an arbitrary point — the fact keeps `total_stars` and loses `total_forks`. On the next cycle line 21 sees `total_stars` on the fact, skips the file, and `total_forks` is never collected for that fact again.

Checking every property the file is known to produce, or writing the whole hash in one step so it cannot be half-applied, closes the gap; the name-based check only happens to be right for the six files that emit a single property.

@yegor256 please take a look.

Thayorns:
## Summary

- `lib/incremate.rb` skipped re-evaluating a metric file once `fact[n]` was truthy, where `n` is the file's basename.
- For multi-property files (`total_stars.rb`, `total_issues.rb`, `some_issue_lifetime.rb`, `total_commits_pushed.rb`, `total_issues_created.rb`, `some_merged_pulls.rb`, `some_pull_hoc_size.rb`, `some_build_success_rate.rb`, `some_release_hoc_size.rb`, `some_review_time.rb`) `n` is consistently the *first* key each file writes, so a process killed mid-write (the `judges` gem timeout can raise `Timeout::ExitException` at any point) left the fact permanently marked "done" while its other properties were never collected.
- Fixed by always writing the property named after the file *last*, so `fact[n]` only becomes truthy once every property from that run has actually been written — no new bookkeeping property needed, and no behavior change for single-property files.
- `avoid_duplicate` now defaults to `true` so a retried file (after a resumed partial write) doesn't re-append values that were already written before the interruption.

Closes #1988

@yegor256 please take a look.

### DIFF
--- a/lib/incremate.rb
+++ b/lib/incremate.rb
@@ -12,7 +12,7 @@ require 'time'
 require_relative 'jp'
 
 def Jp.incremate(
-  fact, dir, prefix, avoid_duplicate: false, pause: 0,
+  fact, dir, prefix, avoid_duplicate: true, pause: 0,
   max_per_fact: nil,
   epoch: $epoch || Time.now, kickoff: $kickoff || Time.now
 )
@@ -40,8 +40,12 @@ def Jp.incremate(
     elapsed($loog, level: Logger::INFO) do
       h = __send__(n, fact)
       h.each do |k, v|
+        next if k.to_s == n
         next if avoid_duplicate && fact.all_properties.include?(k.to_s)
         Array(v).each { fact.__send__("#{k}=", _1) }
+      end
+      if h.key?(n.to_sym) && !(avoid_duplicate && fact.all_properties.include?(n))
+        Array(h[n.to_sym]).each { fact.__send__("#{n}=", _1) }
       end
       throw(:"Collected #{n}: [#{h.map { |k, v| "#{k}: #{v}" }.join(', ')}]")
     end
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks, keying the completion check on the file's own property instead of write order is a sound fix for #1988, with a regression test. CI is green.

@rultor merge

---

Thayorns:

yegor256:
Good diagnosis — keying the "already done" check on write order rather than a property's own presence meant a process killed mid-write could permanently mark a multi-property file done while most of its properties never landed. Writing the file's own property last, with `avoid_duplicate` on by default so a retry doesn't double up, is a solid fix with a test that reproduces the partial-write scenario. Merged.

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1600](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2011, opened 2026-09-03
Guard total_contributors against a nil contributor id

### LINKED ISSUE #1990: `total_contributors.rb` has no nil guard on the contributor id
`judges/dimensions-of-terrain/total_contributors.rb:42` does `contributors << contributor[:id]` for every entry the contributors endpoint returns. Its sibling in the same directory, `total_active_contributors.rb:33`, writes `seen << author unless author.nil?` for the same kind of value, because a GitHub author record can come back without an id.

The two files disagree about whether that can happen, so one of them is wrong. If an entry does arrive without an id, `total_contributors` puts `nil` into the set and counts it as a person, which inflates the metric by one — and `lib/incremate.rb:21` means the inflated number is the one the factbase keeps.

Whichever way the question is settled, the two files should answer it the same way; the cheap direction is to add the same `unless ... .nil?` guard to line 42.

@yegor256 please take a look.

Thayorns:
total_active_contributors.rb already skips entries where the author has no id, since GitHub's contributors endpoint can return a record without one. total_contributors.rb did not have the same guard, so a missing id landed in the set as nil and inflated the count by one.

This adds the same `unless id.nil?` check to total_contributors.rb, along with a test that stubs a contributors response containing one entry without an id.

Closes #1990

### DIFF
--- a/judges/dimensions-of-terrain/total_contributors.rb
+++ b/judges/dimensions-of-terrain/total_contributors.rb
@@ -39,7 +39,8 @@ def total_contributors(_fact)
       end
     next unless list.is_a?(Array)
     list.each do |contributor|
-      contributors << contributor[:id]
+      id = contributor[:id]
+      contributors << id unless id.nil?
     end
   rescue Octokit::NotFound, Octokit::Deprecated => e
     $loog.info("Repository/contributors info not found for #{repo}: #{e.message}")
--- a/test/judges/test-dimensions-of-terrain.rb
+++ b/test/judges/test-dimensions-of-terrain.rb
@@ -726,6 +726,45 @@ class TestDimensionsOfTerrain < Jp::Test
     end
   end
 
+  def test_total_contributors_skips_entries_without_id
+    WebMock.disable_net_connect!
+    rate_limit_up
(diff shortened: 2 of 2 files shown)


Thayorns:
@yegor256 happy to clarify anything about this fix if needed.

yegor256:
Thanks, guarding against a nil contributor id is a correct and well-tested fix. CI is green.

@rultor merge

---

Thayorns:

yegor256:
Good consistency fix — bringing the nil-id guard from `total_active_contributors.rb` over to `total_contributors.rb` closes a real off-by-one when GitHub's contributors endpoint returns a record without an id. Test covers it directly. Merged.

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1624](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2013, opened 2026-09-03
fix skip-recounting.yml to actually assert recounting is skipped

### LINKED ISSUE #1991: `skip-recounting.yml` does not test that recounting is skipped
`judges/dimensions-of-terrain/skip-recounting.yml` sets `TODAY` to 2024-03-03T01:01 and seeds a fact dated 2024-03-03T00:00, so the query in `dimensions-of-terrain.rb:20-24` finds it and no new fact is inserted. The single expectation is `/fb[count(f)=1]`.

That is not what the file is named after. The seeded fact carries only `what` and `when`, so `Jp.incremate` runs over it and collects all ten totals; nothing about recounting is skipped, and the test would pass unchanged if every metric were recomputed on every run. It proves one thing — no duplicate fact — and its name promises another.

Seeding the fact with a couple of totals already on it and expecting those values to be unchanged would test what the name says; alternatively the file should be renamed to match the assertion it actually makes.

@yegor256 please take a look.

Thayorns:
The skip-recounting.yml fixture seeded a fact with only what/when, so Jp.incremate collected every total metric on the first run and the single expectation (/fb[count(f)=1]) only proved that no duplicate fact got inserted, not that recounting was skipped. The test would pass unchanged even if every metric were recomputed on every run.

Seeded the fact with total_commits and total_stars already set, and now assert those values stay unchanged, that total_forks (the sibling key produced by the same total_stars.rb collector) stays unset since that file never re-runs, and that total_files, which wasn't pre-seeded, still gets collected normally.

Closes #1991

### DIFF
--- a/judges/dimensions-of-terrain/skip-recounting.yml
+++ b/judges/dimensions-of-terrain/skip-recounting.yml
@@ -11,5 +11,11 @@ input:
     _id: 42
     what: dimensions-of-terrain
     when: 2024-03-03T00:00:00
+    total_commits: 777
+    total_stars: 555
 expected:
   - /fb[count(f)=1]
+  - /fb/f[total_commits=777]
+  - /fb/f[total_stars=555]
+  - /fb/f[not(total_forks)]
+  - /fb/f[total_files]


Thayorns:

yegor256:
Thanks, this makes the fixture actually assert that recounting is skipped rather than just checking fact count. CI is green.

@rultor merge

---

Thayorns:

yegor256:
Right, the old fixture only proved no duplicate fact was inserted, not that skip-recounting actually skips recounting. Seeding real values and asserting they're untouched, plus asserting the sibling key stays unset and the unseeded metric still gets collected, is a much better test of what the file is named for. Merged.

---

0crat:
@thayorns Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (6) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1664](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2014, opened 2026-09-03
Pin expected totals in dimensions-of-terrain measure-all.yml

### LINKED ISSUE #1992: `measure-all.yml` asserts that the totals exist, never what they are
All ten expectations in `judges/dimensions-of-terrain/measure-all.yml:15-24` have the shape `/fb/f[total_repositories]`, which is satisfied by a fact that carries the property at all. The judge runs against the fake Octokit and the fake GraphQL client, both of which return fixed, known numbers, so the values are perfectly assertable.

Because nothing checks them, the all-zero facts reported in #1206 would have passed every one of these expectations, and so would any future regression that writes a zero instead of a measurement — which is exactly the failure mode this judge keeps hitting.

Pinning the expected numbers, `/fb/f[total_repositories=1]` and so on for the ten totals, turns the file into a test of the arithmetic rather than a test of the property names.

@yegor256 please take a look.

Thayorns:
All ten totals asserted in judges/dimensions-of-terrain/measure-all.yml only checked that a property existed, not what it held. Since the judge runs against the fake Octokit and fake GraphQL clients, which return fixed numbers for the test repository, those numbers can be pinned exactly.

The expectations now read `/fb/f[total_repositories=1]` and so on for all ten totals, so a regression that reports zero (or any other wrong number) fails the test instead of passing it silently.

Closes #1992


Thayorns:

yegor256:
Thanks, pinning the expected totals makes this a real regression test instead of just an existence check. CI is green.

@rultor merge

---

Thayorns:

yegor256:
Good tightening — checking only for existence let a regression that zeroes out a total pass silently. Pinning exact values against the fake clients closes that gap. Merged.

---

0crat:
@thayorns Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1684](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2015, opened 2026-09-03
Assert the design a rescue test proves, not a coincidental exception class

### LINKED ISSUE #1994: `test_total_issues_keeps_local_graph_bug` locks in a bug with no issue behind it
`test/judges/test-dimensions-of-terrain.rb:373-391` builds a fake GraphQL client that raises `NoMethodError`, then asserts `assert_raises(NoMethodError) { total_issues(nil) }`. The name says what it does: it keeps a known defect in place, so that any attempt to make `total_issues.rb` survive a broken graph client turns the suite red.

There is no issue number in the test and nothing in the file explaining which bug is being preserved or why the judge should die on it, so the next person to touch the rescue list in `judges/dimensions-of-terrain/total_issues.rb:16-32` has no way to tell a deliberate constraint from an accident. `NoMethodError` is also broad enough that the assertion passes for reasons other than the one intended — any typo inside the judge satisfies it.

Either the behaviour being preserved deserves its own issue that the test can point at, or the test should assert the specific failure it cares about rather than the exception class.

@yegor256 please take a look.

Thayorns:
\`test_total_issues_keeps_local_graph_bug\` raised a bare \`NoMethodError\` and asserted only on its class. That name and shape made it read as pinning a known defect with no issue behind it, and the assertion was loose enough to pass for reasons unrelated to the one intended — any stray typo inside the judge would satisfy it too.

The test is renamed to \`test_total_issues_lets_unrescued_errors_propagate\` and now raises a distinct error instance defined in the test itself, so the assertion can only pass for the reason it claims: \`total_issues\` does not swallow errors outside its rescue list.

A second test, \`test_total_issues_does_not_rescue_octokit_unauthorized\`, covers the actual boundary of that rescue list — \`Octokit::Unauthorized\` looks close enough to the rescued \`Octokit::Forbidden\` that someone could assume it is also caught, and it isn't.

Closes #1994


Thayorns:

yegor256:
Thanks, this tightens the test to assert the actual design (unrescued errors propagate) instead of a coincidental `NoMethodError`. CI is green.

@rultor merge

---

Thayorns:

yegor256:
This is a good tightening — a bare `NoMethodError` asserted only by class read as pinning an untracked defect and could pass for the wrong reason, and testing the actual `Octokit::Unauthorized` vs `Octokit::Forbidden` boundary is a real gap worth covering. It now conflicts against master though: #2002 landed first and inserted a new test immediately above the one this PR renames, in the same file. Assigning for a manual rebase.

---

Thayorns:
@yegor256 take a look at this one please

Thayorns:
@yegor256 please review

Thayorns:

Thayorns:

Issue #1942, opened 2026-09-01
A rescued error is written as a real comment count in `code-was-reviewed.rb`


Thayorns:
The two comment counters in `judges/code-was-reviewed/code-was-reviewed.rb` turn every rescued error into a zero. `Fbe.octo.issue_comments` at line 110 falls back to `0` in all three of its rescue arms, and `Fbe.octo.pull_request_review_comments` at line 132 does the same. Those zeros are then stored as real values, `n.comments = count` at line 129 and `n.review_comments = ...` at line 130.

Nothing marks the fact as incomplete afterwards. The query at lines 25-31 skips any pull that already carries a `code-was-reviewed` fact, so the judge never comes back to it. One throttled or briefly forbidden run is enough to record a review as having zero comments for good, and that number feeds the reward math.

The fix is to leave the fact alone rather than to invent a number: let the rescue arms `next` out of the transaction the way the outer chains at lines 54-67 do, so the pull is picked up again on the following cycle. A permanent `NotFound` on the comments endpoint is the one case where zero is honest.

@yegor256 please take a look

0crat:
@thayorns Thanks for reporting a new bug! You've earned +18 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+376](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2031, opened 2026-09-05
Skip workflow runs whose usage cannot be read

### LINKED ISSUE #2024: A rescued error is recorded as a zero-second build in `some_build_success_rate.rb`
## What happens

When the run usage call fails, the duration becomes zero and is then used as a
measurement:

https://github.com/zerocracy/judges-action/blob/master/judges/quality-of-service/some_build_success_rate.rb#L33

```ruby
secs =
  begin
    (Fbe.octo.workflow_run_usage(repo, json[:id])[:run_duration_ms] || 0) / 1000
    $loog.info("Workflow run usage not found for #{repo}##{json[:id]}: #{e.message}")
    0
    $loog.warn(...)
    0
  end
{ json: json, secs: secs, completed: json[:run_started_at] + secs }
```

`duration << secs` a few lines later records a build that took zero seconds, which is
not a thing that happened.

It also corrupts two other numbers. `completed` becomes `run_started_at`, so
`runs.sort_by! { _1[:completed] }` puts the run earlier than it belongs, and the
failure-to-success pairing that follows walks the runs in that order:

```ruby
if json[:conclusion] == 'failure' && failed[wid].nil?
  failed[wid] = completed
elsif json[:conclusion] == 'success' && failed[wid]
  ttrs << Integer(completed - failed[wid])
```

so `some_build_time_to_repair` is computed from a wrong ordering and a wrong timestamp.

The `Forbidden` arm even says "transient, will retry next cycle", but the zero it
returns is written into the fact right now, and the fact is not revisited.

## What should happen

A run whose usage could not be read should be skipped, not counted as a zero-second
build, so that a transient API failure lowers the sample size instead of the average.

Thayorns:
When `workflow_run_usage` came back as a 404, a 403 or a deprecation, the rescue arm returned zero seconds and the run was recorded as if it had really taken no time at all. That zero landed in `some_build_duration`, and because `completed` is `run_started_at + secs` it also collapsed to the start time, which reordered `runs.sort_by!` and fed the wrong timestamp into the failure-to-success pairing behind `some_build_mttr`. The `Forbidden` arm called itself transient while writing a permanent number.

Now such a run is dropped: `map` becomes `filter_map` and every arm that cannot read the usage does `next`. A transient API failure lowers the sample size instead of pulling the average down.

The same fabricated zero came from `run_duration_ms` being nil, so that path is skipped too rather than defaulting to zero. The tests named `test_skips_run_on_not_found_workflow_run_usage` and `test_skips_run_on_forbidden_workflow_run_usage` asserted the old zeros despite their names, so their expectations were corrected, and three more cover the success rate, the repair times and the missing duration.

`code-was-reviewed.rb` has the same shape and is tracked separately in #1942, so it is untouched here.

Closes #2024

### DIFF
--- a/judges/quality-of-service/some_build_success_rate.rb
+++ b/judges/quality-of-service/some_build_success_rate.rb
@@ -30,19 +30,21 @@ def some_build_success_rate(fact)
       end
     wfs = workflows.select { |json| json[:status] == 'completed' && !json[:conclusion].nil? }.first(60)
     runs =
-      wfs.map do |json|
+      wfs.filter_map do |json|
         secs =
           begin
-            (Fbe.octo.workflow_run_usage(repo, json[:id])[:run_duration_ms] || 0) / 1000
+            ms = Fbe.octo.workflow_run_usage(repo, json[:id])[:run_duration_ms]
+            next if ms.nil?
+            ms / 1000
           rescue Octokit::NotFound, Octokit::Deprecated => e
             $loog.info("Workflow run usage not found for #{repo}##{json[:id]}: #{e.message}")
-            0
+            next
           rescue Octokit::Forbidden => e
             $loog.warn(
               "[#{$judge}] Access forbidden to workflow run usage for #{repo}##{json[:id]} " \
               "(transient, will retry next cycle): #{e.class}: #{e.message}"
             )
-            0
+            next
           end
         { json: json, secs: secs, completed: json[:run_started_at] + secs }
       end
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 take a look please

Thayorns:
@yegor256 please review

yegor256:
Thanks for the fix, merged!

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +7 for the 95 hits-of-code that you wrote; -7 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1846](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2033, opened 2026-09-05
Skip total_commits instead of writing a zero when the count fails

### LINKED ISSUE #2025: A rescued error is written as a real `total_commits` of zero
## What happens

A failed GraphQL or network call is recorded as a real count of zero:

https://github.com/zerocracy/judges-action/blob/master/judges/dimensions-of-terrain/total_commits.rb#L30

```ruby
total_commits:
begin
  repos.empty? ? 0 : Fbe.github_graph.total_commits(repos:).sum { _1['total_commits'] }
rescue GraphQL::Client::Error, Fbe::Error => e
  $loog.info("Can't count total commits: #{e.message}")
  0
rescue Net::OpenTimeout, Net::ReadTimeout, SocketError, Errno::ECONNRESET, Errno::ETIMEDOUT => e
  $loog.warn("[#{$judge}] Network error counting commits: #{e.message}")
  0
end
```

The `0` is returned as the metric, so the judge writes `total_commits = 0` into the
fact. This is a cumulative total that only ever grows, so a single timeout produces a
drop to zero on the vitals chart, followed by a jump back on the next cycle.

The distinction that is lost is the one between "we looked and there are no commits"
and "we could not look". Both are written as the same number, and the log line is the
only place the difference survives.

## What should happen

When the count could not be obtained, the judge should omit `total_commits` from the
returned hash rather than write a zero, so the previous value stands until the next
successful cycle.

Thayorns:
A failed GraphQL call or a network timeout while counting commits used to be recorded as a real count of zero. Since `total_commits` is a cumulative total that only ever grows, one timeout dragged the vitals chart down to zero and the next cycle threw it back up again. The number lost the distinction between "we looked and there are no commits" and "we could not look", and only the log line kept it.

The judge now returns an empty hash from both rescue branches, so `incremate` writes no property at all and the value from the previous cycle stands until a count succeeds. This is the same shape `total_files` already uses when it gives up on a truncated tree. A genuine zero still reaches the fact when there are no repositories to measure. The two log messages now carry the repository count, and `net/http` is required explicitly, since the network rescue is what the new behaviour rests on.

Three tests cover it: the two rescue branches in isolation, and one that runs the whole judge and asserts the fact carries no `total_commits` after the graph fails.

The same shape lives in `code-was-reviewed.rb`, reported separately as #1942, and is left alone here.

Closes #2025

### DIFF
--- a/judges/dimensions-of-terrain/total_commits.rb
+++ b/judges/dimensions-of-terrain/total_commits.rb
@@ -6,6 +6,7 @@
 require 'fbe/github_graph'
 require 'fbe/octo'
 require 'fbe/unmask_repos'
+require 'net/http'
 require_relative '../../lib/patches/unmask_repos'
 
 def total_commits(_fact)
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 this one stops a rescued GraphQL or network error from being written as a real `total_commits` of zero. Happy to clarify anything or rework it if you'd rather see a different shape.

Thayorns:
@yegor256 please review

yegor256:
Thanks for the fix, merged!

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +6 for the 87 hits-of-code that you wrote; -7 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1822](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2034, opened 2026-09-05
Omit total_builds_ran when a repository cannot be read

### LINKED ISSUE #2026: A repository that could not be read contributes zero to `total_builds_ran`
## What happens

A repository whose workflow runs cannot be read contributes zero to the total:

https://github.com/zerocracy/judges-action/blob/master/judges/quantity-of-deliverables/total_builds_ran.rb#L14

```ruby
Fbe.unmask_repos.sum do |repo|
  Fbe.octo.with_disable_auto_paginate do |octo|
    octo.repository_workflow_runs(repo, ..., per_page: 1)[:total_count]
  end
  $loog.info("Workflow runs not found for #{repo}: #{e.message}")
  0
  $loog.warn(
    "[#{$judge}] Access forbidden to workflow runs for #{repo} " \
  )
  next 0
end
```

The `Forbidden` arm calls the failure transient and promises a retry, then contributes
`0` to a sum that is written to the fact immediately. Nothing revisits that fact, so
the number for that period stays permanently short by whatever that repository ran.

With several repositories the shortfall is invisible, since the total is still a
plausible non-zero number.

The two arms also disagree in form, `0` versus `next 0`, for no reason that shows in
the result.

## What should happen

If any repository's count could not be read, the judge should omit `total_builds_ran`
rather than report a total it knows is incomplete.

Thayorns:
When `repository_workflow_runs` failed for one repository, that repository quietly contributed zero to the sum, and the sum was stored on the fact as if it were complete. Nothing ever revisits a fact that already carries `total_builds_ran`, so the number for that period stayed short forever. The `Forbidden` arm was the worst of it: it logged that the failure was transient and would be retried next cycle, and then made sure no retry could ever happen.

Now an unreadable repository ends the count instead of shrinking it. Both rescue arms return an empty hash, `incremate` writes nothing, and the property stays absent, so the next cycle picks the judge up again and has a real chance of producing an honest total. The two arms also agree in form now, rather than one saying `0` and the other `next 0` for the same effect.

The old `total_builds_ran_skips_forbidden` test asserted the bug, so it is replaced by three: a forbidden repository on its own, a forbidden repository alongside a readable one that used to yield a plausible partial total, and an unreadable one.

Closes #2026

### DIFF
--- a/judges/quantity-of-deliverables/total_builds_ran.rb
+++ b/judges/quantity-of-deliverables/total_builds_ran.rb
@@ -7,25 +7,28 @@ require 'fbe/octo'
 require 'fbe/unmask_repos'
 
 def total_builds_ran(fact)
-  {
-    total_builds_ran:
-      Fbe.unmask_repos.sum do |repo|
-        Fbe.octo.with_disable_auto_paginate do |octo|
-          octo.repository_workflow_runs(
-            repo,
-            created: "#{fact.since.utc.iso8601[0..9]}..#{fact.when.utc.iso8601[0..9]}",
-            per_page: 1
-          )[:total_count]
-        end
-      rescue Octokit::NotFound, Octokit::Deprecated => e
-        $loog.info("Workflow runs not found for #{repo}: #{e.message}")
-        0
-      rescue Octokit::Forbidden => e
-        $loog.warn(
-          "[#{$judge}] Access forbidden to workflow runs for #{repo} " \
-          "(transient, will retry next cycle): #{e.class}: #{e.message}"
-        )
-        next 0
+  total = 0
+  Fbe.unmask_repos.each do |repo|
+    total +=
+      Fbe.octo.with_disable_auto_paginate do |octo|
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 this one makes the judge stay silent instead of reporting a total it knows is short. Happy to clarify anything.

Thayorns:
@yegor256 please review

yegor256:
Thanks for the fix, merged!

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +7 for the 95 hits-of-code that you wrote; -7 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1798](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2035, opened 2026-09-05
fix(#2027): treat an absent field as a non-match in twice

### LINKED ISSUE #2027: `Jp.twice?` drops absent fields, so a missing `label` makes every later label a duplicate
## What happens

`Jp.twice?` drops the fields the fact does not carry, then compares on what is left:

https://github.com/zerocracy/judges-action/blob/master/lib/twice.rb#L9

```ruby
pairs = fields.to_h { [_1, fact[_1]&.first] }
pairs.compact!
fb.query(
  "(and (eq what $what) #{pairs.keys.map { "(eq #{_1} $#{_1})" }.join(' ')})"
).each(fb, pairs.merge('what' => what)).to_a.size > 1
```

so a missing field makes the query broader, not narrower, and the answer flips from
"not a duplicate" to "duplicate".

The caller uses it to decide whether to keep a newly detected event:

https://github.com/zerocracy/judges-action/blob/master/judges/github-events/github-events.rb#L476

```ruby
uniques.each { |w, ff| throw :rollback if Jp.twice?(fbt, f, w, ff) }
```

with these field sets:

```ruby
'label-was-attached' => %w[where repository issue label],
'type-was-attached'  => %w[where repository issue type],
'issue-was-opened'   => %w[where repository issue who],
```

For `label-was-attached` without `label`, the query collapses to "any
`label-was-attached` on this issue in this repository". The first label on an issue is
kept, and every later one is rolled back as a duplicate, so the factbase records one
label per issue no matter how many were attached.

The same holds for `type-was-attached` without `type`, and for the `who` sets when the
actor is missing, which is what happens for a deleted account.

## What should happen

A field that the fact does not carry should not be silently dropped from the
comparison. `Jp.twice?` should treat an absent field as a non-match, using
`(absent ...)` for it, so that removing a field never widens the query.

Thayorns:
`Jp.twice?` built its duplicate-detection query from the fields the fact actually carried, dropping the rest. Dropping a field widens the query instead of narrowing it, so a fact missing one of the compared fields matched every other fact that shared the remaining ones, and the answer flipped from "not a duplicate" to "duplicate".

In `github-events` that decides whether a freshly detected event survives. A `label-was-attached` fact without a `label` collapsed the query to "any label was attached to this issue in this repository", so the first label on an issue was kept and every later one was rolled back. The same happened to `type-was-attached` without a `type`, and to the `who` sets when the actor was gone, which is what a deleted account looks like.

Now every compared field contributes a term: an equality when the fact carries a value, and `(absent ...)` when it does not. Removing a field can no longer broaden the match, and two facts count as the same event only when they agree on the present fields and on the missing ones.

The tests cover an unlabeled event standing next to a labeled one, two unlabeled events that really are duplicates, a second distinct label on the same issue, and an authorless `issue-was-opened` alongside one that names an author.

Closes #2027

### DIFF
--- a/lib/twice.rb
+++ b/lib/twice.rb
@@ -7,8 +7,7 @@ require_relative 'jp'
 
 def Jp.twice?(fb, fact, what, fields)
   pairs = fields.to_h { [_1, fact[_1]&.first] }
-  pairs.compact!
   fb.query(
-    "(and (eq what $what) #{pairs.keys.map { "(eq #{_1} $#{_1})" }.join(' ')})"
-  ).each(fb, pairs.merge('what' => what)).to_a.size > 1
+    "(and (eq what $what) #{fields.map { pairs[_1].nil? ? "(absent #{_1})" : "(eq #{_1} $#{_1})" }.join(' ')})"
+  ).each(fb, pairs.compact.merge('what' => what)).to_a.size > 1
 end
--- a/test/lib/test_twice.rb
+++ b/test/lib/test_twice.rb
@@ -65,15 +65,61 @@ class TestTwice < Minitest::Test
     )
   end
 
+  def test_lets_an_unlabeled_event_through_when_another_one_has_a_label
+    fb = Factbase.new
+    event(fb, 'label-was-attached', 42)
+    insert(fb, 'bug', repository: 42)
+    refute(
+      Jp.twice?(fb, fb.query('(absent label)').each.to_a.first, 'label-was-attached', %w[where repository issue label]),
+      'an event without a label cannot be a duplicate of one that carries a label'
+    )
+  end
+
+  def test_finds_the_second_copy_when_neither_event_has_a_label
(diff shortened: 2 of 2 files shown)


Thayorns:
@yegor256 this makes `Jp.twice?` compare absent fields with `(absent ...)` instead of dropping them, so a missing label or actor no longer widens the query. Happy to clarify anything here.

Thayorns:
@yegor256 please review

yegor256:
Thanks for tracking this down, merged!

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1774](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2040, opened 2026-09-05
Read a route array as a status pair only when it starts with an integer

### LINKED ISSUE #2032: `Jp::FakeGithub` cannot serve a JSON array body in `test/fake_github.rb`
The `reply` method in `test/fake_github.rb:81` decides what a route means by its
class, and it treats every `Array` as the pair `[status, body]`. A route whose
body is itself a JSON array is therefore read as that pair: declaring
`'GET /repos/foo/foo/releases?per_page=100' => [{ id: 2 }, { id: 1 }]` binds
`status` to the first hash and `body` to the second. `answer` at line 66 then
interpolates that hash into the status line it builds at line 70, and the client
gets `HTTP/1.1 {id: 2, ...} Error` instead of `HTTP/1.1 200 OK`, which Faraday
rejects as a malformed status line.

Most of the GitHub endpoints these tests need answer with an array — releases,
issues, commits, comments, workflow runs — so this meets anyone declaring one.
The failure names neither the route nor the array, and it surfaces as a
connection error rather than as a route the server cannot express, so it reads
like a broken socket. The workaround is to write the body as `[200, [...]]`, and
that is not written down anywhere. #2028 asks for twenty eight more test files
to move onto this server, and each of them will hit this.

The shape of the route says what it is, so `reply` should read the shape rather
than the class: treat an `Array` as a `[status, body]` pair only when its first
element is an `Integer`, and serve any other array as the body. That matches the
`Integer` branch above it, which already reads a bare status code that way, and
it leaves every route now in the tests working as it does today.

@yegor256

Thayorns:
`Jp::FakeGithub#reply` decided what a route declaration meant by its class, and every `Array` was read as the pair `[status, body]`. A route whose body is itself a JSON array was therefore torn apart: the first element became the status and the rest became the body, so the server wrote `HTTP/1.1 {id: 2, ...} Error` on the wire and Faraday refused it as a malformed status line. Since releases, issues, commits, comments and workflow runs all answer with an array, this met almost anyone declaring a real GitHub endpoint, and it surfaced as a connection error rather than as a route the server cannot express.

The route now says its own shape. An `Array` is read as a status pair only when its first element is an `Integer`, and any other array is served as the body. That mirrors the `Integer` branch above it, which already reads a bare status code the same way, and every route declared in the tests today keeps working exactly as before.

`test/test_fake_github.rb` covers the server itself for the first time: an array body of one and of several releases, an empty array, a status paired with an array body, and a plain hash body.

Closes #2032

### DIFF
--- a/test/fake_github.rb
+++ b/test/fake_github.rb
@@ -81,7 +81,7 @@ class Jp::FakeGithub
   def reply(route)
     case route
     when Integer then [route, {}]
-    when Array then route
+    when Array then route.first.is_a?(Integer) ? route : [200, route]
     else [200, route]
     end
   end
--- a/test/test_fake_github.rb
+++ b/test/test_fake_github.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+require 'octokit'
+require_relative 'fake_github'
+require_relative 'test__helper'
+
+class TestFakeGithub < Jp::Test
+  def test_serves_array_body
+    seed = Random.new_seed
+    number = Random.new(seed).rand(1_000_000)
+    Jp::FakeGithub.new(
+      'GET /repos/foo/foo/releases' => [{ id: number, name: 'Ω первый' }]
+    ).run do
+      assert_equal(
+        number, Octokit::Client.new.releases('foo/foo').first[:id],
+        "the array body did not reach the client as the single release declared, seed #{seed}"
+      )
+    end
+  end
+
+  def test_serves_empty_array_body
+    Jp::FakeGithub.new(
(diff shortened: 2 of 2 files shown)


Thayorns:
@yegor256 happy to explain any part of this or adjust the shape rule if you would rather the server rejected an ambiguous declaration outright.

yegor256:
Thanks for the fix, merged!

---

0crat:
@thayorns Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +5 for the 76 hits-of-code that you wrote; -6 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1894](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2043, opened 2026-09-06
#2037: log the count where it is not zero

### LINKED ISSUE #2037: `Jp.issue_was_lost` always logs "0 facts marked as stale"
## What happens

`Jp.issue_was_lost` returns early when it marked something stale, and then reports the
count it can no longer have:

https://github.com/zerocracy/judges-action/blob/master/lib/issue_was_lost.rb

```ruby
stale =
  Fbe.fb.query("...").each { |f| f.stale = 'issue' }
Fbe::Tombstone.new.bury!(where, repository, issue)
return if stale.positive?
...
$loog.info("The issue #{issue} was marked as lost, #{stale} facts marked as stale")
```

`return if stale.positive?` means the last line is only reachable when `stale` is zero, so
the log always says:

```
The issue 42 was marked as lost, 0 facts marked as stale
```

The number was meant to tell the reader how many facts the call touched, and it can only
ever be the one value that says nothing.

## What should happen

Either the count should be logged where it is not zero, next to the `return`, or it should
be dropped from the message that stays.

morphqdd:
`return if stale.positive?` meant the line below it was reachable only when `stale` was
zero, so the log always said "0 facts marked as stale".

The count now goes into the branch that has it, and the message that stays does not
mention it.

Closes #2037

### DIFF
--- a/lib/issue_was_lost.rb
+++ b/lib/issue_was_lost.rb
@@ -18,7 +18,10 @@ def Jp.issue_was_lost(where, repository, issue)
       (absent tombstone))"
     ).each { |f| f.stale = 'issue' }
   Fbe::Tombstone.new.bury!(where, repository, issue)
-  return if stale.positive?
+  if stale.positive?
+    $loog.info("The issue #{issue} was marked as lost, #{stale} facts marked as stale")
+    return
+  end
   f =
     Fbe.if_absent do |n|
       n.issue = issue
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks for the contribution, merged!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (21) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+7300](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2044, opened 2026-09-06
#1664: do not ask about reactions of a comment that has none

### LINKED ISSUE #1664: N+1 reactions API calls in lib/pull_request.rb count_appreciated_comments

morphqdd:
`count_appreciated_comments` asked GitHub for the reactions of every comment, so a pull
request with two hundred comments cost two hundred calls, and the quota it spent belonged
to every other judge too.

The list that brings a comment in already carries `reactions.total_count`, and a comment
with none is the common case, so those are not asked about any more. What is left to ask
is the comments that really have reactions, since the count alone does not say who left
them and the author's own do not count.

The two halves of the method were the same code twice, once for issue comments and once
for code comments, so they are one method now with the client method and the word for the
log passed in.

The test builds two comments that GitHub reported as having no reactions and stubs no HTTP
at all. On `master` it fails with `WebMock::NetConnectNotAllowedError` for
`/issues/comments/101/reactions`.

Closes #1664


morphqdd:
Turning this into a draft: it needs zerocracy/fbe#770 first.

The `rake` job fails on `test_pull_request_event_with_comments`, which expects 4 and gets 2.
That test runs against `Fbe::FakeOctokit`, whose code comments say `reactions.total_count: 0`
while its own `pull_request_review_comment_reactions` answers with one reaction for the same
comment. So the fake disagrees with itself, and this change believes the count.

Against GitHub the count is right: `GET /pulls/{n}/comments` carries `reactions.total_count`,
and a comment with none has none.

Filed as zerocracy/fbe#770, fix in zerocracy/fbe#771. Once that is released and picked up
here, this goes green with no change.

PR #2048, opened 2026-09-07
Insert the lost-issue fact inside a transaction

### LINKED ISSUE #2047: `Jp.issue_was_lost` inserts its fact outside a transaction, so `Factbase::Rules` kills the judge
`Jp.issue_was_lost` calls `Fbe.if_absent` at `lib/issue_was_lost.rb:25-31` without a surrounding `Fbe.fb.txn`. It is the only one of the sixteen `Fbe.if_absent` call sites in this repository that does not wrap the call in a transaction; the other fifteen, in `find-all-issues.rb`, `issue-was-closed.rb`, `label-was-attached.rb` and the rest, all do.

That matters because `Fbe.if_absent` populates a freshly inserted fact one property at a time, and the `Factbase::Rules` decorator that `Fbe.fb` installs validates the fact after every single assignment unless it is inside a transaction. The block here assigns `n.issue` before `n.repository`, so the half-built fact is checked against the `basic.fe` rule `(when (exists issue) (exists repository))` and raises `ArgumentError`, which `Fbe::Iterate` re-raises as `Fbe::Error` and which ends the judge. In https://github.com/zerocracy/judges-action/actions/runs/34107167973/job/101694726828 `find-all-issues` died this way twice in one run, on `[ issue: [1566] ]`, after that issue had been deleted, so the repository was never scanned to the end. `test_find_all_issues_with_not_found_min` does exercise the 404 path but never reaches the insert, because its prior fact is fresh and the method returns at the `stale.positive?` guard on line 21.

Wrap lines 25-38 in `Fbe.fb.txn`, passing the transactional factbase to `Fbe.if_absent(fb:)`, the way every other call site in this repository already does. The underlying non-atomic insert in the gem is reported as zerocracy/fbe#797.

yegor256:
Wraps the `Fbe.if_absent` call in `Jp.issue_was_lost` in an `Fbe.fb.txn`, and adds `test_find_all_issues_with_lost_min`, which reproduces the failure.

`Fbe.if_absent` sets the properties of a new fact one at a time, and `Factbase::Rules` checks the fact after each one unless a transaction is open. This block sets `issue` before `repository`, so the fact was checked while it held nothing but `issue`, which the `(when (exists issue) (exists repository))` rule forbids. `Fbe::Iterate` turned that into an `Fbe::Error` and `find-all-issues` died on it twice in one production run, once per cycle, leaving the repository half scanned. This was the only one of the sixteen `Fbe.if_absent` call sites here that ran outside a transaction, so the fix just brings it in line with the rest.

The existing 404 test never caught it because its prior fact is fresh, so the method returns at the `stale.positive?` guard and never reaches the insert. The non-atomic insert in the gem itself is reported separately as zerocracy/fbe#797, and once that lands this wrapper stops being load-bearing, though it stays correct either way.

Closes #2047

### DIFF
--- a/lib/issue_was_lost.rb
+++ b/lib/issue_was_lost.rb
@@ -22,18 +22,20 @@ def Jp.issue_was_lost(where, repository, issue)
     $loog.info("The issue #{issue} was marked as lost, #{stale} facts marked as stale")
     return
   end
-  f =
-    Fbe.if_absent do |n|
-      n.issue = issue
-      n.what = 'issue-was-lost'
-      n.repository = repository
-      n.where = where
+  Fbe.fb.txn do |fbt|
+    f =
+      Fbe.if_absent(fb: fbt) do |n|
+        n.issue = issue
+        n.what = 'issue-was-lost'
+        n.repository = repository
+        n.where = where
+      end
+    if f.nil?
+      $loog.warn("The issue ##{issue} was already lost")
+      next
     end
-  unless f
-    $loog.warn("The issue ##{issue} was already lost")
-    return
+    f.stale = 'issue'
+    f.when = Time.now
+    $loog.info("The issue #{issue} was marked as lost")
   end
-  f.stale = 'issue'
-  f.when = Time.now
-  $loog.info("The issue #{issue} was marked as lost")
 end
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +16 points for this: +24 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+573](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2068, opened 2026-09-08
#2060: refuse a timeout or lifetime that is not a positive integer

### LINKED ISSUE #2060: entry.sh silently turns non-numeric timeout and lifetime inputs into zero
The `entry.sh` conversions at lines 233-245 multiply `INPUT_TIMEOUT` and `INPUT_LIFETIME` directly in Bash arithmetic without validating that either value is numeric. Bash treats an unrecognized value such as `abc` as zero, so `INPUT_TIMEOUT=abc` becomes `--timeout 0` and `INPUT_LIFETIME=abc` becomes `--lifetime 0` while the script continues successfully.

The action then launches `judges update` with zero-second budgets, which stops work immediately and can publish an incomplete factbase instead of reporting invalid configuration. Validate both inputs as positive integer minute counts before arithmetic expansion and fail with a clear message, matching the existing `cycles` validation.

gemshrine:
Fixes #2060

`INPUT_TIMEOUT` and `INPUT_LIFETIME` went straight into `$(( ))`, where Bash reads an unrecognized word as zero. `INPUT_TIMEOUT=abc` therefore became `--timeout 0` and the run continued, stopping every judge at once and publishing whatever factbase it had, with nothing in the log to say the input was wrong.

Both are now checked against `^[1-9][0-9]*$` before the arithmetic, refusing `abc`, `0`, `-5` and `1.5` with the same shape of message the `cycles` input already uses. The empty case is untouched: it still falls back to 10 and 15.

The test pulls each block out of `entry.sh` itself and runs it under `bash`, so what it exercises is the shipped guard rather than a copy of it. Both cases fail on master with "INPUT_TIMEOUT=\"abc\" was accepted".

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -234,12 +234,20 @@ timeout=${INPUT_TIMEOUT}
 if [ -z "${timeout}" ]; then
     timeout=10
 fi
+if ! [[ "${timeout}" =~ ^[1-9][0-9]*$ ]]; then
+    echo "INPUT_TIMEOUT must be a positive integer, got: ${timeout}" >&2
+    exit 1
+fi
 timeout=$((timeout * 60))
 echo "Each judge will spend up to ${timeout} seconds"
 
 lifetime=${INPUT_LIFETIME}
 if [ -z "${lifetime}" ]; then
     lifetime=15
+fi
+if ! [[ "${lifetime}" =~ ^[1-9][0-9]*$ ]]; then
+    echo "INPUT_LIFETIME must be a positive integer, got: ${lifetime}" >&2
+    exit 1
 fi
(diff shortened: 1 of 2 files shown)


yegor256:
Merged, thanks! Silent coercion to zero was the worst possible outcome here.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2020](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2069, opened 2026-09-08
#2058: require the cycle count to be positive

### LINKED ISSUE #2058: entry.sh accepts zero cycles although the input requires a positive count
The `cycles` input is documented and reported as a positive integer, but `entry.sh:247-256` accepts `0` because its regex only checks digits. It then passes `--max-cycles 0` to the `judges update` command. The update loop always starts cycle 1 before checking the maximum, so a caller asking for zero cycles still runs one full cycle and may perform GitHub reads and writes.

This makes the advertised cycle limit unreliable and turns a clearly invalid configuration into an unexpected run. Reject zero (and values such as `00`) in `entry.sh` before invoking `judges`, with the existing validation error, so only positive cycle counts reach `--max-cycles`.

gemshrine:
Fixes #2058

The `cycles` input is documented as a positive integer, but the guard matched `^[0-9]+$`, so `0` (and `00`) passed and reached `judges update` as `--max-cycles 0`. The update loop enters cycle 1 before it compares against the maximum, so asking for zero cycles still ran one full cycle, with the GitHub reads and writes that come with it.

The pattern is now `^[1-9][0-9]*$`, which keeps every valid count and turns `0`, `00`, `007` and `-1` into the validation error that `abc` already produced. The empty case still falls back to 2.

The test takes the block out of `entry.sh` and runs it under `bash`, so it exercises the shipped guard. It fails on master with "INPUT_CYCLES=\"0\" was accepted".

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -246,7 +246,7 @@ echo "The update will run for up to ${lifetime} seconds"
 
 cycles=${INPUT_CYCLES}
 if [ -n "${cycles}" ]; then
-    if ! [[ "${cycles}" =~ ^[0-9]+$ ]]; then
+    if ! [[ "${cycles}" =~ ^[1-9][0-9]*$ ]]; then
         echo "INPUT_CYCLES must be a positive integer, got: ${cycles}" >&2
         exit 1
     fi
--- a/test/test_entry_cycles.rb
+++ b/test/test_entry_cycles.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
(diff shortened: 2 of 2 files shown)


yegor256:
Merged, thanks! Running the guard out of `entry.sh` itself, rather than a copy of it, is what makes this test worth having.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1996](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2071, opened 2026-09-08
#2050: do not measure review time backwards from the merge

### LINKED ISSUE #2050: A pull reviewed after it was merged records a negative some_review_time
## What happens

`some_review_time` measures the distance from the earliest review to the merge:

```ruby
first = all.select { |r| r[:submitted_at] }.min_by { |r| r[:submitted_at] }
times << Integer(pr[:pull_request][:merged_at] - first[:submitted_at]) if first
```

Nothing says the earliest review came before the merge. A pull that was merged first and reviewed afterwards produces a negative number, and that number goes into `some_review_time` alongside the real ones.

## Reproduction

`some_review_time` called with one merged pull, one review, and only the review time moved:

| merged at | earliest review submitted at | `some_review_time` |
| --- | --- | --- |
| 2025-01-10 12:00 | 2025-01-10 10:00 | `[7200]` |
| 2025-01-10 12:00 | 2025-01-10 14:00 | **`[-7200]`** |
| 2025-01-10 12:00 | 2025-01-11 12:00 | **`[-86400]`** |

```ruby
found = { items: [{ id: 1, number: 10, pull_request: { merged_at: } }] }
octo = FakeOcto.new([{ id: 7, user: { id: 1 }, submitted_at: }])
Fbe.stub(:unmask_repos, ->(&b) { b.call('foo/foo') }) do
  Fbe.stub(:octo, octo) do
    Jp.stub(:qosearch, found) do
      some_review_time(fact)
    end
  end
end
```

## Why it matters

Reviewing a pull after it is merged is ordinary: a maintainer merges their own work and a colleague leaves an approving or commenting review the next morning, a bot posts a review after the merge, someone revisits a merged pull during a retrospective. GitHub allows submitting a review on a merged pull, and `pull_request_reviews` returns it like any other. When every review on a pull landed after the merge, `min_by` picks one of those, and the delta is negative.

The number is not filtered anywhere downstream: it lands in the `some_review_time` array on the `quality-of-service` fact and from there into the averages the dashboards show. One pull reviewed a day late subtracts a day from the sum, and a repository with few merged pulls in the window can end up with a negative average review time, which reads as nonsense rather than as an outlier.

The same file records `some_reviewers_per_pull` and `some_reviews_per_pull` from `all`, which is fine, so the defect is confined to the one duration.

`some_triage_time`, in the file next door, has exactly this guard:

```ruby
threshold = Fbe.pmp.quality.qos_min_triage_seconds.value || 60
...
delta = ff.when - issue[:created_at]
next if delta < threshold
```

so the idea that a computed duration needs a floor before it is recorded is already in the codebase; `some_review_time` just does not do it.

## Suggested fix

Take the earliest review that actually preceded the merge, and skip the pull when there is none:

```ruby
merged = pr[:pull_request][:merged_at]
first = all.select { |r| r[:submitted_at] && r[:submitted_at] <= merged }.min_by { |r| r[:submitted_at] }
times << Integer(merged - first[:submitted_at]) if first
```

A pull merged without any prior review has no review time to report, which is different from having a negative one.

If a floor like the triage one is wanted instead, `qos_min_review_seconds` alongside `qos_min_triage_seconds` would fit the existing shape, but the sign check is needed either way: a threshold of 60 seconds still lets `-86400` through only if the comparison is `abs`, and it should not be.

gemshrine:
Fixes #2050

`some_review_time` subtracted the earliest review from the merge without asking which came first. A pull merged before anyone reviewed it produced a negative number, and that number went into the metric next to the real ones, dragging the average down by however long the late review took.

A review submitted after the merge does not measure the time a review took, so the pull no longer contributes a review time, and the log says why. It still contributes to `some_review_size`, `some_reviewers_per_pull` and `some_reviews_per_pull`, which are counts and stay correct either way.

The test drives `some_review_time` through `Jp::FakeGithub` with the merge fixed at 12:00 and only the review time moved. On master the two late cases come back as `[-7200]` and `[-86400]`.

### DIFF
--- a/judges/quality-of-service/some_review_time.rb
+++ b/judges/quality-of-service/some_review_time.rb
@@ -35,9 +35,7 @@ def some_review_time(fact)
       if first
         seconds = Integer(pr[:pull_request][:merged_at] - first[:submitted_at])
         if seconds.negative?
-          $loog.info(
-            "The pull ##{pr[:number]} in #{repo} was first reviewed after it was merged, "             'its review time is not measurable'
-          )
+          $loog.info("The pull ##{pr[:number]} in #{repo} was reviewed after it was merged")
         else
           times << seconds
         end
--- a/judges/quality-of-service/some_review_time.rb
+++ b/judges/quality-of-service/some_review_time.rb
@@ -32,7 +32,16 @@ def some_review_time(fact)
           next
         end
       first = all.select { |r| r[:submitted_at] }.min_by { |r| r[:submitted_at] }
-      times << Integer(pr[:pull_request][:merged_at] - first[:submitted_at]) if first
+      if first
+        seconds = Integer(pr[:pull_request][:merged_at] - first[:submitted_at])
+        if seconds.negative?
+          $loog.info(
+            "The pull ##{pr[:number]} in #{repo} was first reviewed after it was merged, "             'its review time is not measurable'
+          )
+        else
+          times << seconds
+        end
+      end
       sizes << csize
       users = all.map { |r| r.dig(:user, :id) }
       users.uniq!
--- a/test/judges/test-some-review-time.rb
+++ b/test/judges/test-some-review-time.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+require 'factbase'
+require 'judges/options'
+require 'loog'
+require_relative '../fake_github'
+require_relative '../test__helper'
+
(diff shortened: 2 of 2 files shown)


yegor256:
Merged, thanks! Leaving the counts intact while dropping the negative duration is the correct trade-off.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1952](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #2054, opened 2026-09-07
who-is-alive re-checks every user on every cycle, and shares one when field with who-has-name


gemshrine:
## What happens

`who-is-alive` picks `who-has-name` facts whose `when` is more than two days old, checks whether the GitHub user still exists, and, when the user is alive, moves on:

```ruby
Fbe.consider(
  "(and
    (eq what 'who-has-name')
    (lt when (minus (to_time '#{Jp.today.utc.iso8601}') '2 days'))
    ...
) do |f|
  ...
  unless nick.nil?
    $loog.debug("GitHub user @#{nick} (##{f.who}) is alive")
    next
  end
```

Nothing writes `when` back, so the fact stays past the two-day line and is checked again on the next cycle, and every cycle after that.

## Reproduction

One `who-has-name` fact with `when` set 30 days back, `who-is-alive` loaded three times:

```
when unchanged: true
lookups over 3 runs: 3
```

## The deeper half: two judges share one clock

`who-has-name` uses the very same `when` field of the very same fact as its own rate limiter, with a different period:

```ruby
# judges/who-has-name/who-has-name.rb:66
(lt when (minus (to_time '#{Jp.today.utc.iso8601}') '5 days'))
```

so `when` on a `who-has-name` fact is being read as "checked for a new login less than 5 days ago" and as "checked for existence less than 2 days ago" at once. Those cannot both be true of one timestamp. Whichever judge starts writing it back first will silently reset the other one's window:

- if `who-has-name` refreshes `when` on its 5-day pass, a fact it just touched looks fresh to `who-is-alive` too, and the existence check is skipped for the next two days regardless of when it actually last ran;
- between day 2 and day 5 the fact is stale for `who-is-alive` and fresh for `who-has-name`, so the existence check runs on every cycle again.

This matters for the fix, which is why it is worth stating here rather than only in #2053: writing `when` back in one judge alone does not give either judge the period it asks for.

## Why it matters

Today both judges look every known user up on every cycle: two `GET /user/{id}` per user per cycle, on the same user, for data that changes rarely. A repository the action has watched for a while carries hundreds of `who-has-name` facts, all of them permanently past both lines. `Fbe.octo` refuses calls with `Fbe::OffQuota` once the core quota drops below the threshold, so the bill is paid by whichever judge runs next.

## Suggested fix

Give each judge its own marker, so the two periods stop fighting over one field:

```ruby
# who-is-alive, after a successful lookup
[...]

PR #2072, opened 2026-09-08
#2053: move the refresh line forward when the name is re-read

### LINKED ISSUE #2053: who-has-name never updates when, so its five-day refresh runs on every cycle
## What happens

The second half of `who-has-name` re-reads the nickname of every user whose fact is older than five days:

```ruby
Fbe.consider(
  "(and
    (eq what 'who-has-name')
    (lt when (minus (to_time '#{Jp.today.utc.iso8601}') '5 days'))
    ...
) do |f|
  nick = ... Jp.nick_of(f.who) ...
  alive << f.who
  Fbe.overwrite(f, 'name', nick)
end
```

`when` is what decides whether the fact is due for a refresh, and nothing writes it back. `Fbe.overwrite` recreates the fact from its existing properties, so the old `when` survives untouched. The fact therefore satisfies `(lt when ...)` on the next cycle, and the one after that, forever.

The intent is plainly a five-day refresh: that is what the `5 days` in the query is for. What actually happens is a lookup of every known user on every cycle.

## Reproduction

One `who-has-name` fact, its `when` set 30 days back, the judge loaded three times:

```ruby
f.what = 'who-has-name'
f.who = 444
f.name = 'lebowski'
f.when = Time.now - (30 * 24 * 60 * 60)
3.times { load_it('who-has-name', fb) }
```

```
when before:       2026-08-08T13:43:43Z
when after 3 runs: 2026-08-08T13:43:43Z
user lookups: 3
```

The same test with `when` set one day back, so the fact is not due:

```
control lookups (when is 1 day old): 0
```

So the query works; it is the write-back that is missing. A fact that once crosses the five-day line never leaves the set.

## Why it matters

It is one `GET /user/{id}` per known user per cycle, on a resource where nothing changes from one cycle to the next. A repository the action has been watching for a while has hundreds of `who-has-name` facts, all of them permanently past the five-day mark, so the judge spends hundreds of core-quota calls every cycle on data it already has. `Fbe.octo` starts refusing calls with `Fbe::OffQuota` below the threshold, so the cost lands on whatever judge happens to run after this one.

The nickname does need re-checking, because logins change, which is what `test_overwrite_name_if_user_login_changed` covers. Five days is the rate the code asks for and does not get.

## Suggested fix

Write `when` back after a successful lookup, so the fact leaves the due set:

```ruby
if nick.nil?
  f.stale = 'who'
  next
end
alive << f.who
Fbe.overwrite(f, 'name', nick)
Fbe.overwrite(f, 'when', Jp.today)
```

`Fbe.overwrite` recreates the fact, so the second call needs the handle the first one leaves behind; setting both in one call is cleaner:

```ruby
Fbe.overwrite(f, { 'name' => nick, 'when' => Jp.today })
```

Worth a test that runs the judge twice against a stale fact and asserts a single lookup, since nothing in the suite currently notices the difference between "every five days" and "every cycle".

gemshrine:
Fixes #2053

The second pass of `who-has-name` selects facts whose `when` is more than five days old and re-reads the nickname, but nothing wrote `when` back. `Fbe.overwrite` rebuilds the fact from the properties it already carries, so the old `when` survived and the fact matched `(lt when ...)` again on the very next cycle. The five-day refresh was therefore a lookup of every known user on every cycle: one `GET /user/{id}` per user per run, for data that changes rarely, paid out of the same core quota every other judge shares.

`name` and `when` are now written together in one `Fbe.overwrite` call, so the fact is rebuilt once and lands five days past the line, which is what the `5 days` in the query asks for.

The test loads the judge with one fact dated five days back and checks that `when` moved to the current time. On master it stays at 2025-06-18 and the assertion reports that the next cycle would look the same user up again.

`who-is-alive` reads the same `when` on the same facts with a two-day period of its own; that half is #2054, and it needs a marker of its own rather than this one.

### DIFF
--- a/judges/who-has-name/who-has-name.rb
+++ b/judges/who-has-name/who-has-name.rb
@@ -92,7 +92,7 @@ Fbe.consider(
     next
   end
   alive << f.who
-  Fbe.overwrite(f, 'name', nick)
+  Fbe.overwrite(f, { 'name' => nick, 'when' => Time.now })
 end
 
 alive.uniq!
--- a/test/judges/test-who-has-name.rb
+++ b/test/judges/test-who-has-name.rb
@@ -87,6 +87,27 @@ class TestWhoHasName < Jp::Test
     end
   end
 
+  def test_moves_the_refresh_line_forward_after_a_lookup
+    WebMock.disable_net_connect!
+    rate_limit_up
+    stub_github('https://api.github.com/user/12', body: { login: 'user22', id: 12, type: 'User' })
+    fb = Factbase.new
+    fb.with(
+      _id: 3, what: 'who-has-name', where: 'github', who: 12, name: 'user2',
+      when: Time.parse('2025-06-18 20:00:00 UTC')
+    )
+    now = Time.parse('2025-06-23 22:00:00 UTC')
+    Time.stub(:now, now) do
+      load_it('who-has-name', fb)
+    end
+    fact = fb.query('(eq who 12)').each.first
+    assert_equal('user22', fact['name'].first)
+    assert_operator(
+      fact['when'].first, :>, now - 60,
+      'the fact stayed past the five-day line, so the next cycle looks the same user up again'
+    )
+  end
+
   def test_keeps_fact_active_on_forbidden_user_lookup
     WebMock.disable_net_connect!
     rate_limit_up


yegor256:
Merged, thanks! Writing `name` and `when` in one `Fbe.overwrite` is the right shape.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (23) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1928](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2089, opened 2026-09-08
#2061: keep the event actor when the release author is gone

### LINKED ISSUE #2061: ReleaseEvent with a deleted release author crashes github-events
## What happens

`github-events` dereferences the release author without checking for a missing account. In `judges/github-events/github-events.rb:356-364`, a published `ReleaseEvent` executes:

```ruby
if fact.all_properties.include?('who') && fact.who != json[:payload][:release][:author][:id]
  Fbe.overwrite(fact, 'who', json[:payload][:release][:author][:id])
end
```

GitHub can return `release.author: null` when the account that published the release has been deleted. The expression then raises `NoMethodError: undefined method '[]' for nil`, aborting the event scan and leaving later events unprocessed.

## Reproduction

Feed `github-events` a published release event whose actor is present but whose release author is absent:

```ruby
json[:payload][:release][:author] = nil
json[:payload][:action] = 'published'
json[:actor] = { id: 123, login: 'release-bot' }
```

The event reaches the `ReleaseEvent` branch and fails at `release[:author][:id]`. Existing release tests always provide an author, so this payload shape is not covered.

## Suggested fix

Treat a missing release author like the other missing GitHub actors: keep the event fact and mark it `stale = 'who'` (or retain the event actor when that is the intended identity) instead of dereferencing `nil`. Add a regression test with `author: nil` and verify that subsequent events are still scanned.

opusstudiohq-max:
Closes #2061.

GitHub returns `release.author: null` once the account that published a release has been deleted. `github-events` dereferenced it twice while deciding whether to overwrite `who`:

```ruby
if fact.all_properties.include?('who') && fact.who != json[:payload][:release][:author][:id]
```

so the whole event scan died with `NoMethodError: undefined method '[]' for nil` and every event queued behind the ghost release went unprocessed.

The overwrite now happens only when the release actually names an author. When it does not, the fact keeps the actor `fact.who` was already set from at line 159 — the identity the event itself reports — so nothing downstream loses a `who` and the `Fbe.who` calls in the same branch still have something to read. That is the second option the issue suggests, and it needs no `stale` marker because the fact is not missing an author at all.

### Verification

The regression test is a published `ReleaseEvent` with `author: nil` and an actor present, as in the issue. I checked it actually catches the defect rather than passing either way:

- with this change: `test_release_event_survives_a_deleted_release_author PASS`, suite green
- with the change reverted and the test kept: `ERROR — Fbe::Error: Failure in repository #42 at #0: undefined method '[]' for nil`

It also asserts the event actor is kept as `who`, so a later change cannot quietly drop the identity instead of preserving it.

`bundle exec rake` is green on a fork before opening this.

### DIFF
--- a/judges/github-events/github-events.rb
+++ b/judges/github-events/github-events.rb
@@ -359,8 +359,9 @@ Fbe.iterate do
       case json[:payload][:action]
       when 'published'
         fact.what = 'release-published'
-        if fact.all_properties.include?('who') && fact.who != json[:payload][:release][:author][:id]
-          Fbe.overwrite(fact, 'who', json[:payload][:release][:author][:id])
+        author = json[:payload][:release].dig(:author, :id)
+        if author && fact.all_properties.include?('who') && fact.who != author
+          Fbe.overwrite(fact, 'who', author)
         end
         contributors(fact, rname).each { |c| fact.contributors = c }
         Jp.fill_fact_by_hash(fact, info(fact, rname))
--- a/test/judges/test-github-events.rb
+++ b/test/judges/test-github-events.rb
@@ -2711,6 +2711,60 @@ class TestGithubEvents < Jp::Test
     assert_equal([526_301], f.first[:contributors])
   end
 
+  def test_release_event_survives_a_deleted_release_author
+    WebMock.disable_net_connect!
+    rate_limit_up
+    stub_event(
+      {
+        id: '101',
+        type: 'ReleaseEvent',
+        actor: {
+          id: 8_086_956,
+          login: 'rultor',
+          display_login: 'rultor'
+        },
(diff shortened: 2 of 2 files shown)


yegor256:
Merged, thanks! The guard on the release author is minimal and the regression test pins the behaviour down nicely.

---

0crat:
@opusstudiohq-max Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+44](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2104, opened 2026-09-09
Move is-human-or-robot test onto the fake GitHub server

### LINKED ISSUE #2041: fake_github.rb:12-20: Move the rest of the tests onto...
The puzzle `2028-2686cab1` from #2028 has to be resolved: 

https://github.com/zerocracy/judges-action/blob/6fa86d16a33337aae9b0c67c73297606e910b2b9/test/fake_github.rb#L12-L20

The puzzle was created by @Thayorns on 06-Sep-26. 

Estimate: 60 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

Thayorns:
The `is-human-or-robot` test now declares its GitHub routes through `Jp::FakeGithub` and talks to a real socket on localhost, instead of building four separate WebMock stub sets with `stub_github` and `rate_limit_up`. Each of the four tests carries its own route table: the rate limit, the four user lookups, the two 403 replies and the 404 for the vanished user.

The puzzle in `test/fake_github.rb` moves forward with it. Twenty six files are left on `stub_request`, and the count of stubs in the note was stale, so it now reads the real one: one thousand two hundred and thirty six.

`require 'octokit'` went out of the test file, since nothing there names the constant any more once the stubbing is gone.

Closes #2041

### DIFF
--- a/test/fake_github.rb
+++ b/test/fake_github.rb
@@ -9,9 +9,9 @@ require 'socket'
 require_relative '../lib/jp'
 
 # rubocop:disable Elegant/NoComments
-# @todo #2028:60min Move the rest of the tests onto this server.
-#  Twenty seven files under `test/` still build their GitHub with
-#  `stub_request` and `stub_github`, one thousand one hundred and seventy nine
+# @todo #2041:60min Move the rest of the tests onto this server.
+#  Twenty six files under `test/` still build their GitHub with
+#  `stub_request` and `stub_github`, one thousand two hundred and thirty six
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 take a look at this one please

PR #2115, opened 2026-09-10
fix(#2109): wrap the stale marking of dead repositories in a transaction

### LINKED ISSUE #2109: `erase-repository` marks facts stale outside a transaction
The second loop in `judges/erase-repository/erase-repository.rb:34` walks every repository confirmed missing and writes `stale = 'repository'` at line 37 straight onto the factbase. The two judges that do the same job for users both wrap the equivalent writes in `Fbe.fb.txn` — `judges/who-is-alive/who-is-alive.rb:50` and `judges/eliminate-ghosts/eliminate-ghosts.rb:49`.

`Factbase#txn` in `factbase.rb:170` builds a `LazyTaped` over the maps and only applies it once the block returns cleanly, so an exception partway through leaves the factbase as it was. Without it, an error raised while marking the fifth of ten dead repositories commits the first four and abandons the rest, and the run that follows sees a factbase that is neither the before state nor the after state. Marking is idempotent enough that the next cycle recovers, but the two sibling judges get an atomic result from the same operation and this one does not.

Wrapping lines 34 to 39 in `Fbe.fb.txn do |fbt| ... end` and querying through `fbt` matches what `eliminate-ghosts` already does.

@yegor256 please take a look.

Thayorns:
The second loop of `erase-repository` wrote `stale = 'repository'` straight onto the
factbase, one fact at a time. An error raised halfway through the dead repositories
committed the marks made so far and abandoned the rest, so the next run started from
a factbase that was neither the state before nor the state after.

The loop now runs inside `Fbe.fb.txn` and queries through the transaction handle, the
same way `eliminate-ghosts` and `who-is-alive` already mark stale users. Either every
fact of every missing repository gets its flag, or none of them does.

Three tests come with it: one that fails unless the marking happens in a transaction,
one that checks all facts of several dead repositories are marked together, and one
that keeps a repository answering GitHub in the same run untouched.

Closes #2109

### DIFF
--- a/judges/erase-repository/erase-repository.rb
+++ b/judges/erase-repository/erase-repository.rb
@@ -5,6 +5,7 @@
 
 require 'elapsed'
 require 'fbe/consider'
+require 'fbe/fb'
 require 'fbe/octo'
 require 'logger'
 require 'octokit'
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 ping

PR #2116, opened 2026-09-10
fix(#2110): stop scanning repositories when github quota is exhausted

### LINKED ISSUE #2110: Rate-limit errors are caught by the `Forbidden` arm in `erase-repository.rb`
`Octokit::TooManyRequests` is a subclass of `Octokit::Forbidden` (`octokit/error.rb:271` in octokit 10.0.0), so the `rescue Octokit::Forbidden` at `judges/erase-repository/erase-repository.rb:25` catches a 403 that means the rate limit is exhausted as well as one that means the token cannot see the repository. Both land on the same two lines: a warning saying `Access forbidden to GitHub repository #N (transient, will retry next cycle)` and `good[r] = true` at line 30.

That misreports the cause to whoever reads the log, and it leaves the loop running. After GitHub has said to stop, the judge goes back to line 15 and keeps walking facts, spending whatever allowance is left on calls that cannot succeed. The `good[r] = true` also records the repository as verified for the rest of the cycle, which is right for a genuine permission failure and wrong for a repository nobody actually checked. `eliminate-ghosts` keeps `good`, `bad` and `forbidden` as three separate sets (lines 12 to 14) rather than folding two meanings into one value.

Rescuing `Octokit::TooManyRequests` ahead of `Octokit::Forbidden` and breaking out of the loop covers it. The same defect is already accepted for other judges in #1936, #1941 and #1978; none of them touches this file.

@yegor256 please take a look.

Thayorns:
`Octokit::TooManyRequests` is a subclass of `Octokit::Forbidden`, so a 403 that means the quota is gone used to land in the same arm as a 403 that means the token cannot see the repository. The log then blamed access rights, the repository was recorded as verified for the rest of the cycle, and the judge went back to the query and kept spending whatever allowance was left on calls that could not succeed.

The rate limit now has its own arm, placed ahead of the `Forbidden` one, saying which repository was being read when GitHub cut us off. It records nothing about that repository, since nobody actually checked it, and it stops the scan. Repositories already found missing before that point are still marked stale, so the work done earlier in the cycle is not thrown away.

Two tests cover it: one proves no further repository is read after the rate limit, the other proves a repository found missing before it still gets its facts marked stale.

Closes #2110

### DIFF
--- a/judges/erase-repository/erase-repository.rb
+++ b/judges/erase-repository/erase-repository.rb
@@ -10,8 +10,10 @@ require 'logger'
 require 'octokit'
 
 good = {}
+quota = true
 
 Fbe.fb.query('(and (eq where "github") (exists repository) (absent stale))').each do |f|
+  break unless quota
   next if Fbe.octo.off_quota?
   r = f.repository
   next unless good[r].nil?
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 ping

Thayorns:
@yegor256 ping

Thayorns:
@yegor256 addressed, please check

PR #2117, opened 2026-09-10
Stop scanning facts once the quota is gone

### LINKED ISSUE #2107: `erase-repository` keeps scanning facts after the quota is gone
Line 15 of `judges/erase-repository/erase-repository.rb` reads `next if Fbe.octo.off_quota?`. Once the quota is exhausted that condition is true for every remaining fact, and `next` walks all of them anyway. The judge cannot do any further work at that point — every repository lookup it would make is guaranteed to be skipped — but it still pays for the discovery each time, because `off_quota?` in `fbe/octo.rb:170` calls `@origin.rate_limit!`, which is a live `get 'rate_limit'` in `octokit/client/rate_limit.rb:35`.

With the factbase size shown in #700, hitting the wall on the first fact means 7684 more `/rate_limit` round trips to learn the same answer 7684 times. #1978 makes the same argument for the `dimensions-of-terrain` metrics: quota exhaustion is a stop signal for the whole judge, not a per-item skip.

`break if Fbe.octo.off_quota?` is the whole fix. The second loop at line 34 still runs and still marks whatever was confirmed missing before the quota ran out, so nothing is lost by leaving early.

@yegor256 please take a look.

Thayorns:
When the quota runs out, `erase-repository` has nothing left to do: every
repository lookup it would make from that point on is guaranteed to be skipped.
The loop said `next`, so it kept walking the remaining facts and paid for a live
`/rate_limit` round trip on each one, only to learn the same answer again. On a
factbase of the size we run against, that is thousands of requests spent to
confirm the wall is still there. It now says `break`.

Nothing is lost by leaving early. The second loop still runs and still marks
every repository that was confirmed missing before the quota ran out, so the
work already done in this cycle is written down; whatever was left unchecked is
picked up on the next run.

Three tests cover it: the quota is asked about once and not once per fact, a
repository that died before the wall is still retired, and a repository nobody
managed to ask about is left alone.

Closes #2107

### DIFF
--- a/judges/erase-repository/erase-repository.rb
+++ b/judges/erase-repository/erase-repository.rb
@@ -12,7 +12,7 @@ require 'octokit'
 good = {}
 
 Fbe.fb.query('(and (eq where "github") (exists repository) (absent stale))').each do |f|
-  next if Fbe.octo.off_quota?
+  break if Fbe.octo.off_quota?
   r = f.repository
   next unless good[r].nil?
   elapsed($loog, level: Logger::INFO) do
--- a/test/judges/test-erase-repository.rb
+++ b/test/judges/test-erase-repository.rb
@@ -4,9 +4,13 @@
 # SPDX-License-Identifier: MIT
 
 require 'factbase'
+require 'fbe/octo'
+require 'octokit'
 require_relative '../test__helper'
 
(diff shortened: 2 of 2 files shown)


Thayorns:
@yegor256 please take a look, happy to clarify anything here.

PR #2192, opened 2026-09-10
Write the lost issue through the enclosing transaction

### LINKED ISSUE #2074: `Jp.issue_was_lost` runs its own transaction inside an open one, so its writes escape the outer rollback
`lib/issue_was_lost.rb:25` opens its own `Fbe.fb.txn`, added when #2047 was fixed. Four of the call sites are already inside an open `Fbe.fb.txn` when they reach it: `judges/find-all-issues/find-all-issues.rb:104` (transaction opened at :85), `judges/find-earliest-issue/find-earliest-issue.rb:69` (:52), `judges/find-latest-issue/find-latest-issue.rb:69` (:52) and `judges/find-missing-issues/find-missing-issues.rb:85` (:68). The two writes before it, the `stale` flags on `lib/issue_was_lost.rb:19` and `Fbe::Tombstone.new.bury!` on `:20`, also go through the global `Fbe.fb` rather than through the enclosing transaction.

Calling `Fbe.fb.txn` from inside an `Fbe.fb.txn` block is not refused. The guard that raises `You cannot start a transaction inside another transaction` lives in `Factbase::Light#txn`, and it is only reached by calling `txn` on the factbase a transaction yields; `Fbe.fb` is the outer chain, whose innermost `Factbase#txn` simply opens a second, independent journal over the same maps. So the inner transaction commits on its own, and the `stale` flags and the tombstone land immediately. When the enclosing block then fails, and in these four judges it can — an unrescued `Octokit` error, or the `Factbase::Rules` check that runs at commit — the outer transaction rolls back while the `issue-was-lost` fact, the `stale` flags and the tombstone stay behind. The issue is left buried and marked lost although the `issue-was-opened` fact that triggered all of it is gone, and no later cycle will look at that issue again.

Give `Jp.issue_was_lost` an `fb:` keyword defaulting to `Fbe.fb`, use it for the query on `:12`, for `Fbe::Tombstone.new(fb:)` on `:20` and in place of the `Fbe.fb.txn` on `:25`, and pass `fbt` at the four call sites that are already inside a transaction.

Thayorns:
`Jp.issue_was_lost` opened its own `Fbe.fb.txn`. Four judges reach it from inside an already open transaction, and because `Fbe.fb` is the outer chain rather than the factbase a transaction yields, that inner transaction committed on its own. When the enclosing block then failed, the outer transaction rolled back but the `issue-was-lost` fact stayed behind, leaving the issue marked lost with no `issue-was-opened` fact to explain it.

The function now takes an `fb:` keyword defaulting to `Fbe.fb`, uses it for the stale query and for `Fbe.if_absent`, and no longer opens a transaction of its own. `find-all-issues`, `find-earliest-issue`, `find-latest-issue` and `find-missing-issues` pass their `fbt`, so everything they write rolls back together.

Dropping the transaction meant the fresh fact was built outside one, and `Factbase::Rules` checks after every property, so `(when (exists issue) (exists repository))` rejected the half-built fact. Setting `where`, then `repository`, then `issue` keeps it valid at every step, which is what the transaction was hiding.

The tombstone still goes through the global factbase. Routing it through the given `fb` needs `Fbe.overwrite` to stop opening a transaction of its own, which is zerocracy/fbe#843, so that half is left as a puzzle.

Closes #2074

### DIFF
--- a/judges/find-all-issues/find-all-issues.rb
+++ b/judges/find-all-issues/find-all-issues.rb
@@ -101,7 +101,7 @@ require_relative '../../lib/qos_search'
                   Fbe.octo.pull_request(repo, f.issue).dig(:head, :ref)
                 rescue Octokit::NotFound, Octokit::Deprecated => e
                   $loog.info("The pull ##{f.issue} doesn't exist in #{repo}: #{e.message}")
-                  Jp.issue_was_lost(f.where, f.repository, f.issue)
+                  Jp.issue_was_lost(f.where, f.repository, f.issue, fb: fbt)
                   next
                 rescue Octokit::Forbidden => e
                   $loog.warn(
--- a/judges/find-earliest-issue/find-earliest-issue.rb
+++ b/judges/find-earliest-issue/find-earliest-issue.rb
@@ -66,7 +66,7 @@ Fbe.iterate do
             Fbe.octo.pull_request(repo, f.issue).dig(:head, :ref)
           rescue Octokit::NotFound, Octokit::Deprecated => e
             $loog.info("The pull ##{f.issue} doesn't exist in #{repo}: #{e.message}")
-            Jp.issue_was_lost(f.where, f.repository, f.issue)
+            Jp.issue_was_lost(f.where, f.repository, f.issue, fb: fbt)
             next i
           rescue Octokit::Forbidden => e
             $loog.warn(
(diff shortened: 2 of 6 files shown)


Thayorns:
@yegor256 please take a look, happy to clarify anything here.

Thayorns:
@yegor256 ping

PR #2193, opened 2026-09-11
#2100: evaluate the date in the job summary heredoc

### LINKED ISSUE #2100: Job summary prints the shell date expression literally
## The bug
The `Job Summary` step in `.github/workflows/zerocracy.yml` writes its heredoc with a quoted `EOF` delimiter. The `Timestamp` row contains `$(date -u)`, but quoted heredocs disable command substitution, so every run publishes the literal text `$(date -u)` instead of the execution timestamp.

## Expected
Evaluate the timestamp before writing the summary (or use an unquoted delimiter only for that value), so the job summary contains the actual UTC time.

Thayorns:
The Job Summary step in `.github/workflows/zerocracy.yml` wrote its heredoc with a quoted `EOF` delimiter, which disables command substitution, so every run published the literal text `$(date -u)` in the Timestamp row instead of the actual time.

The delimiter is now unquoted, so bash evaluates the date when it writes the summary. The other values in the heredoc are GitHub expressions that are substituted before the shell runs and carry no dollar signs or backticks, so nothing else in the summary changes.

A new test in `test/test_workflow_config.rb` takes the Job Summary script from the workflow, replaces the GitHub expressions with placeholders, runs it under bash in a temporary directory, and checks that the Timestamp row holds an evaluated time.

Closes #2100

### DIFF
--- a/.github/workflows/zerocracy.yml
+++ b/.github/workflows/zerocracy.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Job Summary
         if: always()
         run: |
-          cat > "$GITHUB_STEP_SUMMARY" <<'EOF'
+          cat > "$GITHUB_STEP_SUMMARY" <<EOF
           ## zerocracy run
 
           | Info | |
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 this makes the job summary show the real run time instead of the literal date expression; happy to clarify anything.

Thayorns:
@yegor256 updated, take a look

PR #2194, opened 2026-09-11
#2098: treat an empty sqlite_cache_min_age option as absent

### LINKED ISSUE #2098: An empty sqlite_cache_min_age option suppresses the safe default
## The bug
`entry.sh` treats any `--option=sqlite_cache_min_age=*` entry as configured, including `--option=sqlite_cache_min_age=` (`entry.sh:194-202`). It therefore skips the default `3600` value. When SQLite caching is enabled, `Fbe::Octo` converts that empty value with `Integer('')`, so the action aborts instead of using the documented default.

## Reproduction
Run the action with SQLite cache enabled and an options input containing `--option=sqlite_cache_min_age=`. The judge setup raises `ArgumentError: invalid value for Integer(): \"\"` before processing the factbase.

## Expected
An empty option should be treated as absent (and receive the 3600-second default), or be rejected with a clear input-validation error before the judge starts.

Thayorns:
entry.sh treated any `--option=sqlite_cache_min_age=` entry as configured, including one with an empty value, so it skipped the 3600-second default. With SQLite caching on, Fbe::Octo then called `Integer('')` and the action aborted before processing the factbase.

The block that looks for the option now drops an entry whose value is empty and keeps every other option in its original order. An empty value is therefore treated as absent and receives the default, while a configured value stays untouched.

A new test extracts that block from entry.sh, as the cycles and minutes tests already do, and checks the empty, configured and absent cases.

Closes #2098

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -192,12 +192,17 @@ if [ "${bots_found}" == "false" ]; then
 fi
 
 cache_min_age=false
+declare -a kept=()
 for opt in "${options[@]}"; do
+    if [[ "${opt}" == "--option=sqlite_cache_min_age=" ]]; then
+        continue
+    fi
     if [[ "${opt}" == "--option=sqlite_cache_min_age="* ]]; then
         cache_min_age=true
-        break
     fi
+    kept+=("${opt}")
 done
+options=("${kept[@]}")
 if [ "${cache_min_age}" == "false" ]; then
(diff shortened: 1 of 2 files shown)


Thayorns:

Thayorns:
@yegor256 ping

PR #2195, opened 2026-09-11
#2097: bury the issue before marking its facts stale

### LINKED ISSUE #2097: issue_was_lost leaves facts stale when tombstone creation fails
`Jp.issue_was_lost` marks every matching fact stale before it calls `Fbe::Tombstone.new.bury!` (`lib/issue_was_lost.rb:8-11`). Those property writes are outside a transaction. If the tombstone API call then fails (for example because of a transient GitHub/Baza error), the stale updates remain committed while no tombstone exists.

The next run filters out the now-stale facts, sees a positive stale count, and returns at `:13-17`, so it never retries the missing tombstone. The issue is permanently recorded as lost without the tombstone needed by later judges.

Make the stale transition and tombstone operation atomic from the judge's perspective, or only mark facts stale after a successful tombstone. Add a failure-path test that retries after `bury!` raises and verifies the tombstone is eventually created.

Thayorns:
Jp.issue_was_lost used to mark every matching fact stale before calling Fbe::Tombstone#bury!. When bury! raised, those stale marks stayed in the factbase while no tombstone existed.

The tombstone is now written first, so a failure in bury! leaves the facts untouched and the next run repeats the whole transition. This also stops a retry from inserting a spurious issue-was-lost fact next to the facts that were already marked stale.

Three failure-path tests cover it: facts stay fresh when bury! raises, the issue is buried on the retry, and the retry inserts no extra lost fact.

Closes #2097

### DIFF
--- a/lib/issue_was_lost.rb
+++ b/lib/issue_was_lost.rb
@@ -8,6 +8,7 @@ require 'fbe/tombstone'
 require_relative 'jp'
 
 def Jp.issue_was_lost(where, repository, issue)
+  Fbe::Tombstone.new.bury!(where, repository, issue)
   stale =
     Fbe.fb.query(
       "(and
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 take a look please, happy to clarify anything about the new ordering in issue_was_lost.

PR #2196, opened 2026-09-11
fix(#2096): ignore an empty github_token option in favor of the github-token input

### LINKED ISSUE #2096: An empty github_token option suppresses the valid action token
`entry.sh` treats any `github_token` option as present when its serialized value starts with `--option=github_token=` (`entry.sh:155-159`). That also matches an explicitly empty value. In that case `github_token_found` becomes true and the fallback from `INPUT_GITHUB-TOKEN` at `entry.sh:165-171` is skipped, so a valid action token is never passed to the judges.

For example, with `INPUT_OPTIONS='github_token='` and `INPUT_GITHUB-TOKEN=ghp_valid`, the script prints that the option is set and invokes the judges with an empty token. API calls then fail even though the documented action input is available.

Only a non-empty option value should suppress the environment fallback; otherwise an empty option should be rejected or treated as absent. Add an entry test covering this precedence.

Thayorns:
When `options` carried `github_token=` with nothing after the equals sign, `entry.sh` still treated the token as provided. It skipped the fallback to the `github-token` action input and ran the judges with an empty token, so every API call failed even though a valid token was right there.

The options loop now drops a `github_token` entry whose value is empty and says so in the log, so the empty option counts as absent and the action input takes over. A non-empty `github_token` option still wins over the input, as before.

Two entry tests pin both sides of that precedence: one feeds an empty option next to a valid input and expects the input to be used, the other feeds a non-empty option and expects it to be kept.

Closes #2096

### DIFF
--- a/entries/ignores-an-empty-github-token-option.sh
+++ b/entries/ignores-an-empty-github-token-option.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+set -e -o pipefail
+
+SELF=$1
+
+source "${SELF}/makes/setup-test-env.sh"
+source "${SELF}/makes/test-common.sh"
+setup_test_env "${SELF}" name
+
+run_entry_script "${SELF}" success \
+  "GITHUB_WORKSPACE=$(pwd)" \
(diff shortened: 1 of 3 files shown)


Thayorns:
@yegor256 take a look please, happy to clarify anything about the token precedence here.

Thayorns:
@yegor256 ping

PR #2197, opened 2026-09-11
Publish the run report to the GitHub step summary

### LINKED ISSUE #2086: entry.sh writes summary.md after pushing the factbase, but never publishes it or exposes it to Actions
## Problem
At the end of entry.sh the action writes a Markdown report to `${GITHUB_WORKSPACE}/summary.md`. The file is created after the push command and is not copied to the Actions step summary, uploaded as an artifact, or committed.

## Why this is a bug
The generated run metrics are effectively discarded when the container exits: the factbase was already pushed, and the workspace is not persisted automatically between jobs. Users cannot see the report the script claims to generate, so the code adds I/O without delivering any result and can leave an untracked file in local runs.

## Expected behavior
Write the report to `${GITHUB_STEP_SUMMARY}` (or explicitly upload/commit it), or remove the dead generation. Add an integration test asserting the summary is visible after a run.

Thayorns:
entry.sh built a small metrics table at the end of each run and dropped it into the workspace as summary.md, and only when the run title was also being updated. Nobody using the action ever saw it, apart from our own zerocracy workflow, which picked the file up by hand.

The table now gets appended to $GITHUB_STEP_SUMMARY in a separate block of its own, so it shows up on the run page for every repository that uses the action. Anything an earlier step already wrote there is kept. Since the file is gone, the zerocracy workflow no longer tries to read it.

A new entry, publishes-the-step-summary.sh, runs the script with a step summary that already has some text in it. It checks that the report ends up there, that the earlier text is still there, and that no summary.md is left in the workspace.

One thing to know: zerocracy.yml is pinned to 0.17.23, so its job summary will be missing the metrics until the pin gets bumped to a release that includes this change.

Closes #2086

### DIFF
--- a/.github/workflows/zerocracy.yml
+++ b/.github/workflows/zerocracy.yml
@@ -55,12 +55,6 @@ jobs:
           | Timestamp | $(date -u) |
           | Status | ${{ job.status }} |
           EOF
-          if [ -f summary.md ]; then
-            cat summary.md >> "$GITHUB_STEP_SUMMARY"
-          else
-            echo "" >> "$GITHUB_STEP_SUMMARY"
-            echo "_Detailed summary not available (no data from action container)._" >> "$GITHUB_STEP_SUMMARY"
-          fi
       - name: Flag on failure
         if: failure() && github.repository_owner == 'zerocracy'
         run: |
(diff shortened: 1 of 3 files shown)


Thayorns:
@yegor256 take a look please, happy to clarify anything about the summary change.

PR #2198, opened 2026-09-11
fix(#2085): re-query latest issue on every cycle instead of stopping after first marker

### LINKED ISSUE #2085: `find-latest-issue` stops permanently after its first successful marker and never notices newer issues
## Problem
`find-latest-issue.rb` returns immediately when the iterator marker (`latest`) is positive. The marker is persisted after the first successful lookup, so every later run skips the GitHub request entirely.

## Why this is a bug
The judge is named find-latest-issue and should track the newest issue over time. After one run, opening any newer issue in the repository leaves the factbase pointing at the old issue forever; the judge never refreshes its boundary or details. The early-return condition is appropriate for the earliest issue (which is immutable), but not for the latest one.

## Expected behavior
Re-query the latest issue on subsequent cycles (or compare the returned number with the marker and advance it), while retaining the one-time behavior for the earliest judge. Add a fixture with an existing marker and a newer GitHub issue.

Thayorns:
The find-latest-issue judge returned early whenever its iterator marker was positive. Since the marker is saved after the first successful lookup, every later run skipped GitHub entirely, so newer issues were never noticed and the marker never moved.

The early return is now gone from find-latest-issue, so the judge asks GitHub for the newest issue on each cycle, records it when it is new, and advances the marker. When the newest issue is already known, Fbe.if_absent skips it as before. find-earliest-issue keeps its one-time behavior, because the earliest issue never changes.

Two unit tests cover a stored marker with a newer issue on GitHub, the existing already-found test now stubs the issues list, and a new advance-marker.yml fixture starts from an existing marker and expects it to move to the newer issue.

Closes #2085

### DIFF
--- a/judges/find-latest-issue/advance-marker.yml
+++ b/judges/find-latest-issue/advance-marker.yml
@@ -1,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+---
+options:
+  testing: true
+  repositories: foo/bazz
+input:
+  -
+    _id: 1
+    what: iterate
+    where: github
+    repository: 810
+    latest_issue_was_found: 100
+expected:
+  - /fb[count(f)=2]
+  - /fb/f[what='pull-was-opened' and repository='810' and issue='144' and where='github']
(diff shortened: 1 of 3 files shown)


Thayorns:
@yegor256 take a look please, happy to clarify anything about the marker change

Thayorns:
@yegor256 please check again

PR #2199, opened 2026-09-11
fix(#2083): name the found issue latest in find-latest-issue details

### LINKED ISSUE #2083: `find-latest-issue` records details saying “first we found” for the latest issue
## Problem
`judges/find-latest-issue/find-latest-issue.rb` discovers the newest issue (`direction: :desc`) but writes the details text "The issue ... is the first we found". The corresponding earliest judge uses the same wording.

## Why this is a bug
The details fact is displayed and can be used in reports/reward explanations. A consumer reading the latest-issue fact is told it is the first issue, which is factually false and indistinguishable from the earliest result. This is not merely wording: downstream automation can use details to explain why the issue was selected.

## Expected behavior
Say “latest”/“most recent” in the latest judge's details and add an assertion that the generated text identifies the latest issue.

Thayorns:
The find-latest-issue judge picks the newest issue of a repository, yet the details it records claimed the issue "is the first we found", which is the wording of the earliest judge and tells a reader the opposite of what happened.

The judge now records that the issue "is the latest we found". The two existing assertions on the details text follow the new wording, and a new test feeds a randomly numbered issue, derived from a seed printed on failure, and checks that its details name it the latest one.

Closes #2083

### DIFF
--- a/judges/find-latest-issue/find-latest-issue.rb
+++ b/judges/find-latest-issue/find-latest-issue.rb
@@ -81,7 +81,7 @@ Fbe.iterate do
           f.stale = 'branch'
         end
       end
-      f.details = "The issue #{Fbe.issue(f)} is the first we found, opened by #{Fbe.who(f)}."
+      f.details = "The issue #{Fbe.issue(f)} is the latest we found, opened by #{Fbe.who(f)}."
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 take a look please, happy to clarify anything about the new wording.

Thayorns:
@yegor256 done, could you look again

PR #2200, opened 2026-09-11
fix(#2081): keep build failures per repository so mttr never pairs runs of two repos

### LINKED ISSUE #2081: `some_build_success_rate` pairs failures across repositories that reuse the same workflow ID
## Problem
`judges/quality-of-service/some_build_success_rate.rb` keys the `failed` hash only by `json[:workflow_id]`, while it processes workflow runs from every repository in one invocation.

## Why this is a bug
GitHub workflow IDs are repository-scoped, not globally unique. A failed workflow ID 42 in repo A followed by a successful workflow ID 42 in repo B is treated as one failure/recovery pair, creating a bogus MTTR and deleting the failure state for repo A. Conversely, a real recovery in repo A can be missed after a coincidental run in repo B. The quality-of-service metric is therefore wrong whenever repositories share workflow ID values (which is common).

## Expected behavior
Key failures by repository plus workflow ID (and keep the repository in the run record), with a fixture covering identical IDs in two repositories.

Thayorns:
`some_build_success_rate` kept one `failed` hash for the whole invocation and keyed it only by `workflow_id`. GitHub workflow IDs are repository-scoped, so a failure in one repository could be paired with a success in another repository that uses the same ID. That produced a bogus MTTR, and a real recovery could be measured against a foreign failure, which even gave negative repair times.

Runs are already fetched and sorted one repository at a time inside `Fbe.unmask_repos`, so the hash now lives inside that block. Failure state can't outlive its repository. This does the same as keying by repository plus workflow ID, without a composite key. The repository is not added to the run record because nothing would read it.

Two new tests cover the two failure modes from the issue: a failure followed by a success with the same workflow ID in another repository, and a real recovery that follows a foreign failure with the same ID. Both use a seeded random workflow ID.

Closes #2081

### DIFF
--- a/judges/quality-of-service/some_build_success_rate.rb
+++ b/judges/quality-of-service/some_build_success_rate.rb
@@ -11,8 +11,8 @@ def some_build_success_rate(fact)
   success = []
   duration = []
   ttrs = []
-  failed = {}
   Fbe.unmask_repos do |repo|
+    failed = {}
     workflows =
       begin
         Fbe.octo.repository_workflow_runs(
--- a/test/judges/test-quality-of-service.rb
+++ b/test/judges/test-quality-of-service.rb
@@ -654,6 +654,61 @@ class TestQualityOfService < Jp::Test
     end
   end
 
+  def test_dont_pair_failure_with_success_of_other_repository
+    $loog = Loog::NULL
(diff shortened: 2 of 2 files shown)


Thayorns:
@yegor256 take a look please, happy to clarify anything about the approach

PR #2204, opened 2026-09-11
#2076: pin check run ids as actions job ids in fetch_workflows tests

### LINKED ISSUE #2076: fetch_workflows passes check-run IDs to workflow_run_job, so real builds are never counted
## What is wrong?
`Jp.fetch_workflows` feeds each ID returned by `check_runs_for_ref` directly into `workflow_run_job`. GitHub's check-run ID and Actions job ID are different resources; `workflow_run_job` calls `/actions/jobs/{job_id}` and therefore rejects a check-run ID with 404. The rescue then discards the run, so real pull requests report zero succeeded/failed builds.

## Where?
`lib/pull_request.rb:146-168`

## Reproduction
1. Process a pull request whose head SHA has a GitHub Actions check run.
2. Observe `check_runs_for_ref` returning a check-run ID (for example, `28907016501`).
3. The next request is `GET /repos/<repo>/actions/jobs/28907016501`; GitHub responds 404 because that number is not a job ID.
4. `fetch_workflows` rescues the 404 and omits the build from both counters.

## Expected behavior
Workflow checks belonging to the pull request should be counted using the actual Actions workflow-run/job identifiers.

## Suggested fix
Resolve each check run to its workflow run/job through the check run's `details_url` or use an API endpoint that returns the associated run ID, then call `workflow_run`/`workflow_run_jobs` with that real identifier. Update the fake client and tests so check-run and job IDs are distinct.

Thayorns:
The claim in #2076 does not hold against the live GitHub API. For GitHub Actions a check run and its job share one identifier: every github-actions check run on the heads of merged pull requests #2095 and #2091 and on the current master resolves through /actions/jobs/{id} with a 200, and the returned job carries the same id and a check_run_url ending in that id, for example check run 102357527083 resolves to job 102357527083 of run 34317780383. The example id 28907016501 from the issue answers 404 both as a job and as a check run, so the reported failure cannot be reproduced and lib/pull_request.rb needs no change.

Instead of touching production code, this pull request pins the real behaviour with three tests of Jp.fetch_workflows driven by seeded random twelve-digit identifiers: a check run is counted through the job of its own id while its workflow run id stays distinct, several jobs of one workflow run are all counted from a single fetch of that run, and a check run from another app never reaches the jobs endpoint, since only there the id is not a job id.

Closes #2076

### DIFF
--- a/test/lib/test_pull_request.rb
+++ b/test/lib/test_pull_request.rb
@@ -255,6 +255,95 @@ class TestPullRequest < Jp::Test
     assert_equal({ succeeded_builds: 0, failed_builds: 1 }, result)
   end
 
+  def test_fetch_workflows_counts_check_run_by_its_own_job_id
+    WebMock.disable_net_connect!
+    rate_limit_up
+    $options = Judges::Options.new({})
+    $global = {}
+    $loog = Loog::NULL
+    seed = Random.new_seed
+    random = Random.new(seed)
+    job = random.rand(100_000_000_000...1_000_000_000_000)
+    run = random.rand(10_000_000_000...100_000_000_000)
+    pr = { number: 66, head: { sha: 'dd789' }, base: { repo: { full_name: 'foo/foo' } } }
+    stub_github(
+      'https://api.github.com/repos/foo/foo/commits/dd789/check-runs?per_page=100',
+      body: {
+        check_runs: [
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 have a look please at this one

PR #2205, opened 2026-09-11
fix(#2203): create the missing factbase directory before resolving its path

### LINKED ISSUE #2203: entry.sh aborts when the factbase points into a directory that does not exist yet
The factbase path in `entry.sh:75` goes through plain `realpath`. In the Debian-based `ruby:4.0.5` image that is GNU coreutils, which lets the last path component be missing but requires every directory above it to exist. So a workflow with `factbase: data/new-product.fb` and no `data/` in the checkout gets `realpath: .../data/new-product.fb: No such file or directory`, and `set -e` stops the action before `judges pull` and `judges update` run.

Nothing in the action creates that directory, so such a configuration fails on every fresh checkout, not only once. A factbase directly in the workspace, like `factbase: new-product.fb`, is not affected: its parent is the workspace itself, and judges creates the file on the first run.

Creating the parent directory with `mkdir -p` before the `realpath` call at `entry.sh:75` fixes it, and an `entries/` scenario that sets `INPUT_FACTBASE` to a path inside a missing directory would lock it down. This is the factbase twin of #2202.

@yegor256

Thayorns:
The factbase path in `entry.sh` went through plain `realpath`, which in GNU coreutils tolerates a missing file but not a missing directory above it. A workflow with `factbase: data/new-product.fb` and no `data/` in the checkout therefore stopped right at path resolution, before `judges pull` or `judges update` could run.

Now `entry.sh` creates the parent directory with `mkdir -p` before canonicalizing the path, so judges can create the factbase there on the first run, exactly as it already does for a factbase at the root of the workspace.

Two new `entries/` scenarios cover this: one with a relative path two directories deep and one with an absolute path, both pointing into directories that do not exist yet.

Closes #2203

### DIFF
--- a/entries/creates-a-missing-absolute-factbase-directory.sh
+++ b/entries/creates-a-missing-absolute-factbase-directory.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+set -e -o pipefail
+
+SELF=$1
+
+source "${SELF}/makes/setup-test-env.sh"
+source "${SELF}/makes/test-common.sh"
+setup_test_env "${SELF}" name
+
+run_entry_script "${SELF}" success \
+  "GITHUB_WORKSPACE=$(pwd)" \
+  "GITHUB_REPOSITORY=zerocracy/judges-action" \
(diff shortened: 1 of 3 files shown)


Thayorns:
@yegor256 happy to clarify anything about this change if it helps.

Thayorns:
@yegor256 mind taking another look

PR #2211, opened 2026-09-12
#2187: remove dead uniq! call in find-missing-issues

### LINKED ISSUE #2187: `issues.uniq!` is dead code in `find-missing-issues.rb`
The query in `judges/find-missing-issues/find-missing-issues.rb:35` already carries `(unique issue)`, and line 38 then calls `issues.uniq!` on the array `map(&:issue)` built from its result.

`Factbase::Unique#evaluate` (factbase-0.19.16, `lib/factbase/terms/unique.rb:22`) keeps a `@seen` set across the enumeration and returns false for any tuple it has yielded before, so the query cannot hand back two facts sharing an `issue` value. The array is unique before `uniq!` touches it, and the call is a no-op on every possible input. It reads as if duplicates were expected there, which sends the next person looking for a case that cannot happen.

Deleting line 38 is the whole fix; `issues.sort!` on the line below is still needed for the `min..max` range built on `:40`.

@yegor256

canburakyol:
Fixes #2187.Fbe.fb.query already carries (unique issue), so issues.uniq! is a no-op. Removed the line, kept issues.sort! for the min..max range below.

Windows'ta undle install openssl native gem yüzünden patladığı için lokal test koşamadı (
uby -c Syntax OK). CI'da 
ake testlerinin koşması bekleniyor.

### DIFF
--- a/judges/find-missing-issues/find-missing-issues.rb
+++ b/judges/find-missing-issues/find-missing-issues.rb
@@ -22,7 +22,6 @@ Fbe.consider('(and (eq where "github") (exists repository) (unique repository))'
   issues = Fbe.fb.query(
     "(and (eq repository #{r.repository}) (exists issue) (eq where 'github') (unique issue))"
   ).each.map(&:issue)
-  issues.uniq!
   issues.sort!
   next if issues.empty?
   must = (issues.min..issues.max).to_a


Thayorns:

canburakyol:
okey

PR #2212, opened 2026-09-12
#2190: remove noisy tombstone-count log line

### LINKED ISSUE #2190: Leftover debug logging in `find-missing-issues.rb:101` rescans the tombstone per fact
Line 101 of `judges/find-missing-issues/find-missing-issues.rb` logs `The #{type} #{Fbe.issue(f)} is not tombstoned among #{ts.issues('github', r.repository).count}`. It arrived with the tombstone work in `49402002c0` and reads as a note the author left themselves while chasing #1217: the sentence names no unit, so an operator sees "is not tombstoned among 42" and has nothing to do with the number.

The cost is not only the wording. `Fbe::Tombstone#issues` runs a full factbase query and expands every stored range into an array of numbers, and this line calls it once per fact recovered, inside the transaction opened on `:68`. The two lines around it already carry the information worth keeping: `:100` records the event in `f.details`, and `:102` logs when the missing issue was opened.

Deleting line 101 removes the query and leaves the useful logging in place.

@yegor256

canburakyol:
Fixes #2190.

Drops the per-fact $loog.info(... not tombstoned among ...count) line: it names no unit for the operator and costs a full Tombstone#issues query per recovered fact inside the txn. .details + was found opened ... ago logging stays.

### DIFF
--- a/judges/find-missing-issues/find-missing-issues.rb
+++ b/judges/find-missing-issues/find-missing-issues.rb
@@ -85,7 +85,6 @@ Fbe.consider('(and (eq where "github") (exists repository) (unique repository))'
         end
       end
       f.details = "The missing #{type} #{Fbe.issue(f)} has been opened by #{Fbe.who(f)}."
-      $loog.info("The #{type} #{Fbe.issue(f)} is not tombstoned among #{ts.issues('github', r.repository).count}")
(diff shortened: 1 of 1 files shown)


Thayorns:
Stop your agents from submitting PRs against unlabeled issues—it produces chaos and deduct payment. Please

PR #2213, opened 2026-09-12
#2191: drop dead nil fallback in unmask_repos

### LINKED ISSUE #2191: Dead `|| ''` fallback in `lib/patches/unmask_repos.rb:21`
`lib/patches/unmask_repos.rb:21` reads `masks = (options.repositories || '').split(',')`, and the fallback cannot fire. Line 14 has already raised `Fbe::Error, 'Repositories mask is not specified'` when `options.repositories` is falsy, and line 15 raises `'Repositories mask is empty'` when it has no characters. By the time line 21 runs the value is a non-empty string, guaranteed by the two lines above it.

The line came in with the verbatim copy of the upstream method — `fbe-0.49.1 lib/fbe/unmask_repos.rb:73` has it after the same two raises — so it is dead in both, but this is the copy this repository maintains and #1904 is still open on the copying itself. What it costs is a false hint: it tells the next reader that `options.repositories` may be nil at this point, which is the opposite of what lines 14 and 15 establish, and that reader has to walk back up the method to find out which is true.

`masks = options.repositories.split(',')` says what actually holds.

@yegor256 please take a look.

canburakyol:
Fixes #2191.

Lines 14-15 already guarantee \options.repositories\ is a non-empty string, so \|| ''\ can never fire. Simplify to \options.repositories.split(',').

### DIFF
--- a/lib/patches/unmask_repos.rb
+++ b/lib/patches/unmask_repos.rb
@@ -18,7 +18,7 @@ module Fbe
       )
       repos = []
       octo = Fbe.octo(loog:, global:, options:)
-      masks = (options.repositories || '').split(',')
+      masks = options.repositories.split(',')
       masks.reject { |m| m.start_with?('-') }.each do |mask|
         unless mask.include?('*')
           repos << mask


Thayorns:

PR #2214, opened 2026-09-12
#2189: fix mismatched user id in find-all-issues stub

### LINKED ISSUE #2189: The `user/526302` stub answers with id 526301 in `test-find-all-issues.rb`
The two user stubs in `test_find_all_issues` disagree with each other. `test/judges/test-find-all-issues.rb:193-194` answers `user/526301` with `id: 526_301`, and `:197-198` answers `user/526302` with:

```ruby
body: { login: 'yegor257', id: 526_301, type: 'User', site_admin: false }
```

The search fixture on `:176` and `:181` hands out 526301 for yegor256 and 526302 for yegor257, so the second stub contradicts both its own URL and the payload it is there to support.

Nothing fails today, because the judge takes `who` from the search result on `judges/find-all-issues/find-all-issues.rb:97` and `Fbe.who` wants only the login, so the assertion on `:258` is satisfied either way. It still leaves a fixture that states two different ids for one account, and the next assertion that reads an id back from that endpoint will be comparing against the wrong one.

Changing it to `id: 526_302` makes the stub agree with its URL.

@yegor256 please take a look.

canburakyol:
Fixes #2189.

user/526302 stub returned id: 526_301, contradicting its URL and the search fixture (526302 = yegor257). One-digit fix to 526_302.

### DIFF
--- a/test/judges/test-find-all-issues.rb
+++ b/test/judges/test-find-all-issues.rb
@@ -164,7 +164,7 @@ class TestFindAllIssues < Jp::Test
     )
     stub_github(
       'https://api.github.com/user/526302',
-      body: { login: 'yegor257', id: 526_301, type: 'User', site_admin: false }
+      body: { login: 'yegor257', id: 526_302, type: 'User', site_admin: false }
     )
     fb = Factbase.new
     fb.insert.then do |f|


Thayorns:

canburakyol:
@Thayorns Understood — agents stopped, will only target labeled/bounty issues going forward. Thanks for flagging. Pausing new PRs and monitoring existing ones.

PR #2215, opened 2026-09-12
#2188: drop blind find-issue yml pair

### LINKED ISSUE #2188: Both `find-issue.yml` files expect issue 144, so neither pins the sort direction
`judges/find-latest-issue/find-issue.yml:11` and `judges/find-earliest-issue/find-issue.yml:11` both expect issue 144, and `diff` between the two files reports that one line and nothing else. The only line that separates the two judges is the sort direction — `direction: :desc` at `judges/find-latest-issue/find-latest-issue.rb:37` against `direction: :asc` at `judges/find-earliest-issue/find-earliest-issue.rb:37` — and neither yml can see it, because `list_issues` in the fake takes its options as `_options` and discards them (`fbe-0.49.1/lib/fbe/fake_octokit.rb:463`). Both judges call `.first` on the same three-element list for `foo/bazz` and both get 144, whatever they asked for.

That leaves the earliest expectation wrong on its own terms. The fake dates issue 144 at 2025-05-29 and issues 42 and 43 at 2024-09-20, so 144 is the newest of the three, and against real GitHub `sort: :created, direction: :asc, per_page: 1` returns 42. `find-earliest-issue/find-issue.yml` therefore asserts the latest issue under the name `earliest_issue_was_found`, and passes only because the fake is blind to the parameter. Swapping `:desc` for `:asc` in either judge leaves both files green.

The half of this that lives in this repository is the pair of expectations: neither file should be able to pass the other judge's assertion. Since the fake cannot tell the two directions apart, dropping the yml pair is the honest fix — `test/judges/test-find-latest-issue.rb` and `test/judges/test-find-earliest-issue.rb` already stub `direction=desc` and `direction=asc` into the request URL and do turn red when the direction is wrong. Teaching `list_issues` to honour `sort` and `direction`, so the earliest yml can expect 42, would be a change to the `fbe` gem instead.

This is not the `test/` migration puzzle in #2028: neither file is under `test/`, and neither uses `stub_request` or `stub_github`.

@yegor256 please take a look

canburakyol:
Fixes #2188.

Both find-issue.yml files expect issue 144 because the fake \list_issues\ ignores \sort/direction\, so neither pins the sort direction and earliest asserts the latest issue. Dropping the pair; \	est-find-latest-issue.rb\ (direction=desc) and \	est-find-earliest-issue.rb\ (direction=asc) already pin the direction via URL stubs and turn red when swapped. Fixing the fake itself belongs to the \be\ gem.


Thayorns:
Stop your agents from submitting PRs against unlabeled issues—it produces chaos and deduct payment. Please

PR #2216, opened 2026-09-12
#2186: prefix zero-quota log line with judge name

### LINKED ISSUE #2186: The zero-quota log line in `lib/qos_search.rb` names no judge
Every log line in `Jp.qosearch` is prefixed with the judge it belongs to — `lib/qos_search.rb:36`, `:48`, `:58`, `:74` and `:79` all open with `[#{jg}]`. The one on `:64` does not:

```ruby
$loog.info('Too much GitHub Search API quota consumed already (0 left)')
```

It is a frozen literal, so it carries neither the judge name nor the query that hit the wall. All judges of a cycle write into the same log, and this is the line that explains why searching stopped, so it is the one a reader most needs to attribute. It is also the message that will appear once per judge while the latch holds.

Prefixing it the way its five neighbours are prefixed is the whole fix.

@yegor256 please take a look.

canburakyol:
Fixes #2186.

All other \Jp.qosearch\ log lines open with \[#{\}]\; the zero-quota line was a frozen literal with no attribution. Prefixed it like its five neighbours.


canburakyol:
Rebased onto current master: upstream refactored quota state to per-judge hashes with a local \jg\ variable, so the prefix now uses \#{jg}\ like its neighbours instead of \#{\}\. The zero-quota line was still unprefixed after the refactor, so the fix still applies.

PR #2218, opened 2026-09-12
#2203: create factbase parent dir before realpath

### LINKED ISSUE #2203: entry.sh aborts when the factbase points into a directory that does not exist yet

canburakyol:
Fixes #2203.

GNU \
ealpath\ requires every directory above the last path component to exist, so \actbase: data/foo.fb\ with no \data/\ in the checkout aborted the action under \set -e\ before \pull/update\. \mkdir -p\ the parent dir before the \
ealpath\ call.

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -72,7 +72,9 @@ fi
 
 name="$(basename "${INPUT_FACTBASE}")"
 name="${name%.*}"
-fb=$(realpath "$( [[ ${INPUT_FACTBASE} = /* ]] && echo "${INPUT_FACTBASE}" || echo "${GITHUB_WORKSPACE}/${INPUT_FACTBASE}" )")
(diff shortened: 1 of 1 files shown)


Thayorns:

Thayorns:
@canburakyol @yegor256 this PR is a duplicate of #2205 and should be closed

PR #2219, opened 2026-09-12
#2202: create sqlite-cache parent dir before realpath

### LINKED ISSUE #2202: entry.sh aborts when sqlite-cache points into a directory that does not exist yet
The `sqlite-cache` path in `entry.sh:218` goes through plain `realpath`. In the Debian-based `ruby:4.0.5` image that is GNU coreutils, which lets the last path component be missing but requires every directory above it to exist. So a workflow with `sqlite-cache: .cache/http.sqlite` and no `.cache/` in the checkout gets `No such file or directory` from `realpath`, and `set -e` stops the action before `judges download` and `judges update` run.

Nothing in the action creates that directory, so such a configuration fails on every fresh checkout, not only once. The default `http-cache.sqlite` is not affected, because its parent is the workspace itself.

Creating the parent directory with `mkdir -p` before the `realpath` call at `entry.sh:218` fixes it, and an `entries/` scenario that sets `INPUT_SQLITE-CACHE` to a path inside a missing directory would lock it down.

@yegor256

canburakyol:
Fixes #2202.

Same as #2203 but for \sqlite-cache\: GNU \
ealpath\ needs parent dirs to exist, so \sqlite-cache: .cache/http.sqlite\ with no \.cache/\ aborted the action under \set -e\. \mkdir -p\ the parent dir before the \
ealpath\ call.

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -212,7 +212,9 @@ fi
 
 sqlite=$(printenv "INPUT_SQLITE-CACHE" || true)
 if [ -n "${sqlite}" ]; then
-    sqlite=$(realpath "$( [[ ${sqlite} = /* ]] && echo "${sqlite}" || echo "${GITHUB_WORKSPACE}/${sqlite}" )")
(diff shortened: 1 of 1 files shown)


Thayorns:

PR #2220, opened 2026-09-12
#2183: drop dead refute_equal in marker test

### LINKED ISSUE #2183: `refute_equal` in `test_iteration_end_keeps_old_marker` cannot fail on its own
`test_iteration_end_keeps_old_marker` at `test/judges/test-find-all-issues.rb:262` ends with two assertions on the same expression:

```ruby
refute_equal(0, fb.query('(eq what "iterate")').each.to_a.first.min_issue_was_found)
assert_equal(45, fb.query('(eq what "iterate")').each.to_a.first.min_issue_was_found)
```

The `refute_equal` on `:300` cannot fail unless the `assert_equal` on `:301` fails as well: anything other than 45 fails the second line, and 45 is not 0. It reads like a guard against the marker being reset, but the line below already covers that, and it repeats the query to do it.

Dropping `:300` leaves the test with its one real assertion last, where a reader can find it.

@yegor256 please take a look.

canburakyol:
Fixes #2183.

\
efute_equal(0, ...min_issue_was_found)\ cannot fail on its own: any value other than 45 fails the \ssert_equal(45, ...)\ below it, and 45 is not 0. Dropped the dead line, kept the one real assertion.

### DIFF
--- a/test/judges/test-find-all-issues.rb
+++ b/test/judges/test-find-all-issues.rb
@@ -266,7 +266,6 @@ class TestFindAllIssues < Jp::Test
     end
     load_it('find-all-issues', fb)
     assert_equal(3, fb.size)
-    refute_equal(0, fb.query('(eq what "iterate")').each.to_a.first.min_issue_was_found)
     assert_equal(45, fb.query('(eq what "iterate")').each.to_a.first.min_issue_was_found)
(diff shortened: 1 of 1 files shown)


Thayorns:

PR #2221, opened 2026-09-12
#2185: cover pull branch in find-one-pull yml

### LINKED ISSUE #2185: `find-one-pull.yml` never reaches the pull branch of `find-missing-issues.rb`
The test `judges/find-missing-issues/find-one-pull.yml` seeds two `pull-was-opened` facts for issues 141 and 143, leaving 142 as the missing number the judge should recover. With `testing: true` the judge talks to `Fbe::FakeOctokit`, and its `issue` method returns number 142 from a branch that carries no `pull_request` key (fbe-0.49.1, `lib/fbe/fake_octokit.rb`, the `when 142` case). So `find-missing-issues.rb:78` computes `type = 'issue'` and the `if type == 'pull'` block on `:79` is skipped whole.

That block is the only thing the file's name promises. The `Fbe.octo.pull_request` lookup on `:84`, `f.branch = ref` on `:95` and `f.stale = 'branch'` on `:97` are never reached, and the two expectations are the same two lines `find-one.yml` already carries, down to asserting `what='issue-was-opened'` for the recovered fact. The file duplicates its neighbour and covers no extra path, so the pull side of the judge has no `.yml` test at all.

Seeding 143 and 145 instead leaves 144 missing, which `FakeOctokit#issue` does return with a `pull_request` key and `FakeOctokit#pull_request` answers with `head.ref` of `master`. Expecting `what='pull-was-opened'` and `branch='master'` then puts the branch under test.

@yegor256

canburakyol:
Fixes #2185.

\ind-one-pull.yml\ seeded 141/143 so the fake returned 142 with no \pull_request\ key and the \	ype == 'pull'\ block was never reached — a duplicate of \ind-one.yml\. Seed 143/145 instead so 144 (which the fake returns with a \pull_request\ key and \head.ref\ of \master\) exercises the pull path; expect \pull-was-opened\ + \ranch='master'\.

### DIFF
--- a/judges/find-missing-issues/find-one-pull.yml
+++ b/judges/find-missing-issues/find-one-pull.yml
@@ -9,13 +9,13 @@ input:
     where: github
     when: 2024-10-10T15:00:00Z
     repository: 700
-    issue: 141
+    issue: 143
   -
     what: pull-was-opened
     where: github
     when: 2024-10-11T15:00:00Z
     repository: 700
-    issue: 143
+    issue: 145
 expected:
   - /fb[count(f)=3]
-  - /fb/f[what='issue-was-opened' and repository='700' and issue='142']
+  - /fb/f[what='pull-was-opened' and repository='700' and issue='144' and branch='master']


Thayorns:

PR #2222, opened 2026-09-12
#2182: assert no duplicate fact on second call

### LINKED ISSUE #2182: Nothing is asserted by `test_graceful_on_second_call_for_same_issue` in `test/lib/test_issue_was_lost.rb`
The test `test_graceful_on_second_call_for_same_issue` in `test/lib/test_issue_was_lost.rb:56` calls `Jp.issue_was_lost('github', 42, 123)` twice inside a `Fbe.stub(:fb, fb)` block and then ends. There is no assertion anywhere in the body, so the only failure it can produce is an exception raised by the code under test.

That leaves the behaviour the test is named for unchecked. The second call is supposed to be a no-op: `Fbe.if_absent` in `lib/issue_was_lost.rb:27` finds the `issue-was-lost` fact the first call inserted, returns nil, and the judge logs a warning and stops. Nothing in the test looks at the factbase afterwards, so if that uniqueness check ever stopped matching, a second `issue-was-lost` fact for issue 123 would pile up on every run and the test would still be green. #1647 was about the exception; the state the call leaves behind has no guard.

Asserting the factbase after the second call fixes it: exactly one `issue-was-lost` fact for issue 123, in the shape `test_inserts_fresh_fact_when_none_exists` at `:29` already uses.

@yegor256

canburakyol:
Fixes #2182.

\	est_graceful_on_second_call_for_same_issue\ called \Jp.issue_was_lost\ twice and asserted nothing, so a broken uniqueness check (duplicate \issue-was-lost\ fact piling up every run) would stay green. Assert exactly one fact afterwards, in the shape \	est_inserts_fresh_fact_when_none_exists\ uses.

### DIFF
--- a/test/lib/test_issue_was_lost.rb
+++ b/test/lib/test_issue_was_lost.rb
@@ -59,6 +59,10 @@ class TestIssueWasLost < Minitest::Test
       $loog = Loog::NULL
       Jp.issue_was_lost('github', 42, 123)
       Jp.issue_was_lost('github', 42, 123)
+      lost = fb.query(
+        "(and (eq what 'issue-was-lost') (eq where 'github') (eq repository 42) (eq issue 123))"
+      ).each.to_a
+      assert_equal(1, lost.size, 'second call for the same issue must be a no-op, no duplicate fact')
     end
   end
 


Thayorns:

PR #2232, opened 2026-09-13
#2175: stop falling back to core quota in search fallback

### LINKED ISSUE #2175: `Jp.qosearch` fallback reads the core quota, not the search quota, in `lib/qos_search.rb`
`Jp.qosearch` reads the search allowance from `/rate_limit` on `lib/qos_search.rb:43` and picks the right resource out of it on `:53`, with `json.dig(:resources, :search, :remaining)`. The fallback on `:46`, taken when the client has no `get` method, instead does `left = octo.rate_limit.remaining`, and that is a different number.

In `octokit-10.0.0/lib/octokit/client/rate_limit.rb:14-18`, `rate_limit` builds its answer from `last_response`, so it reports the `x-ratelimit-remaining` header of whatever call happened last — the core allowance for nearly every endpoint this action touches, never the search one. `Fbe::FakeOctokit` has no `get` at all and its `rate_limit` returns a fixed `remaining: 100` (`fbe-0.49.1/lib/fbe/fake_octokit.rb:60-62`), so every `testing: true` run, the two `.yml` fixtures in `judges/find-all-issues/` included, goes down this branch and sees a number that has nothing to do with the search quota.

This is the defect filed as #1458 and closed when the `/rate_limit` probe arrived: the main path was corrected and the fallback kept the old reading. The zero-left latch on `:61` therefore cannot trip on the only path the test suite ever executes, which is also why #1453 still finds that state machine uncovered.

Having the fallback raise, rather than guess from a different resource, is the honest fix — a client that cannot answer `/rate_limit` cannot tell `Jp.qosearch` what it needs, and the `left.nil?` branch on `:55` already knows how to stop searching when the figure is unavailable.

@yegor256 please take a look.

canburakyol:
Fixes #2175. When octo.get is not available, the fallback reads octo.rate_limit.remaining which is the core quota (limit 5000), not the search quota (limit 30). This lets the code make far more search calls than GitHub allows, resulting in 403 errors. The fix: when we cannot determine the search quota, treat it as spent and stop searching. This is safe because core and search quotas are independent. Added test_skips_search_when_get_unavailable, which fails on the old code (returns core quota 5000, allows search) and passes on the new (stops searching).


Thayorns:

PR #2233, opened 2026-09-13
#2178: validate search method before consuming budget

### LINKED ISSUE #2178: `Jp.qosearch` checks its `method` argument after spending a probe and a budget slot
The whitelist that keeps a caller from dispatching an arbitrary method, added for #1728, sits at the very bottom of `Jp.qosearch`, on `lib/qos_search.rb:68-69`. By the time it runs, the `/rate_limit` probe on `:43` has already gone out over the network and `:67` has already charged a slot against the per-minute budget of 25.

So a call with a bad `method:` costs a real HTTP request, leaves `@scount[jg]` incremented for a search that never happened, and only then raises. The counter is not rolled back, so the rest of that minute has one fewer search than it should, and the `RuntimeError` reaches the judge after the side effects rather than before them.

The check needs nothing but its own argument. Moving it up to just after `jg = $judge` on `:20`, ahead of all the bookkeeping, makes the call fail on the spot and spend nothing.

@yegor256 please take a look.

canburakyol:
Fixes #2178. The method argument check was after @scount[jg] += 1, so an invalid method consumed a budget slot before raising. Now the validation runs first. Added test_invalid_method_does_not_consume_budget, which fails on the old code (budget = 1 after invalid call) and passes on the new (budget = 0).


Thayorns:

PR #2234, opened 2026-09-13
#2170: fall back to per-repo commit count when batch fails

### LINKED ISSUE #2170: `total_commits` loses every repository count when one of them fails in `total_commits.rb`
The `total_commits` judge in `judges/dimensions-of-terrain/total_commits.rb:31` hands every repository to one batched GraphQL call, `Fbe.github_graph.total_commits(repos:)`. That method builds a single query with one `repo_i` alias per repository and then walks the aliases in `fbe-0.49.1/lib/fbe/github_graph.rb:150-158`, raising `Fbe::Error` on the first alias whose `ref.target.history` is missing and abandoning the `map`, so the counts already collected for every other repository are thrown away. `Fbe::Graph#query` at `lib/fbe/github_graph.rb:51` does the same one level up, raising whenever the response carries any error at all, and GitHub returns an error for one alias alongside good data for all the rest.

Since #2025 that raise is rescued into `{}`, so `Jp.incremate` leaves the property unset and measures again next cycle. That is the right answer for a timeout and the wrong one for a repository that is permanently unreadable: one such repository in the mask means `total_commits` is never recorded at all, for the whole organization, on this cycle and on every cycle after it. The guards at lines 26 and 27 do not catch it, because a repository can report `size > 0` and a `default_branch` over REST and still have no resolvable ref over GraphQL, and the branch can be deleted between the two calls. Every sibling degrades per repository instead: `total_issues.rb:16` queries one repository at a time and `next`s past the broken one, and `quantity-of-deliverables/total_commits_pushed.rb:31` does the same, which makes `total_commits` the only one of the eight that cannot produce a partial count.

The batching arrived with #1037, which was about spending one request instead of N, and that saving is worth keeping in the normal case. Rescuing `Fbe::Error` around the batch and falling back to one query per repository, dropping only the repositories that fail, keeps the single request when everything is healthy and still counts the rest when one of them is not.

@yegor256 please take a look.

canburakyol:
Fixes #2170. When the batch GraphQL query for total commits fails (e.g. one repo is inaccessible), the old code returned nothing, losing all repository counts. Now it falls back to querying each repo individually and summing the results. Repos that fail individually are skipped with a log message. Network errors still return nothing (transient, retry next cycle).


Thayorns:

PR #2235, opened 2026-09-13
#2173: pin forks count in fake and measure-all expectation

### LINKED ISSUE #2173: `total_forks=0` in `measure-all.yml` cannot fail, the fake has no `forks` field
`judges/dimensions-of-terrain/measure-all.yml:18` expects `/fb/f[total_forks=0]`, and that expectation cannot fail. `Fbe::FakeOctokit#repository` (fbe-0.49.1, `lib/fbe/fake_octokit.rb:399`) returns a hash carrying `stargazers_count: 1` but no `forks` key at all — it has `fork: false`, the boolean saying the repository is not itself a fork, which is a different field. So `json[:forks] || 0` at `judges/dimensions-of-terrain/total_stars.rb:28` folds the absent field into a zero on every run driven by the fake, whatever the judge does with it.

That is the one thing #1992 asked these pins to prevent. The numbers were pinned so a regression writing a zero instead of a measurement would be caught, and `total_forks` is the single metric where a zero and a missing measurement are already indistinguishable, because the fake supplies nothing to measure. Nine of the ten expectations test arithmetic; the tenth tests a gap in the fixture.

Giving the fake a `forks` count and pinning `measure-all.yml:18` to it puts the expectation back on the judge's arithmetic. `lib/patches/fake_octokit.rb` is where this repository already fills gaps in the fake, though the field belongs upstream in `fbe` next to `stargazers_count`. #2145 reports the same shape for `list_milestones`.

@yegor256 please take a look.

canburakyol:
Closes #2173


canburakyol:
Closing per maintainer feedback on unlabeled issues — will refocus on labeled bounties.

Thayorns:

PR #2236, opened 2026-09-13
#2174: seed pull 44 so 45 rescue is exercised

### LINKED ISSUE #2174: Both paginated-pull tests skip the rescue they cover in `test-find-all-issues.rb`
`test_paginated_pulls_continue_after_not_found` at `test/judges/test-find-all-issues.rb:330` and `test_paginated_pulls_continue_after_forbidden` at `:372` each seed a `pull-was-opened` fact for issue 45 and then stub `repos/foo/foo/pulls/45` with a 404 and a 403. Neither stub is ever requested.

The search fixture returns items 45 and 46. When the judge reaches item 45, `Fbe.if_absent` finds the seeded fact and hands back `nil`, so `judges/find-all-issues/find-all-issues.rb:94` takes its `next if f.nil?` branch before the `Fbe.octo.pull_request` call on `:101` is reached. The only pull that gets as far as the rescue clauses on `:102-112` is 46, which the fixtures answer successfully.

Both tests then assert that 46 landed in the factbase, and 46 lands there just as well with those two rescues deleted. That is the truncation from #1362 these tests were added to lock down, and neither of them would notice its return. WebMock does not fail on an unused stub, and neither test calls `assert_requested`.

Seeding the fact for issue 44 instead, and stubbing `repos/foo/foo/issues/44` for the `after` lookup, gives the iterator its starting point while leaving 45 absent from the factbase, so the 404 and the 403 on `pulls/45` are actually reached.

@yegor256 please take a look.

canburakyol:
Closes #2174


Thayorns:

PR #2237, opened 2026-09-13
#2171: distinguish transient listing failure from bad mask

### LINKED ISSUE #2171: "No repos found matching" blames the mask for a failed organization listing
`lib/patches/unmask_repos.rb:64` raises `Fbe::Error, "No repos found matching: ..."` whenever `repos` comes out empty, and names the mask as the thing at fault. The patch's own rescue at `:36-43` can produce an empty list without the mask being wrong at all: a 403, a 500, a timeout or an `Octokit::Unauthorized` while listing an organization logs a warning, `next`s past that mask, and contributes nothing to `repos`. When that was the only mask, line 64 then blames the configuration for a transient GitHub failure.

What an operator reads is `No repos found matching: "zerocracy/*"`, which sends them hunting for a typo in a mask that is perfectly correct, while the real cause sits one warning line above it. The rescue itself says "(will retry next cycle)" and the raise six lines later contradicts it by killing the judge instead.

Tracking whether any mask was skipped by that rescue, and raising a different error naming the skipped mask and the underlying failure when one was, keeps the configuration error apart from the transport error. `test/lib/test_unmask_repos.rb:33-74` covers all three listing failures, but every case keeps a second literal mask `bar/bar` in the list, so the empty outcome is never reached.

@yegor256 please take a look.

canburakyol:
Closes #2171


Thayorns:

PR #2238, opened 2026-09-13
#2168: skip repo with nil default_branch before tree request

### LINKED ISSUE #2168: `total_files` asks for a tree at an empty ref when the default branch is nil
`total_files` in `judges/dimensions-of-terrain/total_files.rb:30` passes `info[:default_branch]` straight into `Fbe.octo.tree`, without checking that the repository has one. Its neighbour `total_commits.rb:26` reads the same field out of the same `Fbe.octo.repository` call and does `next if json[:default_branch].nil?` before using it, so the two files disagree about whether a nil default branch is something to skip.

When the field is nil, Octokit builds the path `/repos/{owner}/{repo}/git/trees/` with an empty ref. The suite already knows this happens: `test/judges/test-dimensions-of-terrain.rb:525` stubs `https://api.github.com/repos/foo/nobranch/git/trees/?recursive=true` with a 404, which is the URL the judge really sends. The 404 lands in the `rescue Octokit::NotFound` arm at `total_files.rb:31` and is logged as `Tree not found for foo/nobranch@`, so a repository that merely has no default branch is reported as one whose tree went missing, and an API call that could never succeed is spent on it first.

Adding `next if info[:default_branch].nil?` after the size check at line 27, the way `total_commits.rb:26` already does, skips the repository before the request instead of after it.

@yegor256 please take a look.

canburakyol:
Closes #2168


Thayorns:

PR #2239, opened 2026-09-13
#2172: use injected loog for archived check

### LINKED ISSUE #2172: `unmask_repos` patch writes to `$loog` and to the injected `loog` in the same method
The patched `Fbe.unmask_repos` in `lib/patches/unmask_repos.rb:10` takes a logger as `loog:` and then writes to two different loggers inside the one method body. The mask-expansion rescue at line 38 and the scan line at line 66 use the parameter, while the two rescues around the archived check, at lines 55 and 58, use the `$loog` global instead. Upstream `fbe/lib/fbe/unmask_repos.rb` uses the parameter throughout, so this crept in with the copy.

A caller that passes its own logger gets most of this method's output where it asked for it and the two archived-check messages somewhere else, which is exactly the output that matters when a repository has gone missing. The five tests in `test/lib/test_unmask_repos.rb` all pass `loog: Loog::NULL` and then set `$loog = Loog::NULL` on the line above anyway, at lines 17, 28, 37, 52 and 66; that second assignment is only there because of these two lines, and without it a 404 on the archived check would call `info` on whatever `$loog` happens to hold.

Replacing `$loog` with `loog` on lines 55 and 58 is the whole fix, and the `$loog = Loog::NULL` lines in the tests can go with it.

@yegor256 please take a look.

canburakyol:
Closes #2172


Thayorns:

PR #2242, opened 2026-09-13
#2164: mark hoc stale when additions and deletions are missing

### LINKED ISSUE #2164: `fix-missing-hoc` writes `hoc = 0` when the API gave no numbers, and can never fix it
## What happens

```ruby
# judges/fix-missing-hoc/fix-missing-hoc.rb:58
f.hoc = (json[:additions] || 0) + (json[:deletions] || 0)
```

The `|| 0` fallbacks say the author expected the fields to be absent sometimes. The selecting query
is guarded by `(absent hoc)`, so once `0` is written the fact no longer matches and the judge can
never correct it.

The sibling judge handles the same shape, the API did not give me the value, by marking the fact
instead of inventing one:

```ruby
# judges/fix-missing-branch/fix-missing-branch.rb:58-62
ref = json.dig(:head, :ref)
if ref.nil?
  f.stale = 'branch'
```

## Reproduction

Read the two judges side by side.

A pull request with a real diff can be recorded as `hoc = 0` for good. `hoc` feeds the size based
scoring, so the pull is scored as an empty change, and no later run can notice or repair it because
the fact looks fully populated.

## What should happen

Treat a missing `additions` and `deletions` pair the way `fix-missing-branch` treats a missing
`head.ref`: leave `hoc` unset and mark the fact, so the value is either right or visibly absent.

canburakyol:
Closes #2164


canburakyol:
Closing per maintainer feedback — unlabeled, will refocus.

Issue #2158, opened 2026-09-10
`find-all-issues` reads one page of 30 results and reports it as a full scan


morphqdd:
## What happens

The search goes through `Jp.qosearch`, whose last call is

```ruby
# lib/qos_search.rb:70
Fbe.octo.with_disable_auto_paginate { |octo| octo.__send__(method, query, **) }
```

Auto pagination is turned off on purpose and no `per_page:`, `page:`, `sort:` or `order:` is passed.
The judge then takes:

```ruby
# judges/find-all-issues/find-all-issues.rb:66-79
json = Jp.qosearch("repo:#{repo} type:#{type} created:>=#{after.iso8601[0..9]}")
json ? json[:items] : []
```

`json[:total_count]` is never read and there is no paging loop.

The query is deliberately wide, `created:>=` the creation date of the oldest known issue, so usually
the whole life of the repository, and `total_count` is routinely in the hundreds or thousands.
Without pagination the judge sees at most the default page of 30, and with no `sort` or `order` those
30 are GitHub's relevance ordering, an arbitrary and non reproducible subset.

## Reproduction

Read `lib/qos_search.rb:70` and `Fbe#with_disable_auto_paginate` at `fbe/lib/fbe/octo.rb:241-249`,
which sets `@origin.auto_paginate = false` for the block. No `per_page` is passed anywhere on this
path.

A judge called "find all issues" examines up to 30 arbitrary issues per repository per run and
ignores the rest, while the message it throws says otherwise:

```ruby
"Checked #{seen.count} #{type}s in #{repo}", ... "created >= #{after.iso8601[0..9]};"
```

In a busy repository missing issues are effectively never back filled.

## What should happen

Page through the results, respecting the 1000 result ceiling and the search budget `Jp.qosearch`
already implements, or at least ask for `per_page: 100`, log `total_count` and say in the log that
the scan was truncated.

0crat:
[...]

PR #2254, opened 2026-09-15
fix(#2177): keep the iterated issue as the find-all-issues marker

### LINKED ISSUE #2177: `find-all-issues` takes its marker from an arbitrary search result in `find-all-issues.rb`
`judges/find-all-issues/find-all-issues.rb:92` sits inside the block that builds the fact and assigns to the iterator's own cursor: `issue = ff.issue`, where `issue` is the `over do |repository, issue|` parameter from `:30`. Whatever that variable holds at `:134` is what `Fbe::Iterate#over` persists as `min_issue_was_found` (`fbe-0.49.1/lib/fbe/iterate.rb:317-322`).

`Fbe.if_absent` yields before it looks for an existing fact (`fbe-0.49.1/lib/fbe/if_absent.rb:64` and `:75`), so the line runs for every item the search returned, not only the ones that were inserted — the `next if f.nil?` on `:94` comes too late to stop it. The cursor therefore ends up holding the number of the last element of the response, and with auto pagination off and no `sort` or `order` on the query that element is an arbitrary one from GitHub's relevance ordering, as #2158 describes. The clamp on `:123` only stops it from going backwards past the fact the iterator started from.

The marker is supposed to mean "the lowest issue still worth revisiting", and on the next run `(gt issue $before)` on `:28` skips every `issue-was-opened` fact below it. When the response happens to end on a high number, the facts in between are passed over until the sorted enumerator exhausts and the iterator restarts from zero.

Deleting `:92` and letting the block return the `issue` it was handed, or setting the cursor from `seen.max` once the loop is done, both give the marker a defined meaning and make the clamp on `:123` unnecessary. A side effect on the enclosing cursor from inside a fact builder is invisible to anyone reading either piece.

@yegor256 please take a look.

Thayorns:
The judge set the iterator's cursor from inside the fact builder. `Fbe.if_absent` yields before it looks for an existing fact, so that assignment ran for every item in the search response. As a result, `min_issue_was_found` was saved as whatever issue GitHub put last in the response, and the next run skipped every `issue-was-opened` fact below that number.

The block now returns the issue the iterator handed it, so the marker still means the lowest issue worth revisiting. The clamp against the starting issue and the extra copy of that issue were only there to make up for the side effect, so I removed them. The log line now shows the one issue the scan started from.

A new test sets the marker, makes the response end on a random higher issue number, and checks that the marker stays on the iterated issue. Paging through search results is still open under #2158.

Closes #2177


Thayorns:
@yegor256 take a look please, happy to clarify anything about the marker change

Thayorns:
@yegor256 addressed, take another look please

PR #2255, opened 2026-09-15
fix(#2111): require fbe/fb instead of unused fbe/consider in erase-repository

### LINKED ISSUE #2111: `erase-repository.rb` requires `fbe/consider` and never requires `fbe/fb`
`judges/erase-repository/erase-repository.rb:7` requires `fbe/consider`, but `Fbe.consider` is never called in the file — the judge drives its own `Fbe.fb.query(...).each` loop at line 14 instead. The require is dead.

The reverse is also true. `Fbe.fb` is called at lines 14 and 36, and `fbe/fb` appears in no `require` in this file. It resolves only because `fbe/consider.rb:8` happens to `require_relative 'fb'`, so the judge depends on an internal detail of a gem file it has no business loading. Drop the dead require and the judge stops loading. `judges/who-is-alive/who-is-alive.rb:7` requires both explicitly and does not have this coupling.

Replacing `require 'fbe/consider'` with `require 'fbe/fb'` on line 7 leaves the file requiring exactly what it uses.

@yegor256 please take a look.

Thayorns:
The erase-repository judge required fbe/consider but never called Fbe.consider. It runs its own Fbe.fb query loop instead. At the same time it never required fbe/fb, even though it calls Fbe.fb twice. That only worked because fbe/consider happens to load fbe/fb internally. Now line 7 requires fbe/fb, so the judge loads exactly what it uses.

The existing tests could not catch this, because the test helper and other judges already load fbe/fb into the same process. The new test runs the judge in a fresh Ruby process. An fbe/consider.rb on the load path raises if it is ever required. Without the fix, the test fails at line 7 of the judge. With the fix, it passes.

Closes #2111


Thayorns:
@yegor256 take a look please, happy to clarify anything about the isolated-process test.

Thayorns:
@yegor256 handled, have a look

PR #2256, opened 2026-09-15
#2167: sleep out the search window instead of giving up once the budget is spent

### LINKED ISSUE #2167: `total_active_contributors` gives up for good past 25 repositories in `lib/qos_search.rb`
`Jp.qosearch` in `lib/qos_search.rb:35` returns `nil` once a judge has spent `Jp::SEARCH_WINDOW_BUDGET` search calls — 25, set on line 10 — inside a 60-second window. `judges/dimensions-of-terrain/total_active_contributors.rb:16` issues exactly one search per repository, and line 30 turns any `nil` into `return {}`, discarding every author counted up to that point.

The twenty-sixth repository therefore ends the measurement. When the `repositories` mask expands to more than 25 non-archived repositories, the loop spends the budget, the next call comes back `nil`, and the metric yields `{}`. `lib/incremate.rb:22` leaves the property unset, the next cycle walks the same list and stops in the same place, and the number is never written at all. Nothing waits for the window to roll over either — the 25 calls go out back to back in well under a second, so the 60 seconds on line 9 never elapse.

The budget is a self-imposed pacer that keeps the judge under GitHub's 30-per-minute search allowance, not a refusal from GitHub, so `Jp.qosearch` should sleep out the remainder of the window and dispatch the call instead of giving up on it. #1983 changed what this budget charges per call; it left untouched what a caller does once the budget is gone.

@yegor256 please take a look.

Thayorns:
Once a judge had spent its 25 search calls inside a 60-second window, Jp.qosearch returned nil for every further call. The budget is our own pacer that keeps us under GitHub's 30-per-minute search allowance, not a refusal from GitHub, yet callers treated the nil as a failure. total_active_contributors issues one search per repository and bails out with an empty hash on nil, so with more than 25 repositories the metric was never written, cycle after cycle.

Now, when the budget is spent, Jp.qosearch sleeps for whatever is left of the window, opens a fresh window with a zero count, and dispatches the call. The quota latch, the rate limit check and the transient error handling stay as they were.

The old test that asserted nil after the budget is replaced by three tests: the call is dispatched after the budget is spent, the sleep lasts only the rest of the window, and the next window gets a full budget, so there is exactly one sleep per spent window. Two other tests lose their trailing nil assertions after the budget, which would now sleep for real.

Closes #2167


Thayorns:
@yegor256 take a look please, happy to clarify anything about the new wait in Jp.qosearch.

Issue #1561, opened 2026-06-15
Vitals page shows repos that no longer exist — stale facts in factbase


VasilevNStas:
The vitals page at https://zerocracy.github.io/judges-action/zerocracy-vitals.html lists repositories that do not exist on GitHub:

- `zerocracy/zealot` — 404
- `zerocracy/j` — 404
- `zerocracy/eva` — 404
- `zerocracy/baza` — 404 (also hardcoded in `.github/workflows/zerocracy.yml:25`)

The generated page (pages-action) only renders data it receives — the root cause is in **judges-action**: the factbase accumulates facts from repos that no longer exist and never cleans them up.

Possible approaches:
1. When a repo returns 404/Not Found during data collection, remove its existing facts from the factbase
2. Add a new judge (similar to `eliminate-ghosts`) that prunes facts for deleted repositories
3. At minimum, remove `zerocracy/baza` from `.github/workflows/zerocracy.yml`'s explicit repo list

The factbase currently has no validation mechanism — it stores whatever it receives without checking whether the source repo still exists.

VasilevNStas:
@yegor256 plz review

0crat:
@vasilevnstas Thanks for suggesting an enhancement! You've earned +18 points. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) bugs too. Your running score is [+1864](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2257, opened 2026-09-15
#2105: revive a repository once GitHub answers it again

### LINKED ISSUE #2105: A repository marked stale by `erase-repository` is never revived
When `Fbe.octo.repository` raises `Octokit::NotFound`, `judges/erase-repository/erase-repository.rb:22` records the repository as gone and line 37 writes `stale = 'repository'` onto every fact that names it. GitHub answers 404, not 403, for a repository that has merely gone private or that the token has lost access to — it hides existence rather than admitting it — so an ordinary permission change is written into the factbase as a deletion.

Nothing ever reverses it. `Fbe.delete_one` is called in exactly two places across the whole tree, `judges/revive-user/revive-user.rb:25` and `judges/who-has-name/who-has-name.rb:108`, and both clear `stale`/`'who'`. Users have a `revive-user` judge that re-checks them and un-stales them when they come back; repositories have no equivalent. The `(absent stale)` filter in the query at line 14 then guarantees the fact is never looked at again, so a repository that was private for one afternoon stays dead for the life of the factbase.

A `revive-repository` judge shaped like `revive-user` would close it: query `(and (eq stale 'repository') (eq where 'github') (unique repository))`, call `Fbe.octo.repository` on each, and `Fbe.delete_one(f, 'stale', 'repository')` on the ones that answer. Note this pulls against #1561, which asks for the opposite — more aggressive pruning of dead repositories — so the two want deciding together.

@yegor256 please take a look.

Thayorns:
When GitHub answers 404 for a repository, `erase-repository` marks every fact of that repository as stale. But GitHub also returns 404 when a repository goes private or the token loses access to it. Nothing ever cleared that mark, so a repository that was hidden for an hour stayed dead in the factbase for good.

This adds a `revive-repository` judge, built like `revive-user`. For each stale repository it asks GitHub again. If the repository answers, the judge removes `stale` from all of that repository's facts. A 404, 410 or 403 leaves the mark in place.

The YAML judge test covers one repository that comes back and one that is still missing. A unit test also runs the judge against the fake GitHub server.

This works against #1561, which asks for dead repositories to be pruned more aggressively, so the two issues may need to be decided together.

Closes #2105


Thayorns:
@yegor256 take a look please, happy to clarify how this sits next to #1561.

Issue #2258, opened 2026-09-15
`quantity-of-deliverables` and `quality-of-service` metrics record a repository scan cut short by `Fbe.over?` as final


Thayorns:
This follows #2169, which covers the same defect in `judges/dimensions-of-terrain`. The metrics of `quantity-of-deliverables` and `quality-of-service` walk repositories with `Fbe.unmask_repos do |repo|`, and the loop in the gem (`fbe/unmask_repos.rb:101`, copied into `lib/patches/unmask_repos.rb:69`) ends with `break if Fbe.over?(...)` as soon as the run is off quota, out of lifetime or out of time. The block never learns about it, so `total_commits_pushed.rb:15`, `total_releases_published.rb:12` and `total_issues_created.rb:13` return sums over whatever prefix of the shuffled repository list they reached.

Both judges run the metrics through `Jp.incremate` with `avoid_duplicate: true` (`quantity-of-deliverables.rb:15`, `quality-of-service.rb:36`), and `lib/incremate.rb:22` skips every property the fact already has, so an undercounted `total_commits_pushed` or `total_hoc_committed` stays in the period fact for good. The `some_*` samples suffer the same way: `some_merged_pulls` stores per-repository counts, and `some_build_success_rate.rb:15`, `some_release_interval.rb:11` and `some_release_hoc_size.rb:13` have no stop check of their own at all. Others, like `some_merged_pulls.rb:14`, only return `{}` on `Fbe.octo.off_quota?`, which does not cover the lifetime and timeout conditions `Fbe.over?` also checks.

Once #2169 makes the patch return `false` for a scan it cut short, these metrics should return `{}` when `Fbe.unmask_repos` does, so the property stays absent and is collected on a later run. They also need to `require_relative '../../lib/patches/unmask_repos'`, since `judges` only loads the top-level `lib/*.rb` files and would otherwise fall back to the gem's version.

@yegor256 please take a look.

0crat:
[...]

PR #2260, opened 2026-09-15
fix(#2169): report an unfinished repository scan so terrain metrics record no partial total

### LINKED ISSUE #2169: A run that stops mid-scan in `unmask_repos.rb` writes a partial total as final
`lib/patches/unmask_repos.rb:69` ends the repository loop with `break if Fbe.over?(...)` as soon as the run is off quota, out of lifetime or out of time. The block being yielded to is never told that this happened, so every metric under `judges/dimensions-of-terrain` returns the total it had reached when the loop was cut short — `total_active_contributors.rb:36`, `total_repositories.rb:13`, `total_stars.rb:29`, `total_releases.rb:30`, `total_files.rb:48`, `total_contributors.rb:44`, `total_issues.rb:31` — as though it had seen the whole list.

`lib/incremate.rb:22` never revisits a property the fact already carries, so that short number is the one the factbase keeps for good. `repos.shuffle!` on line 65 sharpens it: the repositories that made it into the count differ from run to run, so the recorded value is not even a consistent prefix of the truth, and nothing in the judge's own log says a repository was missed.

This is not the whole-run case `Fbe.over?` already covers in `lib/incremate.rb:26`, which only stops the judge before it starts the next metric file, and it is not #1977, where the skipped repository is one that raised and was rescued. `Fbe.unmask_repos` should report that it stopped early — through a return value, or a second block argument — so a metric can return `{}` rather than a total it knows is short.

@yegor256 please take a look.

Thayorns:
The `unmask_repos` patch used to `break` out of its loop when `Fbe.over?` said the run was off quota, out of lifetime or out of time. The caller never learned that, so every `dimensions-of-terrain` metric returned the total it had reached, and `Jp.incremate` kept that short number for good.

The block form of the patch now returns `false` when it stops before the last repository, including the check before the scan starts, and `true` once every repository was yielded. The form without a block still returns the list. Each `total_*` metric in `dimensions-of-terrain` returns `{}` when the scan did not finish, so the property stays absent and the next run measures it again. In `total_files` the truncated-tree case now returns `{}` directly, which makes its separate flag unnecessary.

The same gap in `quantity-of-deliverables` and `quality-of-service` is left for #2258.

Closes #2169


Thayorns:
@yegor256 happy to clarify anything about how the scan now reports that it stopped early.

Thayorns:
@yegor256, fixed, have a look

PR #2262, opened 2026-09-15
#2251: skip last commit when compare returns no commits

### LINKED ISSUE #2251: github-events crashes when GitHub compare returns no commits
## Summary
The `github-events` judge assumes that every successful compare response contains at least one commit. A comparison of identical revisions can legitimately return an empty `commits` array, and the judge then raises instead of recording the release statistics.

## Reproduction
Give `github-events` a release whose previous and current tags resolve to the same commit (or mock `Fbe.octo.compare` with:

```ruby
{ total_commits: 0, files: [], commits: [] }
```

Then execute `info` for that release.

## Actual result
`info` evaluates `json[:commits].first[:sha]`. With an empty array, `first` is `nil` and the judge raises `NoMethodError` while processing an otherwise valid API response.

## Expected result
A zero-commit comparison should be handled as a valid no-op: retain the total and file-change values and leave `last_commit` unset (or use the previous revision), allowing the cycle to finish.

## Evidence
`judges/github-events/github-events.rb`, method `info`, dereferences `json[:commits].first[:sha]` without checking whether the API returned any commits. GitHub can return an empty comparison when both refs point at the same commit, which is also the scenario that produces `total_commits: 0`.

Thayorns:
The github-events judge read the SHA of the first commit from every compare response. A comparison of two refs that point at the same commit is answered by GitHub with an empty commits array, so the release handling crashed with NoMethodError and aborted the scan of the whole repository.

Now info keeps the total commits and the hits of code from the response, and leaves last_commit unset when the comparison has no commits.

Two tests stub such a comparison for a published release and check that the fact records zero commits and carries no last_commit.

Closes #2251


Thayorns:

PR #2264, opened 2026-09-15
Keep sub-second precision of workflow run duration

### LINKED ISSUE #2252: some_build_success_rate rounds sub-second workflow durations down to zero
## Summary
The build-success-rate judge converts GitHub's `run_duration_ms` with integer division. Workflows that finish in under one second are recorded as zero seconds, which distorts duration and MTTR metrics and can make their completion time equal to their start time.

## Reproduction
Return a workflow usage payload with `run_duration_ms: 999` from `Fbe.octo.workflow_run_usage`.

## Actual result
`ms / 1000` evaluates to `0` because both operands are integers in Ruby. The run is stored with `secs: 0`, and its `completed` timestamp is unchanged from `run_started_at`.

## Expected result
Preserve the millisecond precision (for example, divide by `1000.0`) or apply an explicit rounding policy so a 999 ms run is not reported as a zero-duration build.

## Evidence
`judges/quality-of-service/some_build_success_rate.rb` assigns `secs = ms / 1000` before calculating `completed` and appending `some_build_duration`/`some_build_mttr`. GitHub's field is explicitly milliseconds, while Ruby's integer division truncates the value.

Thayorns:
The build success rate judge turned GitHub's `run_duration_ms` into seconds with `ms / 1000`. Ruby drops the remainder when it divides one integer by another, so any run under a second was recorded as zero seconds and its completion time matched its start time.

This divides by `1000.0` instead, so the duration keeps its fractional part. The duration facts already hold floats, so nothing downstream needs to change.

Besides the duration itself, the zero also broke how runs are ordered. A failed run and a successful run that start in the same second got the same completion time, so the failure could be paired with a success that actually finished before it, giving a repair time of zero. There is a test for each case: the sub-second duration, and the MTTR pairing within a single second.

Closes #2252


Thayorns:

Thayorns:
@yegor256 pushed a fix, take a look when you can

PR #2265, opened 2026-09-15
fix(#2163): convert the closure of a merged pull into a merge instead of deleting it

### LINKED ISSUE #2163: `fix-wrong-closures` deletes the fact of a merged pull and creates nothing instead
## What happens

```ruby
# judges/fix-wrong-closures/fix-wrong-closures.rb:62-73
next issue if json[:merged_at].nil? && json[:state] != 'open'
Fbe.fb.query(
  "(and (eq where 'github') (eq repository #{repository}) (eq issue #{issue}) (eq what 'pull-was-closed'))"
).delete!
$loog.info(
  "The closure of #{repo}##{issue} is wrong, it is " \
  "#{json[:merged_at].nil? ? 'open' : 'merged'} now, forgotten"
)
```

The condition covers two different cases. If the pull was reopened, deleting the `pull-was-closed`
fact is right, the event did not happen. If `merged_at` is set the event did happen, it was recorded
under the wrong name, and the right outcome is a `pull-was-merged` fact rather than no fact at all.

Nothing creates one. `github-events` only walks the recent events stream, so a merge older than the
event window is not rediscovered.

## Reproduction

Read the block above.

The merge disappears from the factbase along with everything other judges had computed for it: `hoc`
from `fix-missing-hoc`, `comments`, `comments_to_code`, `comments_by_author`,
`comments_by_reviewers`, `comments_appreciated`, `comments_resolved` from `fix-missing-comments`,
`branch` and `who`. All of it goes with `.delete!` and none of it is recomputed, because the fact
those judges would fix is gone.

## What should happen

When `merged_at` is present, convert the fact, write `what = 'pull-was-merged'` and set `when` from
`merged_at`, keeping the derived properties. Keep the `.delete!` for the reopened case only.

Thayorns:
`fix-wrong-closures` used to delete a `pull-was-closed` fact whenever GitHub reported the pull as merged. That threw away a real merge together with the hoc, comments, branch and who that other judges had already computed for it, and nothing recreated it, because `github-events` only walks the recent event window.

Now, when `merged_at` is present, the judge rewrites the fact through `Fbe.overwrite` with `what = 'pull-was-merged'` and `when` taken from `merged_at`, so every derived property survives. The fact is still deleted when the pull was reopened, and also when a `pull-was-merged` fact for the same pull already exists, so the conversion never produces a second merge.

The yml fixtures that expected merged closures to vanish are renamed and now expect converted facts. The unit tests cover the conversion, the merge time, a kept hoc and the already known merge.

Closes #2163


Thayorns:
@yegor256 take a look please, happy to clarify anything about the conversion

PR #2266, opened 2026-09-15
fix(#2158): search issues oldest first and log the total count in find-all-issues

### LINKED ISSUE #2158: `find-all-issues` reads one page of 30 results and reports it as a full scan

Thayorns:
The find-all-issues judge searched the whole life of a repository with a single call to `Jp.qosearch` and read one page of results with no sort and no order. On a busy repository it saw an arbitrary, relevance-ordered subset and never looked at `total_count`. The page actually holds 100 results rather than 30, because Fbe sets `per_page` to 100 on the client, but the subset was still arbitrary and not reproducible.

The search now asks GitHub for `sort: created` and `order: asc`, so every run looks at the oldest issues created since the iteration cursor and the cursor walks forward across runs instead of sampling at random. The summary line now says how many items were checked out of the `total_count` GitHub reported, so a truncated scan shows up in the log.

A new test stubs only the sorted query and checks that the oldest missing issue gets restored; the existing stubs were moved to the sorted query too.

Closes #2158


Thayorns:
@yegor256 take a look please, happy to clarify anything about the sorted search

Thayorns:
@yegor256 ready for review again

Issue #2148, opened 2026-09-10
`entry.sh` overrides a `repositories=` pair given inside `options`


morphqdd:
## What happens

```bash
# entry.sh:121-126
if [ -z "${INPUT_REPOSITORIES}" ]; then
    echo "The 'repositories' plugin parameter is not set, using current repository: ${GITHUB_REPOSITORY}"
    options+=("--option=repositories=${GITHUB_REPOSITORY}");
else
    options+=("--option=repositories=${INPUT_REPOSITORIES}");
fi
```

This append happens after the `INPUT_OPTIONS` parsing loop at lines 107 to 120, and
`Judges::Options#normalize_to_h` builds a hash from the pairs in order, so the last
`--option=repositories=` wins, `judges/lib/judges/options.rb:118-131`.

Every other overlapping option in the same file is handled the other way round. `github_token` at
156 to 161, `bots` at 181 to 186 and `sqlite_cache_min_age` at 195 to 200 all scan the `options`
array first and only add the dedicated input when the pair is absent. `repositories` is the only one
that does not scan.

`README.md:111-116` documents the pair as supported:

> `repositories=..` is a comma-separated list of masks (alternative to `repositories` action input)

## Reproduction

Set only `options: repositories=yegor256/*` and leave the `repositories` input empty. The value is
replaced by the single current repository, and the log at line 122 says the parameter "is not set".

The action then scans one repository instead of the configured list, and the log line points the user
at a setting they did supply.

## What should happen

Scan `options` for `--option=repositories=` first, exactly like `github_token` and `bots`, and fall
back to `INPUT_REPOSITORIES` or `GITHUB_REPOSITORY` only when it is absent.

0crat:
@morphqdd Thanks for reporting a new bug! You've earned +18 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+9266](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #2267, opened 2026-09-15
#2156: state the real options precedence per parameter in README

### LINKED ISSUE #2156: README states a precedence rule that contradicts the next paragraph and the code
## What happens

```
README.md:104-110

The dedicated action inputs above take precedence when a parameter
  is also configurable via `options` (the k=v pairs below exist for
  backward compatibility and advanced scenarios):

* `github_token=...` is a default GitHub token, usually to be set to
  `${{ secrets.GITHUB_TOKEN }}`; it takes precedence over the
  `github-token` action input when both are set
```

Line 104 says the input wins and line 109 says the pair wins, about the same parameter, three lines
apart. The code agrees with line 109:

```bash
# entry.sh:156-170
for opt in "${options[@]}"; do
    if [[ "${opt}" == "--option=github_token="* ]]; then
        github_token_found=true
...
if [ "${github_token_found}" == "false" ]; then
...
        options+=("--option=github_token=$(printenv "INPUT_GITHUB-TOKEN")");
```

The input is added only when the pair is absent, so the pair wins. `bots` at `entry.sh:181-192` has
the same shape. `repositories` behaves the third possible way, the input is appended last and always
wins, although the README documents the pair as an alternative.

## Reproduction

Read `entry.sh:155-173` and `:181-192` against `entry.sh:121-126`, and
`judges/lib/judges/options.rb:118-131`, where a plain hash makes the last pair win.

A user with `github-token: ${{ secrets.ZEROCRACY_PAT }}` as an input and an old
`github_token=${{ secrets.GITHUB_TOKEN }}` still sitting in `options` reads line 104, concludes the
PAT is in use, and cannot explain why private repositories are invisible. The low privilege token is
the one in effect.

## What should happen

State the real rule per parameter: the `options` pair wins for `github_token`, `bots` and
`sqlite_cache_min_age`, the action input wins for `repositories` and `sqlite-cache`.

Thayorns:
The README claimed that dedicated action inputs take precedence over the same parameters passed as k=v pairs in `options`, and three lines later said the opposite for `github_token`. The entry point decides per parameter: it adds the `github-token` and `bots` inputs only when no matching pair is present, so the pair wins, while it appends `repositories` and `sqlite_cache` after the pairs, and the judges gem keeps the last value of a repeated option, so the input wins.

The paragraph now states that rule per parameter, the `github_token=` bullet no longer contradicts it, the `repositories=` bullet says the input, or the current repository when the input is empty, overrides the pair, and the `bots=` pair gets its own bullet. The fact that a `repositories=` pair can never take effect is tracked separately in #2148.

Closes #2156


Thayorns:
@yegor256 take a look here please, when you have time; happy to clarify anything about the precedence wording.

PR #2268, opened 2026-09-15
fix(#2152): pass judge options to update through a private options file

### LINKED ISSUE #2152: Both tokens are passed on the command line, so they sit in the process table
## What happens

Both secrets are passed as command line arguments:

```bash
# entry.sh:170
options+=("--option=github_token=$(printenv "INPUT_GITHUB-TOKEN")");
# entry.sh:211
"--token=${INPUT_TOKEN}" \
```

and the same at lines 224, 285 and 317. They become the `argv` of the `bundle exec judges` process
and are readable from `/proc/<pid>/cmdline` by anything in the same container for the whole `update`,
up to `lifetime`, fifteen minutes by default.

The script already treats these strings as secrets for the log. `entry.sh:36-45` emits `::add-mask::`
for both, with a comment explaining the leak. The process table is the same class of exposure and is
left open.

The gem offers the alternative: `bin/judges:114-115` declares `--options-file`, "File containing
options, one key=value pair per line".

## Reproduction

Read the five call sites above and `bin/judges:113-115`, where `--option` and `--options-file` are
alternatives.

Every judge and every gem loaded into the update process runs beside a process whose command line
holds the Zerocracy token and the GitHub token. Anything that shells out and dumps `ps`, a crash
reporter, a debugging judge, a `qbash` call logging its environment, captures both in plaintext, and
`::add-mask::` does not redact a value that reaches the log by a path the runner did not scan.

## What should happen

Write the `--option=` pairs, `github_token` at least, into a mode 0600 temporary file and pass
`--options-file`, and hand the Zerocracy token to `judges` through the environment rather than
`--token=`.

Thayorns:
`entry.sh` used to hand `github_token` to `judges update` as an `--option` on its command line, whether it came from the `github-token` input or from `options`, so the GitHub token sat in `/proc/<pid>/cmdline` for the whole update, readable by any judge or gem running beside it.

Now that pair is written into a `mktemp` file, which is created with mode 0600, and passed through `--options-file`. The other options stay on the command line, so the echoed command still shows them and the entry checks that look for them keep working. A single `EXIT` trap removes the file, and the same trap now also cleans up the temporary judges directory of the dry mode, which used to have its own trap.

The new test runs `entry.sh` in dry mode against a fake `judges` and checks that the token never reaches argv, whether it comes from `github-token` or from `options`, that it does reach the options file, and that the file is private to its owner.

The Zerocracy token still travels as `--token`, because judges reads it only from that flag. yegor256/judges#583 asks for an environment fallback, and the four call sites can move once it lands.

Closes #2152


Thayorns:
@yegor256 take a look please, happy to clarify anything about the options file

PR #2282, opened 2026-09-15
#2125: sample the window the fact says it covers

### LINKED ISSUE #2125: `some_backlog_size` samples seven days and ignores `qos_days`
## What happens

The backlog sampler keeps the last seven days of whatever range it is handed:

```ruby
# judges/quality-of-service/some_backlog_size.rb:14
(fact.since.utc.to_date..fact.when.utc.to_date).last(7).each do |date|
```

The length of the window comes from `Fbe.pmp.quality.qos_days`, read in
`judges/quality-of-service/quality-of-service.rb:13` and used by `Jp.cover_qo` to size
`since..when`. Every other `some_*` metric walks that whole range. This one overrides it with a
literal `7` and nothing explains where the number comes from.

## Reproduction

Read from `some_backlog_size.rb:14`, `quality-of-service.rb:13` and `lib/cover_qo.rb:12-33`.

With `qos_days: 30` the backlog series is sampled from 7 of the 30 days and the first 23 are never
queried, while the fact still claims to cover the whole window. The gap-filling path in
`lib/cover_qo.rb:25-33` can also hand out windows longer than `slice`, so this happens at the default
setting too.

## What should happen

The metric should walk the whole `fact.since..fact.when` range, or the sample size should be a
configured option instead of a literal.

morphqdd:
The sampler kept the last seven days of whatever range it was handed, with a literal `7` and nothing saying where the number came from. The length of the window comes from `qos_days`, and every other `some_*` metric walks that whole range.

With `qos_days: 30` the series was sampled from 7 of the 30 days and the first 23 were never queried, while the fact still claimed to cover the whole window.

It walks `since..when` now, the range the fact carries.

Closes #2125


morphqdd:
Closing this one. `test_quality_of_service_some_backlog_size` asserts the series `[1, 2, 2, 2, 2, 1, 1]`, seven entries, and two more tests stub exactly the days that sampling asks for. The seven is pinned on purpose, not left over, so widening the window is a call for the maintainer rather than something to slip in behind a test rewrite.

What the issue describes is still true: the fact claims to cover `qos_days` while only the last seven days of it are sampled. I have written what I found on #2125.

PR #2290, opened 2026-09-15
#2120: resume at the oldest event a capped scan processed

### LINKED ISSUE #2120: Events past `max_events` are lost for good on the first scan of a repository
## What happens

The event cursor is chosen like this:

```ruby
# judges/github-events/github-events.rb:495-504
elsif id <= latest || latest.zero?
  $loog.info("... next time will scan until ##{first}")
  first
else
  ... id
end
```

`first` is the newest event id seen and `id` is the oldest one actually processed. When the loop
stops early because of `max_events` at line 452, the unscanned tail must be picked up next time, so
returning `id` is right. On the first scan of a repository `latest` is `0`, the `latest.zero?` arm
wins, and the cursor jumps to `first`.

The two arms are joined by `||` although they mean opposite things: `id <= latest` says the loop
reached events that are already known, `latest.zero?` says there is no cursor yet.

## Reproduction

Three `CreateEvent`s with ids 30, 20 and 10, `max_events: 2`, empty factbase:

```
MAX_EVENTS cursor=30 tags=["v30", "v20"]
```

Event 10 is never recorded and the cursor is already 30, so the next run starts above it and stops
at once. The event is lost for good, with nothing in the log to say so.

`test-action.sh:13` runs the action with `max_events=3`, so a real run reaches this. The repository's
own `test_stops_scanning_above_max_events` in `test/judges/test-github-events.rb:2071-2077` pins the
loss: it asserts `events_were_scanned: 14` and that `event_id: 11` is absent.

## What should happen

When the scan stopped early because of `max_events`, the oldest processed id should be returned
whether or not `latest` was zero. `first` belongs only to the case where the loop stopped because it
reached events that were already scanned.

morphqdd:
The two arms of the cursor choice were joined by `||` although they mean opposite things: `id <= latest` says the loop reached events that are already known, `latest.zero?` says there is no cursor yet. On the first scan of a repository `latest` is `0`, so the `latest.zero?` arm won even when `max_events` had cut the scan short, and the cursor jumped to the newest event seen. Everything past the cap was then never scanned, on that cycle or any later one.

The judge now records why the loop stopped, and a scan cut short by `max_events` resumes at the oldest event it actually processed.

The existing `test_stops_scanning_above_max_events` asserted `events_were_scanned: 14`, the newest of the four events, while event 11 was never scanned. That assertion is the defect written down, so it is updated to 12, the oldest of the three that were processed.

Opened as a draft: `judges/github-events/github-events.rb` is touched by #2262, #2093, #2090 and my own #2278.

Closes #2120


morphqdd:
Holding this one as a draft. Two packs, `quick-scan` and `scan-over-existing`, fail with it, and the reason is not the change.

`Fbe::FakeOctokit#repository_events` returns its three events as 123, 124, 125, in ascending order. The real endpoint returns events newest first, and the whole cursor logic here is written for that: `first` is meant to be the newest event seen and `id` the oldest one processed. Under the fake they are the other way round, so the numbers the packs pin say nothing about which of the two the judge should resume from.

So the packs cannot express the behaviour this fixes until the fake orders its events the way GitHub does. Updating them to the numbers the fake produces would pin the quirk rather than the contract.

I will file the fake ordering separately and come back to this one.

Issue #167, opened 2024-07-28
`commits`, `hoc`, and `last_commit` are missing in `release-published`


yegor256:
When we detect `release-published` fact in `github-events.rb`, we don't count commits and hoc of the release, but we should. Let's do it.

Also, now we don't record any information about its commits. Let's put the `last_commit` property into the fact with the SHA of the last commit that was released in this release.

Yegorov:
@yegor256 may I try this one?

yegor256:
@Yegorov sure, go ahead

Suban05:
@yegor256 please close

Issue #2269, opened 2026-09-15
`log_not_contains` passes for patterns starting with a dash in `makes/test-common.sh`


Thayorns:
`log_not_contains` in `makes/test-common.sh:53` runs `grep -qF "$pattern" log.txt`. When the pattern starts with `-`, grep reads it as an option, fails with "invalid option" and exits 2, so the `if` sees a failure and the helper reports that the pattern is absent.

`entries/masks-tokens-in-verbose-mode.sh:39` and `:41` pass `--option=github_token=...` and `--token=...`, which is exactly that shape, so the check that tokens never reach the verbose trace can never fail, whatever the log holds. A log file containing `--option=github_token=ghp_x` makes `log_not_contains "--option=github_token=ghp_x"` return 0. `log_contains` at line 46 has the same flaw, only it fails loudly instead.

Passing `--` before the pattern, `grep -qF -- "$pattern" log.txt`, in both helpers makes grep treat it as text. @yegor256 take a look please

0crat:
@thayorns Thanks for reporting a new bug! You've earned +18 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+2392](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #2114, opened 2026-09-09
`scan_repos.yml` asserts nothing about the fact with no `repository`


Thayorns:
The `expected` block in `judges/erase-repository/scan_repos.yml:30` asserts on `_id` 1, 2, 3, 4 and 6 and says nothing about `_id` 5. That fact, declared at line 22, is the one with `where: github` and no `repository` at all — the only input in the fixture that exercises the `(exists repository)` filter in the query at `judges/erase-repository/erase-repository.rb:14`.

`/fb[count(f)=6]` proves the fact survived, but survival is not the property under test. A change that dropped `(exists repository)` from the query would send `f.repository` in as `nil`, and the fixture would still pass every line it asserts. The one input placed there to catch that has no assertion behind it.

Adding `- /fb/f[_id=5 and not(repository) and not(stale)]` alongside the others closes it, and matches the shape the other five already use.

@yegor256 please take a look.

Thayorns:
@yegor256 take a look please

Thayorns:
@yegor256 can we accept this one?

Thayorns:
@yegor256, can we accept this issue?

Thayorns:
@yegor256, ready for you to accept whenever you get a chance

Issue #2108, opened 2026-09-09
The `Forbidden` arm in `erase-repository` logs a timing line that names no repository


Thayorns:
The block handed to `elapsed` in `judges/erase-repository/erase-repository.rb:18` ends with a `throw` in two of its three arms — line 21 for a repository that was found and line 24 for one that was not — and `elapsed` turns the thrown symbol into the message it prints (`elapsed.rb:56`, `print_it.call(tag.to_s)`). The `Octokit::Forbidden` arm at line 25 has no `throw`, so the block returns normally and `elapsed.rb:52` prints its default instead: `Finished in 3ms`, with nothing saying which repository took those three milliseconds.

The `warn` on line 26 does carry the id, but it is a separate line at a different level, so the INFO trace that the other two arms populate has a hole in it exactly where a slow forbidden repository would show up. `judges/eliminate-ghosts/eliminate-ghosts.rb:37` throws in its forbidden arm for this reason and keeps its trace complete.

Adding `throw(:"GitHub repository ##{r} is not accessible (transient)")` after the `warn` on line 29 lines this arm up with the other two.

@yegor256 please take a look.

Thayorns:
@yegor256 take a look please

Thayorns:
@yegor256 can we accept this one?

Thayorns:
@yegor256, can we accept this issue?

Thayorns:
@yegor256, would you accept this one?

PR #2298, opened 2026-09-15
#2263: take the newest commit of the comparison as last_commit

### LINKED ISSUE #2263: `info` stores the oldest commit as `last_commit` in `github-events.rb`
The `info` method in `judges/github-events/github-events.rb:80` sets `last_commit` of a `release-published` fact to `json[:commits].first[:sha]`. GitHub's compare API (`GET /repos/{owner}/{repo}/compare/{basehead}`) returns commits in chronological order, oldest first, so the fact gets the earliest commit after the previous release, not the head of the new one. For example, `compare/master~3...master` on this repository currently lists `beca5c68e` (2026-09-09) first and `af0136cac` (2026-09-14), the head of `master`, last.

This contradicts #167, which introduced `last_commit` as the SHA of the last commit released. The tests don't catch it: the stubbed compare responses in `test/judges/test-github-events.rb` (around lines 963-1003) are not in GitHub's order, and the assertions at lines 1011 and 1014 pin `last_commit` to the first element.

Reading `json[:commits].last[:sha]` instead records the newest commit of the comparison, and the fixture should list its commits oldest first so the assertions follow the real API.

@yegor256 take a look please

opusstudiohq-max:
`info` reads `json[:commits].first[:sha]`. The compare API returns the commits between base and head in chronological order, oldest first, so the first element is the earliest commit after the previous release and the last one is the head of the new release. `last_commit` therefore held the oldest commit of the range, not the newest.

`json[:commits].last[:sha]` records the head of the comparison, which is what #167 introduced the property for.

The two fixtures in `test_release_event_contributors` listed their commits newest first, which is not the order the API uses, and the first of them also included the base commit itself, which the API leaves out. Both are reversed now so they read oldest first. The two `last_commit` assertions are untouched and still pass, because the sha each one pins moved from the front of the list to the back.

One assertion did have to follow. `contributors` collects author ids into a `Set`, which keeps insertion order, so reversing the second fixture changes the order in which the three authors first appear. The ids are the same three; the expected array is now `[2_566_462, 2_566_464, 2_566_463]`. Nothing else in the test moved.

Verified on my fork first: `rake`, `make` and the rest green.

Closes #2263

---


Thayorns:
@opusstudiohq-max you are again opening a pull request against an issue that was never triaged. This one resolves #2263, which carries no label.

- #2301 &rarr; #2269
- #2300 &rarr; #2108
- #2299 &rarr; #2106
- **#2298 &rarr; #2263 (this one)**

opusstudiohq-max:
You are right, and the problem is larger than the four you listed.

I checked every open pull request I have across this orbit against the rule you state. Fifteen of them target an issue that carries no label:

- `zerocracy/judges-action`: #2299 → #2106, #2300 → #2108, #2325 → #2114
- `zerocracy/fbe`: #953 → #886, #954 → #893, #955 → #892, #959 → #887, #960 → #895
- `zerocracy/pages-action`: #876 → #792, #877 → #791, #878 → #793
- `yegor256/factbase`: #933 → #848, #934 → #849, #935 → #850
- `yegor256/judges`: #595 → #583

I am closing all fifteen, not only the ones in this repository. The branches stay on my fork, and I will reopen each one if and when its issue is triaged.

Two small corrections for the record, in case they were labelled after you wrote: #2269 and #2263 both carry `bug` now, so #2301 and #2298 are not in the set above. #2114 is, and you did not list it — I found it applying your rule to my own queue.

I also checked whether this was a convention specific to this repository before acting on it, and it is not: every recently merged pull request in `zerocracy/fbe` and `zerocracy/pages-action` closes an issue labelled `bug` or `bug/good-title`. So the norm is clear and I had simply not been reading for it. My screening checked whether an issue was already claimed by another pull request; it did not check whether the project had accepted the issue in the first place. That is the part I got wrong, and it is now a hard gate in how I pick work.

Thank you for taking the time to write it out rather than just closing them.

opusstudiohq-max:
Same point as on #2301: opening these ahead of triage was the wrong call, and I'll wait for a label from now on. This one's issue (#2263) has since been labelled `bug`, so it now points at an accepted defect — happy to leave it for review, or to close it if you'd prefer.

PR #2299, opened 2026-09-15
#2106: ask about the quota once per repository, not once per fact

### LINKED ISSUE #2106: `erase-repository` calls `off_quota?` once per fact instead of once per repository
The loop in `judges/erase-repository/erase-repository.rb:15` checks `Fbe.octo.off_quota?` before it checks its own cache at line 17, so the quota check runs for every fact while the repository lookup it guards runs only for every distinct repository. That check is not free: `off_quota?` in `fbe/octo.rb:170` opens with `rate = @origin.rate_limit!`, unconditionally, and octokit's `rate_limit!` in `octokit/client/rate_limit.rb:35` is `get 'rate_limit'` — a network round trip on every call, whatever the rate-limit middleware already knows.

The scale is in the log pasted into #700: 7685 facts spread over ten distinct repositories. The `good` hash added for that issue brought the `repositories/` calls down to ten, but the quota check in front of it still fires 7685 times, so the judge makes three orders of magnitude more requests than the work needs. `judges/eliminate-ghosts/eliminate-ghosts.rb:18` has the same two checks in the opposite order and does not pay this.

Swapping lines 15 and 17 fixes it — read `r = f.repository` first, `next unless good[r].nil?`, and only then ask whether there is quota left to spend on a repository that has not been seen yet.

@yegor256 please take a look.

opusstudiohq-max:
The loop asked `Fbe.octo.off_quota?` before it asked its own `good` hash whether the repository had been seen. The quota check guards the `Fbe.octo.repository(r)` call, and that call already runs once per distinct repository, so the guard ran once per fact for no reason. `judges/eliminate-ghosts/eliminate-ghosts.rb:18` reads the repository and consults its cache first, and this brings `erase-repository` to the same shape.

Behaviour is unchanged. A repository is still never fetched without quota, and a fact is still skipped when the quota is gone, since a fact whose repository has not been seen yet always reaches the check.

One thing in the issue I could not reproduce, and it seems fair to say so rather than repeat it. The issue expects one HTTP round trip per `off_quota?` call, from `rate = @origin.rate_limit!` at `fbe/octo.rb:170` in fbe 0.49.1, which is the version this repository locks. I ran six facts on a single repository against a stubbed `/rate_limit`, with and without this change, and both made exactly one request to it; adding `Cache-Control: no-store` to the stub changed nothing. Something in the fbe stack — `Faraday::HttpCache` is in the builder at `octo.rb:113` — answers the repeated calls without going out. So I am claiming fewer calls, not fewer requests, and I dropped the test I had written for it, since a test that passes either way says nothing.

Closes #2106

---


Thayorns:
@opusstudiohq-max you are again opening a pull request against an issue that was never triaged. This one resolves #2106, which carries no label.

- #2301 &rarr; #2269
- #2300 &rarr; #2108
- **#2299 &rarr; #2106 (this one)**
- #2298 &rarr; #2263

opusstudiohq-max:
Closing this until #2106 is triaged. It carries no label, so the project has not accepted it as a defect yet, and a pull request opened ahead of triage asks for review work on a problem nobody has agreed to fix — and goes stale at a cost to the reviewers and reporters here, not only to me. Raised by @Thayorns on zerocracy/judges-action#2298, and it applies to my whole queue, not only to that repository. The branch stays on my fork; I will reopen this if and when #2106 gets a label.

PR #2300, opened 2026-09-15
#2108: name the repository in the timing line of the forbidden arm

### LINKED ISSUE #2108: The `Forbidden` arm in `erase-repository` logs a timing line that names no repository

opusstudiohq-max:
`elapsed` prints the symbol the block throws, and falls back to `Finished in Nms` when the block returns normally. Two of the three arms throw and name the repository; the `Octokit::Forbidden` arm did not, so the INFO trace had a line with a duration and nothing to attach it to, which is exactly where a slow forbidden repository would sit. `judges/eliminate-ghosts/eliminate-ghosts.rb:37` throws in its forbidden arm for the same reason.

The `throw` goes after `good[r] = true` so the arm still records the repository as one not to retry in this cycle. The `warn` on the line above is untouched.

The test reads the judge's log through `Loog::Buffer` and looks for `GitHub repository #403123 is not accessible`. That phrase exists only in the thrown symbol: without this change the trace line reads `Finished in Nms`, and the only other mention of the repository is the `warn`, which says `Access forbidden to`.

Verified on my fork first: `rake`, `make` and the rest green.

Closes #2108

---


Thayorns:
@opusstudiohq-max you are again opening a pull request against an issue that was never triaged. This one resolves #2108, which carries no label.

- #2301 &rarr; #2269
- **#2300 &rarr; #2108 (this one)**
- #2299 &rarr; #2106
- #2298 &rarr; #2263

opusstudiohq-max:
Closing this until #2108 is triaged. It carries no label, so the project has not accepted it as a defect yet, and a pull request opened ahead of triage asks for review work on a problem nobody has agreed to fix — and goes stale at a cost to the reviewers and reporters here, not only to me. Raised by @Thayorns on zerocracy/judges-action#2298, and it applies to my whole queue, not only to that repository. The branch stays on my fork; I will reopen this if and when #2108 gets a label.

PR #2301, opened 2026-09-15
#2269: stop grep from reading a dashed pattern as an option

### LINKED ISSUE #2269: `log_not_contains` passes for patterns starting with a dash in `makes/test-common.sh`

opusstudiohq-max:
`grep -qF "$pattern"` reads a pattern that starts with a dash as an option, exits 2 without looking at the file, and `log_not_contains` takes that failure as proof the pattern is absent. Passing `--` before the pattern makes both helpers treat it as text.

Turning the helper on made `masks-tokens-in-verbose-mode.sh` fail, and what it caught is the check itself rather than a leak. `entry.sh:37-40` says in its own comment that `set -x` emits the full command line including `--option=github_token=...`, and that the protection is the `::add-mask::` workflow command emitted before tracing starts. So the token is in the trace by design, the runner redacts it, and `log_not_contains "--option=github_token=${github_token}"` asked for something the design never promised. It could not fail before and could not pass afterwards.

The two assertions now say what their own error messages already said: the mask must come first. `log_precedes` takes two patterns and fails when the second appears above the first, so the check is on the order of the two lines rather than on the presence of one of them. On the current build the mask lands at line 10 and the traced command at line 105.

Worth noting while it is in view: `--token=${zerocracy_token}` never reaches the trace at all, so only one of the two assertions was ever describing something real.

`entries/log-helpers-handle-dashed-patterns.sh` covers both helpers against a log file it writes itself, so it needs no image and runs in the same pass as the others. It fails on `master` at the first assertion, `log_contains` reporting a line it is looking straight at.

Verified on my fork first: `rake`, `make`, `shellcheck`, `bashate` and the rest green, with `masks-tokens-in-verbose-mode.sh` passing on its own terms.

Closes #2269

---


Thayorns:
@opusstudiohq-max you are again opening a pull request against an issue that was never triaged. This one resolves #2269, which carries no label.

- **#2301 &rarr; #2269 (this one)**
- #2300 &rarr; #2108
- #2299 &rarr; #2106
- #2298 &rarr; #2263

opusstudiohq-max:
You're right, and thanks for putting it plainly. Opening pull requests against issues that hadn't been triaged, several in a short burst, was the wrong call: an unlabelled issue isn't an accepted defect, and pulls that sit ahead of triage put review work onto problems nobody agreed to fix and risk the stale-point cost you mention, which lands on this repo and not only on me. I'll wait for a label before opening a pull request from here on.

On the four you listed: #2299 and #2300 are already closed. This one and #2298 have since been labelled `bug` (#2269 and #2263 were triaged in the meantime), so they now point at accepted defects. I'm happy to leave them for review, or to close them as well if you'd still rather they hadn't been opened ahead of triage — your call.

PR #2325, opened 2026-09-16
#2114: assert the fact that has no repository

### LINKED ISSUE #2114: `scan_repos.yml` asserts nothing about the fact with no `repository`

opusstudiohq-max:
`_id` 5 is the only input in the fixture with `where: github` and no `repository`, so it is the one fact that exercises `(exists repository)` in the query at `judges/erase-repository/erase-repository.rb:14`. The `expected` block asserted on 1, 2, 3, 4 and 6 and said nothing about it.

`/fb[count(f)=6]` only proves it survived. Drop `(exists repository)` from the query and the fact goes into the loop with `f.repository` as `nil`; the fixture would still pass every line it asserted.

`- /fb/f[_id=5 and not(repository) and not(stale)]` now says what that input is there to say, in the same shape as the five beside it.

Verified locally: `judges --verbose test --disable live --lib lib --judge erase-repository judges` — `erase-repository/scan_repos OK`, 1 judge and 1 test passed.

Closes #2114

---


opusstudiohq-max:
Closing this until #2114 is triaged. It carries no label, so the project has not accepted it as a defect yet, and a pull request opened ahead of triage asks for review work on a problem nobody has agreed to fix — and goes stale at a cost to the reviewers and reporters here, not only to me. Raised by @Thayorns on zerocracy/judges-action#2298, and it applies to my whole queue, not only to that repository. The branch stays on my fork; I will reopen this if and when #2114 gets a label.

PR #2330, opened 2026-09-17
#2303: guard changed_files the way additions and deletions are guarded

### LINKED ISSUE #2303: `some_pull_hoc_size` writes `nil` into `some_pull_files_size` and aborts the judge
## What happens

`judges/quality-of-service/some_pull_hoc_size.rb:31-32`:

```ruby
hocs << ((pull[:additions] || 0) + (pull[:deletions] || 0))
files << pull[:changed_files]
```

`additions` and `deletions` were given `|| 0` guards by the fix for #1648 and #1652. `changed_files` on the very next line was left alone. When a pull payload carries no `changed_files`, the metric returns `some_pull_files_size: [nil]`, and `Jp.incremate` (`lib/incremate.rb:45`) hands that to Factbase, which refuses it and takes the whole `quality-of-service` run down.

```
result = {some_pull_hoc_size: [8], some_pull_files_size: [nil]}
write FAILED: ArgumentError: The value of 'some_pull_files_size' can't be nil
```

## What should happen

Guard it the same way the line above it is guarded, or skip the pull instead of recording an absent value.

morphqdd:
`additions` and `deletions` are guarded with `|| 0`, `changed_files` on the
next line was not, so a pull payload without it wrote `nil` into
`some_pull_files_size` and Factbase refused the whole fact, taking the
judge down.

Closes #2303


morphqdd:

PR #2331, opened 2026-09-17
#2302: check the search quota before a Search API call

### LINKED ISSUE #2302: Five quality-of-service metrics check the core quota before a Search API call
## What happens

Five metrics guard a `Jp.qosearch` call with a bare `Fbe.octo.off_quota?`:

* `judges/quality-of-service/some_merged_pulls.rb:14,19`
* `judges/quality-of-service/some_issue_lifetime.rb:15`
* `judges/quality-of-service/some_pull_hoc_size.rb:14`
* `judges/quality-of-service/some_review_time.rb:17`
* `judges/quality-of-service/some_triage_time.rb:16`

In fbe the signature is `off_quota?(threshold: nil, resource: :core)` with `threshold ||= resource == :search ? 5 : 50`. So the bare call asks about the 5000 per hour core budget with a threshold of 50, while the call it protects spends from the 30 per minute search budget. `some_backlog_size.rb:11,15`, in the same directory, is the only one that writes `off_quota?(resource: :search)`.

Both directions are wrong. With four search requests left the guard says go ahead, and `Jp.qosearch` only refuses at exactly zero (`lib/qos_search.rb:61`), so the run keeps firing searches into an empty search budget. With the search budget untouched but core below 50, the metric returns `{}` and writes nothing, dropping a whole cycle of data for a resource it never touches.

Probe against the same GitHub state seen by both calls:

```
off_quota? core=false search=true
```

This is not #2175, which is about the fallback inside `Jp.qosearch` itself. These are the call sites in the metric files.

## What should happen

These five should pass `resource: :search`, the way `some_backlog_size` already does.

morphqdd:
The guards in front of a `Jp.qosearch` call asked about the core budget with
a threshold of 50, while the call they protect spends from the search budget.
They now pass `resource: :search`, the way `some_backlog_size` already does.

Closes #2302


morphqdd:

PR #2332, opened 2026-09-17
#2305: rescue the exceptions Faraday actually raises in the search call

### LINKED ISSUE #2305: The transient rescue list in `Jp.qosearch` names five exceptions that Faraday never lets through
## What happens

`lib/qos_search.rb:76-80`:

```ruby
rescue Octokit::ServerError,
  Net::OpenTimeout, Net::ReadTimeout, SocketError,
  Errno::ECONNRESET, Errno::ETIMEDOUT => e
```

Faraday's net_http adapter lists `Errno::ECONNRESET`, `SocketError` and `Net::OpenTimeout` among its own exceptions and re-raises each one as `Faraday::ConnectionFailed`; `Timeout::Error`, the parent of `Net::ReadTimeout`, and `Errno::ETIMEDOUT` are re-raised as `Faraday::TimeoutError`. None of the five raw names can reach this rescue, and the two classes that do reach it are not in the list. A transient network failure therefore escapes `Jp.qosearch` and crashes the judge instead of returning `nil` for a retry next cycle.

Probe with Octokit pointed at a closed port, real socket, no stubs:

```
ESCAPED from Jp.qosearch: Faraday::ConnectionFailed
```

The same mistake in `code-was-reviewed.rb` was accepted and fixed as #1968. The fix was not carried over here.

One more, in the clause just above: `rescue Octokit::TooManyRequests` catches only the 403 whose body matches `rate limit exceeded` or `secondary rate limit`. A 403 whose body says `abuse` becomes `Octokit::AbuseDetected`, which is what GitHub returns for hammering the Search API, and nothing in this chain catches it.

## What should happen

Rescue `Faraday::ConnectionFailed` and `Faraday::TimeoutError`, and catch `Octokit::Forbidden` rather than only its `TooManyRequests` subclass.

morphqdd:
Faraday re-raises the socket and timeout exceptions as
`Faraday::ConnectionFailed` and `Faraday::TimeoutError`, so none of the five
names in the rescue could ever be reached and a transient failure crashed
the judge. The rescue now names the two classes that do arrive, and catches
`Octokit::Forbidden` instead of only its `TooManyRequests` subclass, so an
abuse 403 from the Search API stops the searches too.

Closes #2305


morphqdd:

PR #2333, opened 2026-09-17
#2307: skip a pull that has no merge time

### LINKED ISSUE #2307: `some_review_time` dereferences `pull_request[:merged_at]` with no guard
## What happens

`judges/quality-of-service/some_review_time.rb:36`:

```ruby
Integer(pr[:pull_request][:merged_at] - first[:submitted_at])
```

Two levels into a search payload with no check on either. Its neighbour `some_issue_lifetime.rb:21-22` guards both `closed_at` and `created_at` on the same kind of payload before subtracting. This is the only place in the judge that trusts a nested search field.

When an item comes back without `merged_at`, the metric raises and the whole `quality-of-service` run dies rather than skipping one pull.

```
CRASH: NoMethodError: undefined method '-' for nil
```

## What should happen

Skip the pull when `merged_at` is absent, the way the neighbouring metric skips an issue with no `closed_at`.

morphqdd:
`pull_request[:merged_at]` was dereferenced with no check, so a search item
without it took the whole `quality-of-service` run down. The pull is now
skipped, the way `some_issue_lifetime` skips an issue with no `closed_at`.

Closes #2307


morphqdd:

PR #2334, opened 2026-09-17
#2310: check the deadline after the pause, not before it

### LINKED ISSUE #2310: `Jp.incremate` checks the deadline and then sleeps past it
## What happens

`lib/incremate.rb:26-30`:

```ruby
break if Fbe.over?(epoch:, kickoff:)
if pause.positive? && evaluated.positive?
  sleep(pause)
end
```

The deadline is checked before the pause, not after it. `Fbe.over?` trips at ninety per cent of `lifetime` or `timeout`. With `qos_pause_seconds` set, a run with ten seconds of slack left passes the check, sleeps the full pause, and then starts a metric that is already past the deadline. The guard cannot see the delay the same method is about to introduce.

## What should happen

Sleep first and re-check `Fbe.over?` afterwards, or subtract the pause from the remaining budget before deciding to continue.

morphqdd:
The deadline was checked before the pause, so a run with a few seconds of
slack slept the whole pause and then started a metric that was already past
the deadline. The pause now happens first and the check after it.

Closes #2310


morphqdd:

PR #2335, opened 2026-09-17
#2306: report a coverage gap that is left unmeasured

### LINKED ISSUE #2306: `Jp.cover_qo` never fills a coverage gap shorter than one window
## What happens

`lib/cover_qo.rb:40` reads `large = gaps.reject { |g| g[:when] - g[:since] < slice }`, keeping only gaps that are at least one full window long. With the default `qos_days` of 32, a hole of up to 31 days in the factbase is dropped: no fact is inserted for it and no metric ever measures that stretch. The log line at `:46` reports only the holes that were filled, so a permanently uncovered period leaves no trace.

```
hole 2025-02-02..2025-03-05 (31 days), qos_days=32 -> facts before=2 after=2
hole 2025-02-02..2025-03-07 (33 days)              -> facts before=2 after=3
```

#2049 touches the same file but reports a different mechanism, a window contained inside another one swallowing the gap after it.

## What should happen

Every gap should be filled, or the code should say why a gap shorter than one window is deliberately left unmeasured.

morphqdd:
A hole shorter than one window is skipped on purpose, and the existing tests
say so, but nothing in the log said it, so an uncovered stretch left no
trace at all. Such a gap is now reported with its dates and the reason.

Closes #2306


morphqdd:

PR #2336, opened 2026-09-17
#2314: query workflow runs with exact timestamps, not whole days

### LINKED ISSUE #2314: `total_builds_ran` truncates both ends of its range to whole days, so the boundary day is counted in two slices
## What happens

`judges/quantity-of-deliverables/total_builds_ran.rb:16` builds the range as

```ruby
"#{fact.since.utc.iso8601[0..9]}..#{fact.when.utc.iso8601[0..9]}"
```

that is, two dates with the time dropped. GitHub reads `created:A..B` as inclusive on both whole days. Consecutive slices share an endpoint, because `lib/cover_qo.rb:37` fills gaps so that slice N's `when` is slice N+1's `since`. The shared day therefore lands in both queries and every run on that day is counted twice, in two different facts.

The truncation also drags in runs from outside the slice whenever a boundary is not midnight: a slice starting at `12:00` picks up the whole morning as well.

Probe with a recording octokit stub and two adjacent facts:

```
slice1 created=2023-01-10..2023-01-11
slice2 created=2023-01-11..2023-01-13
2023-01-11 is inside both
```

#1345 asked for this range and it was implemented. The day truncation is what the fix left behind.

## What should happen

The range should use the exact timestamps, so that each workflow run belongs to exactly one slice.

morphqdd:
The range dropped the time, and GitHub reads `created:A..B` as inclusive on
both whole days, so the day two adjacent slices share was counted in both of
them. The query now carries the exact timestamps.

Closes #2314


morphqdd:

PR #2337, opened 2026-09-17
#2322: count only published releases, the way total_releases_published does

### LINKED ISSUE #2322: `total_releases` counts draft releases, while `total_releases_published` excludes them
## What happens

`Fbe.octo.releases(repo)` returns draft releases to any token with push access on the repository, which the judges' token has. `judges/dimensions-of-terrain/total_releases.rb:15` iterates them and the loop at line 27 increments for every element, with no filter.

The other release metric in the codebase does filter: fbe's `total_releases_published` counts only entries where `isDraft` is false and `publishedAt` is set. So `dimensions-of-terrain` and `quantity-of-deliverables` disagree about what a release is.

A repository that keeps a long-lived draft release, the usual pattern for release-notes automation, reports a permanently inflated `total_releases`, and the number moves when a draft is deleted even though nothing was ever released.

## What should happen

`total_releases` should count published, non-draft releases, matching `total_releases_published`.

morphqdd:
`Fbe.octo.releases` returns drafts to a token with push access, and the loop
counted every element, so a repository keeping a long-lived draft reported a
number that no release ever produced. Drafts and entries with no
`published_at` are now skipped, which is what `total_releases_published`
counts.

Closes #2322


morphqdd:

PR #2338, opened 2026-09-17
#2319: keep pull requests out of the issue count

### LINKED ISSUE #2319: `total_issues_created` adds pull requests into the issue count
## What happens

fbe returns issues and pull requests as two separate search counts, `type:issue` and `type:pr`, which GitHub keeps disjoint. `judges/quantity-of-deliverables/total_issues_created.rb:31` writes

```ruby
issues += json['issues'] + json['pulls']
```

so `total_issues_created` is the sum of both, while `total_pulls_submitted` gets the pulls alone. Every pull request is therefore reported twice.

It also disagrees with the identically named metric in the sibling judge: `dimensions-of-terrain/total_issues.rb:28` writes `json['issues']` from `total_issues_and_pulls`, which reads `repository.issues.totalCount` and excludes pull requests. Someone comparing the two series sees one name meaning two different things.

Probe with fbe's own fake values, 17 issues and 8 pulls:

```
{total_issues_created: 25, total_pulls_submitted: 8}
```

## What should happen

`total_issues_created` should hold the issue count, with pull requests reported only in `total_pulls_submitted`. If the sum is deliberate, the property should be named for what it holds.

morphqdd:
`total_issues_created` summed issues and pulls, so every pull request was
counted twice, once there and once in `total_pulls_submitted`. It now holds
the issue count alone, which is also what the identically named metric in
`dimensions-of-terrain` means.

Closes #2319


morphqdd:

PR #2339, opened 2026-09-17
#2317: call unmask_repos with a block, so its guards can run

### LINKED ISSUE #2317: `total_builds_ran` and `total_reviews_submitted` iterate repositories with `.each`, so the `Fbe.over?` guard never runs
## What happens

`Fbe.unmask_repos` only checks `Fbe.over?` when it is given a block. The guard at the top is `return if block_given? && Fbe.over?(...)`, and the per-repository `break if Fbe.over?(...)` lives inside `repos.each do |repo| ... yield(repo)` (`lib/patches/unmask_repos.rb:16` and `:68-71`).

`judges/quantity-of-deliverables/total_builds_ran.rb:11` and `total_reviews_submitted.rb:12` call `Fbe.unmask_repos.each do |repo|` instead. They take the plain array return and walk it themselves, so neither guard is reachable. Once the run is off quota or past its lifetime, the other three metrics in the same directory stop at the first repository while these two keep calling GitHub for every remaining one.

fbe's octokit decorator raises `Fbe::OffQuota` before every call once the quota is gone, and `Fbe::OffQuota` is in neither file's rescue list, so the run ends in an unhandled exception rather than a clean stop.

Probe with `Fbe.over?` stubbed to true and two repositories configured:

```
total_builds_ran     -> {total_builds_ran: 6}
api hits: [[:repository, "foo/foo"], [:repository, "bar/bar"], [:runs, "foo/foo"], [:runs, "bar/bar"]]
total_commits_pushed -> {total_commits_pushed: 0, total_hoc_committed: 0}
```

## What should happen

Both should use the block form, `Fbe.unmask_repos do |repo|`, like their three siblings.

morphqdd:
`Fbe.unmask_repos` checks `Fbe.over?` only when it is given a block, and
these two walked the returned array themselves, so neither guard could run
and the two metrics kept calling GitHub after the run was over. Both now use
the block form, like their three siblings.

Closes #2317


morphqdd:

PR #2340, opened 2026-09-17
#2315: rescue the error a GraphQL call actually raises

### LINKED ISSUE #2315: `Fbe::Error` from a GraphQL call is not rescued in four `quantity-of-deliverables` metrics
## What happens

`Fbe::Graph#query` turns every GraphQL failure into `raise(Fbe::Error, ...)`, and `pull_requests_with_reviews` and `total_commits` raise `Fbe::Error` by hand when the repository node comes back nil.

These four files do not list `Fbe::Error` in their rescues:

* `judges/quantity-of-deliverables/total_issues_created.rb:18`
* `judges/quantity-of-deliverables/total_releases_published.rb:16`
* `judges/quantity-of-deliverables/total_reviews_submitted.rb:32`
* `judges/quantity-of-deliverables/total_commits_pushed.rb:35`

What they do list, `Octokit::NotFound`, `Octokit::Deprecated` and `Octokit::Forbidden`, cannot be raised by a GraphQL call at all. So those arms are dead while the error that actually arrives escapes and kills the judge for every remaining repository and every remaining slice.

The same pair of defects was accepted and fixed for `dimensions-of-terrain/total_commits.rb` in #1979 and #1980. The `quantity-of-deliverables` siblings were never revisited. `dimensions-of-terrain/total_issues.rb:17` shows the correct shape.

Probe with a graph stub raising `Fbe::Error("Repository 'foo/foo' not found")`:

```
total_issues_created: CRASH Fbe::Error
total_releases_published: CRASH Fbe::Error
total_reviews_submitted: CRASH Fbe::Error
total_commits_pushed: CRASH Fbe::Error
```

## What should happen

Each of the four should rescue `Fbe::Error` and move on to the next repository, and drop the Octokit arms that the GraphQL path cannot produce.

morphqdd:
A GraphQL failure arrives as `Fbe::Error`, which none of the four rescued,
while the Octokit classes they did list cannot be raised by a GraphQL call
at all. So the dead arms are gone and `Fbe::Error` is rescued, and the
metric moves on to the next repository instead of taking the judge down.

Closes #2315


morphqdd:

PR #2341, opened 2026-09-17
#2320: write an array value element by element and refuse a nil

### LINKED ISSUE #2320: `Jp.fill_fact_by_hash` validates property names but not values, so an array or a nil fails one layer deeper
## What happens

The whole body of `Jp.fill_fact_by_hash` (`lib/fill_fact.rb:16`) is a guard over the property name. The value goes straight into `fact.public_send(:"#{prop}=", value)`.

Factbase accepts only scalars and refuses nil, so a hash whose value is an array or nil raises from inside Factbase, with a message that names neither the caller's hash nor the judge.

```
array: ArgumentError: The type 'Array' of 'commits' is not allowed
nil:   ArgumentError: The value of 'hoc' can't be nil
```

`Jp.incremate`, the other writer of metric hashes onto facts, handles exactly this with `Array(v).each { fact.__send__("#{k}=", _1) }` (`lib/incremate.rb:45`), which is how a multi-valued property is meant to be written. `fill_fact_by_hash` has no equivalent, so a caller cannot hand it the same shape of hash that a metric file returns.

This is the value-side counterpart of #2144, which was accepted for the name side.

## What should happen

The method should write array values element by element, as `Jp.incremate` does, and refuse a nil with its own message rather than letting Factbase raise.

morphqdd:
The value went straight into the fact, so an array or a nil failed one layer
deeper, with a message naming neither the hash nor the judge. An array is
now written element by element, the way `Jp.incremate` writes one, and a nil
is refused here.

Closes #2320


morphqdd:

PR #2342, opened 2026-09-17
#2316: require the unmask_repos patch in every quantity metric

### LINKED ISSUE #2316: The `quantity-of-deliverables` metrics do not require `lib/patches/unmask_repos`, so the version they get depends on load order
## What happens

All eight `dimensions-of-terrain` metrics do `require_relative '../../lib/patches/unmask_repos'`. None of the five `quantity-of-deliverables` metrics do; they require only `fbe/unmask_repos`.

The patch exists because the upstream method has no rescues: upstream `unmask_repos` lets `Octokit::Forbidden`, `Octokit::Unauthorized`, `Octokit::ServerError` and network errors out of `organization_repositories`, and `repos.reject! { octo.repository(repo)[:archived] }` raises on a repository that answers 404.

So whether `quantity-of-deliverables` survives an unreadable repository depends on whether a `dimensions-of-terrain` file happened to be loaded into the same process first.

Probe, fresh process loading only the `quantity-of-deliverables` files:

```
unmask_repos defined at: [".../fbe-0.49.1/lib/fbe/unmask_repos.rb", 62]
after loading a dimensions file: [".../lib/patches/unmask_repos.rb", 10]
```

The identical defect for a single dimensions file was accepted and fixed as #1987.

## What should happen

Every file that calls `Fbe.unmask_repos` should require the patch itself, as the `dimensions-of-terrain` files do.

morphqdd:
The five `quantity-of-deliverables` metrics required only `fbe/unmask_repos`,
so which version they got depended on whether a `dimensions-of-terrain` file
had been loaded into the same process first. Each of them now requires the
patch itself.

Closes #2316


morphqdd:

PR #2343, opened 2026-09-17
#2125: take the backlog sample size from the project options

### LINKED ISSUE #2125: `some_backlog_size` samples seven days and ignores `qos_days`

morphqdd:
The sampler kept the last seven days whatever the window was, so with
`qos_days: 30` the first twenty three days were never queried while the fact
claimed to cover them. The sample size now comes from
`qos_backlog_days` in the project options, and stays at seven when it is
not set.

Closes #2125


morphqdd:

PR #2344, opened 2026-09-17
#2323: do not report a contributor count that GitHub has capped

### LINKED ISSUE #2323: `total_contributors` silently saturates at the 500 contributors GitHub will name
## What happens

`judges/dimensions-of-terrain/total_contributors.rb:29` collects `contributor[:id]` from `Fbe.octo.contributors(repo)`, which is `GET /repos/{owner}/{repo}/contributors`. GitHub documents that endpoint as linking only the first 500 author email addresses to GitHub users; the rest come back as anonymous entries, and only when `anon` is requested, which it is not here. Anonymous entries carry no id, and the nil guard for them was added in #1990.

So for a repository with more than 500 contributors the set stops growing, and the judge writes the saturated number as the whole contributor count with nothing saying it was capped.

## What should happen

The metric should detect the cap and leave the property unset, or record that the value is a floor rather than a count.

morphqdd:
GitHub names at most 500 contributors on that endpoint and the rest arrive
anonymous, with no id, so a bigger repository reported the saturated number
as if it were the whole count. When the list comes back at the cap the metric
now reports nothing and says so in the log, instead of writing a floor.

Closes #2323


morphqdd:

PR #2345, opened 2026-09-17
#2318: do not report an active contributor count taken from one page

### LINKED ISSUE #2318: `total_active_contributors` reads one page of a hundred commits per repository and records it as the whole month
## What happens

`judges/dimensions-of-terrain/total_active_contributors.rb:16-19` issues its search through `Jp.qosearch`, which runs it inside `Fbe.octo.with_disable_auto_paginate` (`lib/qos_search.rb:70`). With auto-pagination off and fbe's client default of `per_page = 100`, `commits[:items]` holds at most a hundred commits and no second page is ever fetched.

The judge then walks those items and writes the distinct author count as `total_active_contributors`, with nothing saying it was a sample. Any repository with more than a hundred commits in the trailing thirty days, which is most active ones, is measured from a truncated slice, and `Jp.incremate` never recomputes the property.

Probe, recording octokit stub:

```
[[:with_disable_auto_paginate], [:search_commits, "repo:foo/foo author-date:>2024-02-02", {}]]
```

No `per_page` is passed and pagination is off, so one page is all the judge ever sees.

This is not #1986, which is about the Search API's 1000-result ceiling, ten times further out than the cap that actually binds here. It is the same defect accepted for `find-all-issues` in #2158 and for the quality-of-service metrics in #2304; this file was in neither.

## What should happen

Either page through the result set, or record that the value is a capped sample rather than a count.

morphqdd:
`Jp.qosearch` runs with pagination off, so the commit search answers at most
one page of a hundred and the judge wrote the authors of that page as the
whole month. When the page comes back full the metric now reports nothing
and says so in the log.

Closes #2318


morphqdd:

PR #2346, opened 2026-09-17
#2304: say when a quality metric measured only one page of results

### LINKED ISSUE #2304: Four quality-of-service metrics sample at most one hundred search hits and record it as the whole window
## What happens

`Jp.qosearch` runs every search inside `with_disable_auto_paginate` (`lib/qos_search.rb:70`), and fbe sets `per_page = 100`. So `found[:items]` holds at most one hundred entries and there is no second page.

These four iterate `found[:items]` and record what they see as the measurement for the window:

* `judges/quality-of-service/some_pull_hoc_size.rb:17`
* `judges/quality-of-service/some_review_time.rb:20`
* `judges/quality-of-service/some_issue_lifetime.rb:19`
* `judges/quality-of-service/some_triage_time.rb:32`

`some_merged_pulls.rb:18` runs the same query and reads `found[:total_count]`. For a repository with 250 merged pulls in the window, `some_merged_pulls` records 250 while `some_pull_hoc_size` measures 100 of them, and nothing in the fact says it was a sample.

```
total_count=250 but hoc samples=100
```

The same defect was accepted and fixed for one metric in #1905, for `some_backlog_size`. The four siblings were not revisited.

## What should happen

Either page through the result set, or record the cap in the fact so a truncated sample is not read as a census.

morphqdd:
`Jp.qosearch` runs with pagination off, so these four measured at most a
hundred items and recorded that as the whole window, with nothing saying it
was a sample. A full page is now reported in the log, through one helper the
four share.

Closes #2304


morphqdd:

PR #2347, opened 2026-09-17
#2309: check the quota before the three metrics that do not search

### LINKED ISSUE #2309: Three quality-of-service metrics spend hundreds of core API calls per repository with no quota check
## What happens

Every search-backed metric in `judges/quality-of-service` consults `Fbe.octo.off_quota?` before each repository. These three consult nothing:

* `some_build_success_rate.rb:15-63` calls `repository_workflow_runs`, which auto-paginates, then up to 60 `workflow_run_usage` calls per repository, one per run, in a loop with no quota check inside it.
* `some_release_hoc_size.rb:13-52` calls `Fbe.octo.releases(repo)`, which auto-paginates the entire release history before the `break` even looks at a date, then adds a `compare` call per consecutive pair.
* `some_release_interval.rb:11-34` calls `releases` the same way.

`Fbe.over?` is consulted between whole metrics in `lib/incremate.rb:26`, never inside one, so nothing can stop these loops once they start. On a project with many repositories a single metric can drain the core budget before any later metric gets a turn.

## What should happen

Check `Fbe.octo.off_quota?` per repository, and inside the per-run loop of `some_build_success_rate`, the way the search metrics do.

morphqdd:
These three walk workflow runs and release history with no quota check at
all, so on a project with many repositories one metric could drain the core
budget before any later metric got a turn. They now check
`Fbe.octo.off_quota?` per repository, and the run loop of
`some_build_success_rate` checks it per run.

Closes #2309


morphqdd:

PR #2348, opened 2026-09-17
#2308: attribute a recovery that happened after the window ended

### LINKED ISSUE #2308: `some_build_success_rate` discards every recovery that crosses a window boundary
## What happens

`judges/quality-of-service/some_build_success_rate.rb:14,55-60` creates `failed = {}` fresh for each fact, and records mean time to recovery only when the failure and the success that ends it both fall inside the same `fact.since..fact.when` scan.

A workflow that breaks near the end of one window and is fixed in the next contributes nothing. The open failure is forgotten when the method returns, and the next window starts with an empty `failed`, so the recovering run at `:57` finds no failure to close.

The metric therefore cannot report a recovery time longer than one window, and the cases it silently drops are exactly the slow recoveries the metric exists to surface.

## What should happen

Carry the open failure across windows, or read it back from the factbase, so a recovery is attributed to the failure that caused it instead of being discarded.

morphqdd:
A failure near the end of a window and its success in the next one were
never paired, because the open failure was forgotten when the metric
returned, so the slow recoveries the metric exists to surface were the ones
it dropped. A failure still open at the end of the window now looks for the
first success of that workflow after `fact.when`.

Closes #2308


morphqdd:

PR #2369, opened 2026-09-18
#2360: compare a bot login without regard to case

### LINKED ISSUE #2360: A bot is filtered out only when the case of its login matches the option exactly
## What happens

`lib/humans.rb` takes the bot names from the options as they are and compares them exactly:

```ruby
comments.reject do |c|
  c.dig(:user, :type) == 'Bot' || bots.include?(c.dig(:user, :login))
end
```

A GitHub login is not case sensitive: `SomeBot` and `somebot` are one account, and the API may
render either. `Array#include?` on strings is case sensitive, so the same bot slips through when
the case differs from what the operator wrote in the option:

```ruby
$options.bots = 'SomeBot,other-bot'
Jp.human_comments([{ user: { login: 'somebot', type: 'User' } },
                   { user: { login: 'SomeBot', type: 'User' } }])
```

```
[{user: {login: "somebot", type: "User"}}]
```

One copy of the very same bot is filtered and the other is kept as a human.

## What should happen

The login should be compared without regard to case, so a bot is excluded however GitHub happens
to render its name.

morphqdd:
A GitHub login is not case sensitive, and the comparison was, so the same bot was filtered out or
kept depending on the case the API happened to render.

Closes #2360


morphqdd:

PR #2370, opened 2026-09-18
#2358: count a run that ended in anything but success as a failed one

### LINKED ISSUE #2358: A check run that was cancelled or timed out is counted neither as a success nor as a failure
## What happens

`lib/pull_request.rb:180-185` sorts a check run into one of two buckets:

```ruby
case workflow[:conclusion]
when 'success'
  succeeded += 1
when 'failure'
  failed += 1
end
```

There is no other branch. GitHub also answers `cancelled`, `timed_out`, `neutral`,
`action_required` and `stale`, and each of those adds to neither counter, so the run leaves no
trace at all. A pull request whose only CI run was cancelled:

```
{succeeded_builds: 0, failed_builds: 0}
```

which is what a pull request with no CI at all looks like.

## What should happen

A run that ended in something other than success should be counted, so that a cancelled or
timed-out build is not the same as a build that never happened.

morphqdd:
A check run that was cancelled, timed out, or ended as neutral or action_required added to neither
counter, so a pull request whose only run was cancelled looked exactly like one nobody ran CI on.
Anything that is not a success now counts as a failure, except a run that is still without a
conclusion or was skipped, which counts as nothing at all.

Closes #2358


morphqdd:

PR #2371, opened 2026-09-18
#2363: keep the lifetimes that were measured when a later search fails

### LINKED ISSUE #2363: A failed pull request search in `some_issue_lifetime` throws away the issue lifetimes already measured
## What happens

`judges/quality-of-service/some_issue_lifetime.rb:12-26` measures two independent things in one
loop:

```ruby
{ issue: 'some_issue_lifetime', pr: 'some_pull_lifetime' }.each do |type, prop|
  ages = []
  Fbe.unmask_repos do |repo|
    ...
    found = Jp.qosearch(...)
    return {} if found.nil?
```

`return` leaves the whole method, not the current pass. The issue lifetimes are gathered first and
put into `ret`, and then a search failure on the pull requests throws them away. With the first
search answering one closed issue and the second answering nil:

```
calls made: 2
result: {}
```

`Jp.qosearch` answers nil for a transient reason, a rate limit or a timeout, so a measurement that
succeeded is lost because a different measurement did not.

## What should happen

A search that fails should cost only the property it was gathering, and the property that was
already measured should still be written.

morphqdd:
`return` left the whole method, not the pass it was in, so a search that failed while gathering
pull request lifetimes threw away the issue lifetimes that had already been measured. Each pass
now keeps what it collected and only the property whose search failed is left out.

Closes #2363


morphqdd:

PR #2372, opened 2026-09-18
#2362: say that a truncated tree costs the whole total, not one repository

### LINKED ISSUE #2362: One truncated tree stops the whole `total_files` scan and discards the repositories already counted
## What happens

`judges/dimensions-of-terrain/total_files.rb:41-48` handles a truncated tree like this:

```ruby
if tree[:truncated]
  $loog.info("Tree for #{repo}@#{info[:default_branch]} is truncated, skipping total_files")
  truncated = true
  break
end
files += (tree[:tree] || []).count { |item| item[:type] == 'blob' }
...
return {} if truncated
```

The log line says one repository is skipped, and the code stops the whole scan. Every repository
already counted loses its files, and every repository after this one is never asked about. With
`foo/one` holding two files, `foo/two` answering a truncated tree, and `foo/three` holding five:

```
result: {}
```

`foo/three` was never queried, and the two files of `foo/one` were counted and then thrown away.
The same run with `foo/two` alone would report nothing, which is the intended behaviour; the loss
of the other two repositories is not.

## What should happen

The truncated repository should be skipped the way the log line says, with the rest of the
repositories counted, or the fact should be written with the repositories that were readable.

morphqdd:
A truncated tree makes the judge report nothing at all, which is right: a sum that is missing one
repository is not the total it claims to be, and the same reasoning is behind #2169. The log line
said the opposite, that one repository was being skipped, so a reader of the log had no way to
know the metric was dropped for everybody. The message now says what happens.

Closes #2362


morphqdd:

PR #2373, opened 2026-09-18
#2361: say when only part of the completed runs was measured

### LINKED ISSUE #2361: `some_build_success_rate` measures the first sixty runs and records them as the whole window
## What happens

`judges/quality-of-service/some_build_success_rate.rb:31` keeps the first sixty completed runs of
a repository and drops the rest:

```ruby
wfs = workflows.select { |json| json[:status] == 'completed' && !json[:conclusion].nil? }.first(60)
```

Nothing looks at how many completed runs there were, and nothing is written to the log when the
cap bites. With seventy completed runs in the window:

```
runs available: 70
some_build_success_rate size recorded: 60
some_build_duration size recorded: 60
```

The three properties this judge writes, `some_build_success_rate`, `some_build_duration` and
`some_build_mttr`, therefore describe sixty runs while the fact claims to describe the window. The
sixty are the ones GitHub returned first, which is the newest ones, so the numbers lean towards
whatever happened last.

## What should happen

Either every completed run in the window should be measured, or the fact should say that the
numbers come from a sample, so nobody reads sixty runs as the whole window.

morphqdd:
Sixty completed runs were measured and recorded as if they were the window, with nothing said
about the rest. The cap stays, since the calls it saves are the point of it, and the judge now
says in the log how many of the completed runs it actually measured.

Closes #2361


morphqdd:

PR #2377, opened 2026-09-18
#2356: skip a pull on a transient error in its comment and build helpers

### LINKED ISSUE #2356: A 5xx inside the comment and build helpers aborts `pull-was-merged.rb` despite its transient rescues
`judges/pull-was-merged/pull-was-merged.rb` rescues `Octokit::ServerError` on its pull, issue and reviews lookups (lines 82, 111 and 138, added in #1725) and skips the pull. Most of its API calls for a pull happen later, inside `Fbe.fb.txn`, through `Jp.comments_info`, `Jp.fetch_workflows` and `Jp.count_suggestions` (lines 162, 163 and 169). Those helpers rescue only `NotFound`, `Deprecated` and `Forbidden`: the comment, reaction, check run, workflow job, workflow run and review comment calls at `lib/pull_request.rb:16, 29, 99-126, 134, 152, 167, 208`.

A 502 on, say, `check_runs_for_ref` leaves the transaction, which discards the fact, and `fbe/iterate.rb:332` turns it into `Fbe::Error`. The judge stops for every repository left in the cycle, and the next cycle begins at the same pull, while the same 502 on `pull_request` a few lines earlier would only have skipped it.

Calling the three helpers before `Fbe.fb.txn`, inside a rescue with the judge's existing transient classes and `next issue`, would make the whole pull follow the rule #1725 set for its first three calls.

@yegor256 please take a look

Thayorns:
`pull-was-merged` already skips a pull when its pull, issue or reviews lookup hits a 5xx, a rate limit or a network timeout. The remaining calls for the pull, made by `Jp.comments_info`, `Jp.fetch_workflows` and `Jp.count_suggestions`, ran inside `Fbe.fb.txn` and rescued only `NotFound`, `Deprecated` and `Forbidden`, so a 502 there escaped the iterator as `Fbe::Error` and stopped the judge for every repository left in the cycle.

The three helpers now run before the transaction, inside a rescue with the same transient classes the judge already uses, and the pull is skipped with `next issue` so the next cycle retries it. The transaction only writes the collected values into the fact.

A new test stubs a 502 on the issue comments of one pull and checks that the next pull is still recorded as merged.

Closes #2356


Thayorns:

PR #2378, opened 2026-09-18
Skip a merged pull when its comments or builds are forbidden

### LINKED ISSUE #2352: A 403 in `lib/pull_request.rb` leaves zero comments and builds in a `pull-was-merged` fact permanently
`judges/pull-was-merged/pull-was-merged.rb:162-169` fills the new fact from `Jp.comments_info`, `Jp.fetch_workflows` and `Jp.count_suggestions` inside `Fbe.fb.txn`. Each helper catches `Octokit::Forbidden` (`lib/pull_request.rb:20, 33, 89, 107, 119, 138, 156, 171, 212`), logs "transient, will retry next cycle", and returns `[]`, `0` or zero builds. The transaction then commits normally, so the fact is saved with `comments: 0`, `comments_appreciated: 0`, `succeeded_builds: 0` or `suggestions: 0`, depending on which call was refused.

No next cycle ever comes for those numbers. The query at `pull-was-merged.rb:34-39` skips the pull from then on because its fact exists, and `fix-missing-comments.rb:24` only picks facts with no `comments` at all. One secondary rate limit, which GitHub reports as a 403, freezes those counts at zero, while the log promises a retry.

When a helper call is refused, the judge should skip the pull for this cycle and commit nothing, the same way it already does for a 403 on the pull, issue and reviews lookups at lines 73, 102 and 127.

@yegor256 please take a look

Thayorns:
When GitHub answered 403 to one of the comment or check-run lookups, the helpers in `lib/pull_request.rb` logged a retry and returned zero. `pull-was-merged` then saved the fact with those zeros, and since the fact now existed, the pull was never looked at again. The same happened in `fix-missing-comments`, which wrote `comments: 0` and dropped the fact from its query.

Now `comments_info` and `fetch_workflows` let `Octokit::Forbidden` through from the code comments, issue comments and check runs requests. `pull-was-merged` gathers comments, builds and suggestions before opening the transaction and skips the pull on a 403, just like it already does for the pull, issue and reviews lookups. `fix-missing-comments` leaves `comments` absent so the next cycle retries. `github-events` already rescues 403 per event and skips it, so it needs nothing.

Reactions, resolved threads, workflow runs and review comments still turn a 403 into zero. That part is left as a puzzle in `lib/pull_request.rb` to keep this change small.

Closes #2352


Thayorns:
@yegor256 take a look please, happy to clarify anything about the change.

PR #2379, opened 2026-09-18
#2357: count only a positive reaction as appreciation

### LINKED ISSUE #2357: `Jp.count_appreciated_comments` counts a thumbs-down as appreciation
## What happens

`lib/pull_request.rb:99-126` counts a comment as appreciated once per reaction that somebody other
than the author of the comment left on it:

```ruby
Fbe.octo.issue_comment_reactions(repo, comment[:id])
  .count { |reaction| reaction.dig(:user, :id) != comment.dig(:user, :id) }
```

Nothing looks at `reaction[:content]`. GitHub has negative reactions too, and `-1` and `confused`
are counted the same way as `+1`, `heart` and `hooray`. With one comment carrying a `-1` from one
person and a `confused` from another:

```
comments_appreciated = 2
```

The property is written into the fact by `lib/pull_request.rb:51`, so a comment two people
disagreed with is recorded as appreciated twice.

## What should happen

Only the positive reactions should count, the way a reader of the property would expect.

morphqdd:
Every reaction from somebody other than the author counted, so a thumbs-down and a confused face
were recorded in `comments_appreciated` exactly like a heart. Only the positive reactions count
now. The reaction fixtures in the existing tests carried no `content` at all, which real GitHub
always sends, so they carry one.

Closes #2357


morphqdd:

PR #2380, opened 2026-09-18
#2350: evaluate unique after all filters in pull-was-merged query

### LINKED ISSUE #2350: `pull-was-merged` loses a pull when its first fact is stale, because `unique` runs before the filters
The query in `judges/pull-was-merged/pull-was-merged.rb:27` lists `(unique repository issue)` third, before the two `(empty ...)` checks, the `(or (eq what ...))` list and `(absent stale)`. Factbase's `and` evaluates its operands in order and stops at the first false one (`factbase/terms/and.rb:24`), and `unique` records the pair the moment it is evaluated (`factbase/terms/unique.rb:28`). So the first fact of a pull that the query reaches takes the `(repository, issue)` pair even when a later operand rejects it, and every other fact of that pull is then rejected by `unique`.

One concrete way to get there: the author of a pull deletes their account, and `who-is-alive.rb:53-54` marks every fact carrying that `who` as stale, including the `pull-was-opened` fact. The `code-was-reviewed` fact of the same pull carries the reviewer and stays clean, so it should trigger the judge, but it never does when the stale opening fact is reached first. The pull ends up with no `pull-was-merged` or `pull-was-closed` fact, and whether it does depends only on the order the facts were inserted.

`pull-was-opened.rb:31` had this in #887 and #899 and now places `unique` after all of its filters. Moving `(unique repository issue)` to the end of the `and` here, after `(eq where 'github')`, fixes it, and a test with a stale `pull-was-opened` followed by a clean `code-was-reviewed` would keep it that way.

@yegor256 please take a look

Thayorns:
The query in `pull-was-merged` evaluated `(unique repository issue)` third, before the `empty` checks, the `what` list, `absent stale` and `eq where`. Since factbase's `and` stops at the first false operand and `unique` records the pair as soon as it is evaluated, the first fact of a pull reached by the query claimed its `(repository, issue)` pair even when a later filter rejected it, and every other fact of that pull was then refused.

This moves `unique` to the end of the `and`, the same place `pull-was-opened` keeps it. Three tests cover the ways a pull got lost: a stale opening fact ahead of a clean review fact, a fact with an unrelated `what` ahead of the opening fact, and a fact from another platform ahead of the github one.

Closes #2350


Thayorns:
@yegor256 happy to clarify anything about the query order change if it helps the review.

PR #2382, opened 2026-09-18
#2359: count a review comment as a suggestion only when it suggests a change

### LINKED ISSUE #2359: `Jp.count_suggestions` counts every top-level review comment, not the ones that suggest a change
## What happens

`Jp.count_suggestions` (`lib/pull_request.rb:190-223`) counts like this:

```ruby
comments.count do |comment|
  comment.dig(:user, :id) != author && comment[:in_reply_to_id].nil?
end
```

Nothing reads the body of the comment. A review comment counts as a suggestion as long as it is
not a reply and not written by the author of the pull request, so ordinary praise counts:

```ruby
# body: "Looks good, nice change! Thanks a lot for doing this."
Jp.count_suggestions('foo/foo', 77, 1, [{ id: 900, user: { id: 2 } }])
```

```
1
```

The number goes into `nn.suggestions` in `judges/pull-was-merged/pull-was-merged.rb:169`, where it
reads as the count of suggested changes.

## What should happen

Only a comment that carries a suggestion block should be counted, so that the property holds what
its name says.

morphqdd:
Any top-level comment from a reviewer counted, so praise counted as a suggested change, and the
number went into `suggestions` where it reads as the count of suggestions. A comment now counts
when it carries a suggestion block, which is what GitHub renders as a change the author can apply.
The review-comment fixtures in the existing tests carried no body at all, so they carry one.

Closes #2359


morphqdd:

